PR #173, opened 2018-08-31
Add lower type bounds to the entity argument X

### LINKED ISSUE #108: Path.get(PluralAttribute<X, C, E>) lower bound missing on X
Path interface declares:

<tt><E, C extends java.util.Collection<E>> Expression<C> get(PluralAttribute<X, C, E> collection);</tt>

it should be instead

<tt><E, C extends java.util.Collection<E>> Expression<C> get(PluralAttribute<? super X, C, E> collection);</tt>

with lower bound on X.
The same apply on <tt>Path.get(MapAttribute<X, K, V>)</tt>.

krichter722:
Add lower type bounds to the entity argument X of the return value of
Path.get(PluralAttribute) and Path.get(MapAttribute). The absense of
these lower bounds prevent usage of the JPA criteria API for a
collection field f which is declared in an entity A which then is
extended by another entity B and used in the type bounds of TypedQuery.

closes #108


lukasj (commented):
this change breaks backwards compatibility, so this can be considered for inclusion in the next major version

krichter722:
@lukasj Is there a process for including the the change in the next major release? Afaik 2.2 and 3.1 have been released without any feedback on this. There's a least one bug/use case which is adressed by this change.

gavinking:
You mean to change it to:

    <E, C extends java.util.Collection<E>> Expression<C> get(PluralAttribute<? super X, C, E> collection);

Right?

If so I agree that this should be done in the next major release.

gavinking:
> You mean to change it to: ... Right?

Oh yeah, that's what the attached commit does :)

PR #234, opened 2019-11-20
Update nullable javadoc for javax.persistence.Column

caodangtinh:
Since this option only use when generating the DDL for the column by the persistence provider but current document is not clear explain about that.


lukasj:
fix the merge conflict in the new PR, please. Thx!

PR #254, opened 2020-02-15
Fix section headers

stuartwdouglas:
Change all section headers to align with the original spec.

This also fixes the generated table of contents.

Signed-off-by: Stuart Douglas <sdouglas@redhat.com>


lukasj:
this shows merge conflict and it is probably fixed already. Can you submit the change in the new PR if needed, please? Thx!

PR #263, opened 2020-03-04
removing redundant parentheses.

i-love-tadpole:
Signed-off-by: bo he <qq111372757@gmail.com>


i-love-tadpole:
If the little error do not need to fix please just noticing me and I will close the pr.

PR #272, opened 2020-04-09
Fixed asciidoc error by updating the asciidoctorj version

moghaddam:
After fetching the latest version of the master, generating asciidocs stopped working. This commit fixes that by updating the asciidoctorj version.

### DIFF
--- a/spec/pom.xml
+++ b/spec/pom.xml
@@ -31,7 +31,7 @@
         <site.output.dir>${project.build.directory}/staging</site.output.dir>
         <maven.site.skip>true</maven.site.skip>
         <asciidoctor.maven.plugin.version>2.0.0-RC.1</asciidoctor.maven.plugin.version>
-        <asciidoctorj.version>1.6.2</asciidoctorj.version>
+        <asciidoctorj.version>2.2.0</asciidoctorj.version>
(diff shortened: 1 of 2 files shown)


spec/pom.xml
```diff
@@ -31,7 +31,7 @@
         <site.output.dir>${project.build.directory}/staging</site.output.dir>
         <maven.site.skip>true</maven.site.skip>
         <asciidoctor.maven.plugin.version>2.0.0-RC.1</asciidoctor.maven.plugin.version>
-        <asciidoctorj.version>1.6.2</asciidoctorj.version>
+        <asciidoctorj.version>2.2.0</asciidoctorj.version>
         <asciidoctorj.pdf.version>1.5.0-alpha.16</asciidoctorj.pdf.version>
```
lukasj: actually what I've intended to do was to remove `asciidoctorj` and `jruby` dependencies and update `asciidoctorj.pdf.version` to `1.5.3`. Wondering what happened on my end so I forgot to add `git add` this ... :-/

moghaddam: You're right. I applied the changes you suggested and it works :+1:

moghaddam: So I think we can close this PR

lukasj: If you'll update this PR, I merge it right away ;-)

moghaddam: Done.

Issue #271, opened 2020-04-08
Split the specification document


lukasj:
Currently the source of the spec document has some 30k lines leading to pdf with >600 pages.

To allow more people to work together on the spec at the same time, the doc should be split into smaller parts, preferably one file per chapter/appendix for now

moghaddam:
As you know I just started working on the spec and I'm completely agree with you. Currently it's not easy to navigate in that file. I can do the split while converting the spec to Jakarta EE.

viswanadhk:
one file per chapter means almost 12 splits now. The main point is consistency. As long as the formatting consistency remains the same, I think we can split.

moghaddam:
I finished the splitting and its PR (#273) is waiting for review.

lukasj:
fixed by #273, thanks!

PR #273, opened 2020-04-09
Split spec document

moghaddam:
Split spec document based on chapters as per #271


moghaddam:
Copyright block added. Also I renamed all chapter files to start with ch0N (eg. ch01-introductin.adoc).

PR #290, opened 2020-09-01
Drop references to older/Java EE versions of the schema from the spec

lukasj:
With relaxed overall backwards compatibility related requirements of Jakarta EE 9, I'd like to remove explicit requirements wrt supporting older/Java EE schemas defined in the persistence spec

PR #291, opened 2020-09-07
replace java persistence with jakarta persistence

lukasj:
update spec changelog
drop old schemas from the jar

Signed-off-by: Lukas Jungmann <lukas.jungmann@oracle.com>


dmatej:
Deleted XSD files?

m0mus (approved):
LGTM

lukasj:
@dmatej Yes. The spec itself now refers to 2.2 and 3.0 (to match EE 8 and 9), so these are kept, older ones are available online and vendors are free to ship them in the implementations. Or do you prefer to keep them in the repo but not in the 3.0.0 jar file?

dmatej:
Not sure in this case, I'm only worried if it would not break some backward compatibility so people would have troubles with upgrades. Online access is impossible on some closed environments without access to the internet and I'm not sure if all vendors include all these files (also I still remember issues with JPA 1?/2? when we had api from EclipseLink and another from Hibernate ... with differences). On application servers it is usually not a problem, but ... world still uses really old things.

I can't say I am against this step, it's just a question or "theme" to be well considered. We can try it and just be prepared to react :-)

But generally I would rather prefer it would be in a standalone PR and commit.

dmatej (approved):

lukasj:
@dmatej I agree it should have been done through #290 I've missed it there. I also understand your concerns. On the other hand Jakarta EE 9 aims for being backward compatible with Jakarta EE 8 only while leaving the compatibility with Java EE X up to vendors. I also believe that in most cases for versions 2.0, 2.1 (and 2.2) it should be about changing document's namespace together with javax->jakarta replacements. For version 1.0 it may be more work. Note that this can be the only chance we have to drop old javax related stuff.
Another point is that some people believe that schemas should have never been part of api jar and it should be implementors'/vendors' responsibility to ship them. I think it is better to ship only the (latest) required to be supported versions in the api jar to allow offline access to them, to relax requirements on vendors a bit to lower the entry bar for new implementations and to guarantee all gets the same variants of schemas (there were cases where official final schemas were slightly different from vendor's ones)

Issue #264, opened 2020-03-10
Avoid transitive dependencies in module-info


Sanne:
Hello,
many thanks for introducing a module-info, that's great progress.

I wonder if it would be possible to apply some minor tuning; in particular I would prefer, if possible, if you could remove the `transitive` qualifier from all modules.

For context, I work on the Hibernate project, and while I understand this might be surprising, we actually do have several projects which don't necessarily make use of neither `java.instrument` nor `java.sql`, while still using the JPA API.

I would suggest that importing of such modules should be left to the implementors?

Tangentially, we don't use `java.instrument` and I noticed the JPA API only requires it for the exception on a signature from `jakarta.persistence.spi.ClassTransformer` - which in several runtimes we don't use. So ideally it would be great to be able to use an optimised JDK which doesn't load the `java.instrument` module at all, but I realize this is an existing API and we can't do much about it yet.

Happy to send a PR to remove the `transitive` qualifier, if you agree?

Thanks!

scottmarlow:
Thanks for raising this @Sanne!  I was just looking at [module-info](https://github.com/eclipse-ee4j/jpa-api/blob/master/api/src/main/java/module-info.java).  

How about java.logging?

I wonder if the module-info.java source document via comments, the dependencies instead:
`
/* requires java.logging;
    requires transitive java.instrument;
    requires transitive java.sql; */
`

Sanne:
> How about java.logging?

Hibernate ORM doesn't that either, but this one isn't marked as `transitive` it's not a problem: I'd say that makes it an internal implementation detail.

Clearly I'd be even happier if it wasn't depending on it but since we're not here to debate internal implementation details I won't complain :)

anthonyvdotbe:
Both modules appear in the public API (e.g. `java.instrument` in javax.persistence.spi.ClassTransformer and `java.sql` in javax.persistence.spi.PersistenceUnitInfo), so I believe they should be marked as transitive (if I'm not mistaken, javac will also give an Xlint warning if they're not).

The modular solution would be to define a "JPA lite" which doesn't depend on those modules, and redefine the API module as an aggregate module:
````java
module jakarta.persistence {
    requires transitive jakarta.persistence.core;
    requires transitive jakarta.persistence.spi;
}
````

Sanne:
yes, I relize they are transitive so to make it easier to consumers - but I'm one of those consumers and I'm not going to need them, so I'd rather see it as an explicit opt-in we should do.

[...]

PR #313, opened 2021-04-27
#264: avoid transitive dependency on java.instrument

lukasj:
Making dependency on `java.instrument`/`java.lang.instrument.IllegalClassFormatException` optional by introducing new `Transformer` interface superseding current `ClassTransformer` interface, which has been deprecated. Specification has been updated to make implementation of  `ClassTransformer` by the persistence provider explicitly optional.


m0mus (approved):
LGTM

Issue #102, opened 2015-02-28
Add datetime functions


lukasj:
There is currently no way to extract parts of a datetime attribute in a standardized way. Hibernate, EclipseLink and DataNucleus have support for that. I propose to add the functions:

*   YEAR
*   MONTH
*   DAY
*   HOUR
*   MINUTE
*   SECOND

m-reza-rahman:
Is this really needed any more with the Java SE date time API? If not, this should probably be closed to reduce clutter.

Reza Rahman
Jakarta EE Ambassador, Author, Blogger, Speaker

Please note views expressed here are my own as an individual community member and do not reflect the views of my employer.

AleksNo:
Hi, 

those functions would be still really nice because they would be usable in the having-clause for filtering data. Currently you have to use native SQL or CriteriaBuilder#function (i guess there is also a JPQL equivalent for this) to do such things, AFAIK. But in both cases you use vendor specific database functions.

Cheers

beikov:
It sure is needed for various queries, but I think it might be more appropriate to standardize the `EXTRACT` function like in the SQL standard.

gavinking:
> It sure is needed for various queries, but I think it might be more appropriate to standardize the `EXTRACT` function like in the SQL standard.

Exactly.

lukasj:
`EXTRACT` sounds good to me. Is anyone willing to take ownership of this to make sure we get this into 3.1?

gavinking:
Well, I was planing on writing up a proposal for addition of a number of functions to JPAQL, `extract()` being just one of them. (This was an issue I looked at quite deeply about a year or so ago.)

So I was going to write that up and propose it in an issue this weekend. So assuming we can reach agreement on the precise list functions that we want to add, I'm very happy to write that up and send a PR to the spec.

gavinking:
(FTR, `extract()` is probably the most complex of the functions I would like to see added, since different databases support a rather diverse variety of field types. So perhaps that's something that should still be discussed separately here.)

beikov:
>the precise list functions that we want to add

I guess there are a few that are interesting, but maybe we should split this effort into separate issues so that we can merge certain efforts earlier if we reach agreement on parts?

Regarding temporal functions, I would love to see the following:

* Functions or operator support for arithmetic on temporal and interval/duration types `extract(seconds from ts1 - ts2)` or `ts1 + interval '1 hour'`
[...]

PR #314, opened 2021-05-03
specify JPAQL EXTRACT function

gavinking:
Proposal for addition of `EXTRACT()` to JPAQL.

I made the following choices, which should be reviewed:

- I defined `EXTRACT` as a (new) keyword for consistency with how other function names are defined. (I don't like that the JPA spec treats function names as keywords, FTR.)
- But I defined the field types as *identifiers*, because I didn't want to add lots of new keywords, and because I wanted to make it easy to extend this list in either (a) future spec revisions, and (b) vendor extensions to the spec.
- (For now) I included only the field types `YEAR`, `QUARTER`, `MONTH`, `WEEK`, `DAY`, `HOUR`, `MINUTE`, `SECOND`, `DATE`, `TIME`. 
 
Note that our product&mdash;and some databases&mdash;support more field types than these, but I've limited the functionality to the field types that are easiest to implement.

Thoughts, please?

See #102.

### DIFF
--- a/spec/src/main/asciidoc/ch04-query-language.adoc
+++ b/spec/src/main/asciidoc/ch04-query-language.adoc
@@ -320,9 +320,9 @@ CHAR_LENGTH, CHARACTER_LENGTH, POSITION, and UNKNOWN are not currently
 used: they are reserved for future use.],
 BOTH, BY, CASE, CHAR_LENGTH, CHARACTER_LENGTH, CLASS, COALESCE, CONCAT,
 COUNT, CURRENT_DATE, CURRENT_TIME, CURRENT_TIMESTAMP, DELETE, DESC,
-DISTINCT, ELSE, EMPTY, END, ENTRY, ESCAPE, EXISTS, FALSE, FETCH, FROM,
-FUNCTION, GROUP, HAVING, IN, INDEX, INNER, IS, JOIN, KEY, LEADING, LEFT,
-LENGTH, LIKE, LOCATE, LOWER, MAX, MEMBER, MIN, MOD, NEW, NOT, NULL,
+DISTINCT, ELSE, EMPTY, END, ENTRY, ESCAPE, EXISTS, EXTRACT, FALSE, FETCH,
+FROM, FUNCTION, GROUP, HAVING, IN, INDEX, INNER, IS, JOIN, KEY, LEADING,
+LEFT, LENGTH, LIKE, LOCATE, LOWER, MAX, MEMBER, MIN, MOD, NEW, NOT, NULL,
 NULLIF, OBJECT, OF, ON, OR, ORDER, OUTER, POSITION, SELECT, SET, SIZE,
 SOME, SQRT, SUBSTRING, SUM, THEN, TRAILING, TREAT, TRIM, TRUE, TYPE,
 UNKNOWN, UPDATE, UPPER, VALUE, WHEN, WHERE.
@@ -1694,7 +1694,8 @@ functions_returning_numerics ::=
     SQRT(arithmetic_expression) |
     MOD(arithmetic_expression, arithmetic_expression) |
     SIZE(collection_valued_path_expression) |
-    INDEX(identification_variable)
+    INDEX(identification_variable) |
+    extract_datetime_field
 ----
 
 The ABS function takes a numeric argument and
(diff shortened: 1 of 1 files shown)


gavinking:
Oh, I realized there's an issue I missed. We need to define that `EXTRACT(DATE...)` and `EXTRACT(TIME...)` return `LocalDate` and `LocalTime` (rather than `javax.sql` types) when they occur in the `select` clause.

Need to dig a little and consider what's the best way to do this.

gavinking:
> Need to dig a little and consider what's the best way to do this.

Well, where precisely does the spec say that `CURRENT_DATE` is a `java.sql.Date` and not a `LocalDate`. I'm not immediately seeing it.

gavinking:
> Well, where precisely does the spec say that `CURRENT_DATE` is a `java.sql.Date` and not a `LocalDate`. I'm not immediately seeing it.

It seems to me that the spec actually leaves this undefined. Ooops!

Or perhaps I'm missing something?

beikov:
How about at least standardizing the field names and semantics of `day of year` etc. but state in the spec that the support is optional? If EclipseLink manages to also support these optional field types, we can re-evaluate in the next revision of the spec whether this should stay optional. Other than that, I like the change, although I would also like to see the "result type" of `extract(date from ..)` and `extract(time from ..)` being specified.

Not sure if this should be done as part of this PR, but the JPA Criteria API should also be updated to incorporate this functionality.

gavinking:
> How about at least standardizing the field names and semantics of `day of year` etc. but state in the spec that the support is optional? 

I'm not a huge fan of optional features in a spec, but I'm curious to hear what others think.

> Not sure if this should be done as part of this PR, but the JPA Criteria API should also be updated to incorporate this functionality.

Good point. This should certainly be added to the criteria API, but I would prefer to get feedback on this much of the change first.

lukasj:
> I'm not a huge fan of optional features in a spec, but I'm curious to hear what others think.

Neither am I. IMO optional stuff should be limited to bare minimum - what's the point of having functionality nobody implements? On the contrary, it makes sense to make "deprecated" or "to-be-removed" functionality optional before actually removing it.

>  If EclipseLink manages to also support these optional field types

this should no longer be criteria for inclusion of new features in the spec. The thing is that at the time of the spec release, there must be "some" - can be eclipselink, hibernate, other-awesome-impl - implementation available to be used for spec ratification. At the platform level, at least one compatible implementation supporting _all_ optional features must be available - I'm not 100% sure this criteria applies also to particular specs but I wouldn't be surprised if it does.

In any case - it would be nice to be able to use at least 2 different implementations for the spec ratification process this time.

I'm more worried about TCKs - it is very unlikely someone will write them for us...

> I would prefer to get feedback on this much of the change first.

Isn't there an inconsistency between `WEEK` and `YEAR`? The former is tight to ISO-8601 while the other does not seem to be - is that intentional? Other than that it looks good to me.

gavinking:
> The thing is that at the time of the spec release, there must be "some" - can be eclipselink, hibernate, other-awesome-impl - implementation available to be used for spec ratification.

Oh! I did not know about that!

Even so, I would prefer to be sure that other implementors are going to be happy to implement a feature before throwing it into the spec. (Or, at very least, give them a decent opportunity to object.)

We don't want to discourage people from implementing JPA.

> In any case - it would be nice to be able to use at least 2 different implementations for the spec ratification process this time.

Yes.

> I'm more worried about TCKs - it is very unlikely someone will write them for us...

If it's not too hard to write a TCK test (I've never looked) we can contribute tests too. But I don't know how that would work if the RI doesn't yet implement the feature.

> Isn't there an inconsistency between `WEEK` and `YEAR`? 

I will double-check. (I imagine it's intentional.)

beikov:
>Neither am I. IMO optional stuff should be limited to bare minimum - what's the point of having functionality nobody implements? On the contrary, it makes sense to make "deprecated" or "to-be-removed" functionality optional before actually removing it.

Well, Hibernate already implements this. I was thinking that it might be a good idea to mark some field types as optional first to give EclipseLink some time to work on this, but if we all agree that all field types make sense there is no need for making anything optional since Hibernate could be the implementation for ratification.

Tomas-Kraus:
I'll implement some TCK test based on this PR grammar definition to see whether both Hibernate and Eclipselink are OK. :)

Tomas-Kraus:
What exactly do you mean by the `For the following field type identifiers, EXTRACT returns an integer
value` sentence? Integer as common Number that can hold the value or exactly Integer/int type?

Current specification will work well with anything from Short to Long.
But it may not be sufficient in the future.  For example PostgreSQL EXTRACT supports EPOCH - https://www.postgresql.org/docs/8.1/functions-datetime.html - which requires Long type.

Current Eclipselink returns EXTRACT result as Long instance. What other JPA implementations do?

I used following code in TCK:
```
  public void testExtractDateYear() throws Fault {
    try {
      TypedQuery<Number> q1 = getEntityManager().createQuery("SELECT EXTRACT(YEAR FROM d.id) FROM DataTypes2 d WHERE d.id = :id", Number.class);
      q1.setParameter("id", t4Date());
      long y = q1.getSingleResult().longValue();
      System.out.println("DATETIME -- YEAR: " + y);
      if (y != TD4_YEAR) {
        throw new Fault("EXTRACT(YEAR FROM date) returned wrong value: expeted " + TD4_YEAR + " but got " + y);
      }
    } catch (Exception ex) {
      throw new Fault(ex);
    }
  }
```
so it will accept any Number value and evaluate it as long. But it would be nice to know exact return type specification here.

beikov:
So from the PostgreSQL documentation it follows that on SQL level we will have to assume that this returns a `numeric`. I guess only `second` and `epoch` can return a actual decimal number though which is why we should probably map these to `BigDecimal` by default, but the other fields should IMO map to `BigInteger` to preserve precision. Maybe we can have some exceptions like for `quarter` etc. where the value space is known, but you get the point. As long as we don't lose precision, I'm fine with any mapping.

Tomas-Kraus:
I don't like BigInteger/BigDecimal because of poor performance compared to smaller number types. Long shall be large enough for almost everything. Seconds (miliseconds, microseconds, ...) may require BigDecimal in some cases, but Long/Float may be good enough in many use-cases.
Old epoch / unixtime is just number of seconds so Long is OK. But there are implementations with miliseconds precission. But they still fit into Long.
Maybe Number interface is good enough as a return type and we can let implementations to choose what's best for them.

lukasj (approved):
I believe this is good enough and since I've seen no complains against this, I'll merge this. Further corrections can be made in subsequent PRs if needed.

spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -1735,11 +1736,62 @@ AND INDEX(w) = 0
...
+For the following field type identifiers, EXTRACT returns an integer
+value:
+
+- YEAR means the calendar year.
+- QUARTER means the calendar quarter, numbered from 1 to 4.
+- WEEK means the ISO-8601 week number.
+- DAY means the calendar day of the month, numbered from 1.
+- HOUR means the hour of the day in 24-hour time, numbered from 0 to 23.
+- MINUTE means the minute of the hour, numbered from 0 to 59.
+- SECOND means the second of the minute, numbered from 0 to 59.
```
gavinking, line 1770: Ugh, @lukasj I screwed up here. 

Actually I think it's much better if `SECOND`, unlike the other fields, includes the fractional part. So it's a floating point value, not an integer. (Databases are pretty inconsistent here, but this interpretation _can_ be implemented portably.)

Issue #152, opened 2017-07-14
Native support for UUID class


lukasj:
It would be really great if `UUID` would be a valid type for (at least PK) attributes. Certainly it can be provided by an application using an adapter class, but in the context of massively distributed environments it seems to be a `common` need of all cloud applications.

mkarg:
@mkarg I was the original reporter.

sebersole:
Ideally we'd allow the user some way to specify the database storage for these.  Maybe even a limited set like is done for enums.  The ability to specify the JDBC type code (and let the provider figure it out) would be good also

rbygrave:
In terms of how UUID is stored [VARCHAR, BINARY, NATIVE + an optimised option] a point I'd like to add is that this storage choice is more per persistence context / database than per entity.  That is, it is more usual/expected that all UUID for Postgres are stored natively vs all UUID with Oracle stored as BINARY etc.

That is, I'm not sure we need the storage type on the `@UUID` [noting that from an Ebean perpective we'd say we do not need `@UUID` at all - it is more provider side configuration of how the Java UUID type is stored/mapped as a database column.   

I'm not sure I've explained that well but the short version is I don't think we need `@UUID( Types.VARCHAR )` for a specific entity property (but instead this is more a global provider config/type mapping).

sebersole:
Just like with `@GeneratedValue#strategy` or `@Enumerated#value` there will be a default/auto choice which will in fact do exactly what you suggest.  

However, just because something is "more usual/expected" does not mean that is how everyone uses these things.  I could make the same argument for example wrt id-generators.  I could say that it is "more usual/expected" when running on Oracle to use a sequence; but not everyone does that for every id.

And that last bit ("for every id") is important - I might generally want to store UUID as a PGSQL UUID but for a particular attribute store it as a VARCHAR.  That's not completely unreasonable.  So you'd need a way to indicate, for this one attribute, use a different strategy

rbygrave:
>  for a particular attribute store it as a VARCHAR. That's not completely unreasonable.

Yes, fair point.  I think this case is currently covered by either using a "custom named generator" via `GeneratedValue.generator` or a `@Column` mapping?  Perhaps I am missing something?  Not as elegant/nice/explicit as a `@UUID(Types.VARCHAR)`.

sebersole:
[...]

Issue #151, opened 2017-07-14
GenerationType.UUID


lukasj:
Modern applications more and more are using _replication_ technologies, inducing the problem of _global_ ID generation.

These systems tend to abstain from _central_ ID algorithms like `TABLE` or `SEQUENCE`, but instead of _UUIDs_ as PKs. The benefit is clear: The JVM of the client can produce a UUID super fast even if the network is down, and no PK clash will ever occur.

I think it would be really great particularly for cloud applications if a new `GenerationType.UUID` would instruct persistence providers to e. g. simply call JRE's `UUID.randomUUID()` method and prevent *any* network roundtrips (or providers could decide to ask for UUID from server or JDBC driver by SQL).

This is feasible and very easy to add to EclipseLink.

mkarg:
@mkarg I was the original reporter.

gavinking:
I think it's worth doing this and #152. In Hibernate this looks like:

```java
@Id
@GeneratedValue
UUID id;
```

And from the discussion above it looks like other products also already support it.

@lukasj Would you like me to write this up for the spec?

PR #319, opened 2021-05-05
spec UUID support

gavinking:
Draft support for `UUID`.

See #151 and #152.

Thoughts?

### DIFF
--- a/api/src/main/java/jakarta/persistence/GenerationType.java
+++ b/api/src/main/java/jakarta/persistence/GenerationType.java
@@ -42,7 +42,14 @@ public enum GenerationType {
      * Indicates that the persistence provider must assign 
      * primary keys for the entity using a database identity column.
      */
-    IDENTITY, 
+    IDENTITY,
+
+    /**
+     * Indicates that the persistence provider must assign
+     * primary keys for the entity by generating an RFC 4122
+     * Universally Unique IDentifier.
+     */
+    UUID,
 
     /**
      * Indicates that the persistence provider should pick an 
--- a/spec/src/main/asciidoc/ch02-entities.adoc
+++ b/spec/src/main/asciidoc/ch02-entities.adoc
@@ -160,19 +160,19 @@ accessor methods. However, portable applications must not override the
 object/relational mapping metadata that applies to the persistent fields
 or properties of entity superclasses.
 
-The persistent fields or properties of an
-entity may be of the following types: Java primitive types,
-_java.lang.String_, other Java serializable types (including wrappers
-of the primitive types, _java.math.BigInteger_, _java.math.BigDecimal_
-, _java.util.Date_, _java.util.Calendarfootnote:[Note that an instance
-of Calendar must be fully initialized for the type that it is mapped to.]_,
+The persistent fields or properties of an entity may be of the following
+types: Java primitive types, _java.lang.String_, other Java serializable
+types (including wrappers of the primitive types, _java.math.BigInteger_,
+_java.math.BigDecimal_, _java.util.UUID_, _java.util.Date_,
+_java.util.Calendarfootnote:[Note that an instance of Calendar must be
+fully initialized for the type that it is mapped to.]_,
 _java.sql.Date_, _java.sql.Time_, _java.sql.Timestamp_, _byte[]_,
 _Byte[]_, _char[]_, _Character[]_, _java.time.LocalDate_,
 _java.time.LocalTime_, _java.time.LocalDateTime_,
-_java.time.OffsetTime_, _java.time.OffsetDateTime_, and user-defined
-types that implement the _Serializable_ interface _)_; enums; entity
-types; collections of entity types; embeddable classes (see <<a487>>); collections of
-basic and embeddable types (see <<a494>>).
+_java.time.OffsetTime_, _java.time.OffsetDateTime_,
+and user-defined types that implement the _Serializable_ interface _)_;
+enums; entity types; collections of entity types; embeddable classes
+(see <<a487>>); collections of basic and embeddable types (see <<a494>>).
 
 Object/relational mapping metadata may be
 specified to customize the object/relational mapping and the loading and
@@ -397,9 +397,9 @@ arise when mapping from legacy databases when the database key is
 comprised of several columns. The _EmbeddedId_ or _IdClass_ annotation
 is used to denote a composite primary key. See <<a14687>> and <<a14836>>.
 
-A simple primary key or a field or property
-of a composite primary key should be one of the following types: any
-Java primitive type; any primitive wrapper type; _java.lang.String_;
+A simple primary key or a field or property of a composite primary key
+should be one of the following types: any Java primitive type;
+any primitive wrapper type; _java.lang.String_; _java.util.UUID_;
 _java.util.Date_; _java.sql.Date_; _java.math.BigDecimal_;
 _java.math.BigInteger_.footnote:[In general, however,
 approximate numeric types (e.g., floating point types) should never be
@@ -1259,11 +1259,11 @@ are applied in order:
 the _Embeddable_ annotation, it is mapped in the same way as if the
 field or property were annotated with the _Embedded_ annotation. See
 <<a14634>> and <<a14672>>.
-* If the type of the field or property is one
-of the following, it is mapped in the same way as it would if it were
-annotated as _Basic_: Java primitive types, wrappers of the primitive
-types, _java.lang.String_, _java.math.BigInteger_,
-_java.math.BigDecimal_, _java.util.Date_, _java.util.Calendar_,
+* If the type of the field or property is one of the following, it is
+mapped in the same way as it would if it were annotated as _Basic_:
+Java primitive types, wrappers of the primitive types,
+_java.lang.String_, _java.math.BigInteger_, _java.math.BigDecimal_,
+_java.util.UUID_, _java.util.Date_, _java.util.Calendar_,
 _java.sql.Date_, _java.sql.Time_, _java.sql.Timestamp_,
 _java.time.LocalDate_, _java.time.LocalTime_,
 _java.time.LocalDateTime_, _java.time.OffsetTime_,
--- a/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
+++ b/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
@@ -534,12 +534,12 @@ of mapping to a database column. The _Basic_ annotation can be applied
 to a persistent property or instance variable of any of the following
 types: Java primitive types, wrappers of the primitive types,
 _java.lang.String_, _java.math.BigInteger_, _java.math.BigDecimal_,
-_java.util.Date_, _java.util.Calendar_, _java.sql.Date_,
-_java.sql.Time_, _java.sql.Timestamp_, _java.time.LocalDate_,
-_java.time.LocalTime_, _java.time.LocalDateTime_,
-_java.time.OffsetTime_, _java.time.OffsetDateTime_, _byte[]_,
-_Byte[]_, _char[]_, _Character[]_, enums, and any other type that
-implements _Serializable_.footnote:[Mapping of
+_java.util.UUID_, _java.util.Date_, _java.util.Calendar_,
+_java.sql.Date_, _java.sql.Time_, _java.sql.Timestamp_,
+_java.time.LocalDate_, _java.time.LocalTime_, _java.time.LocalDateTime_,
+_java.time.OffsetTime_, _java.time.OffsetDateTime_,
+_byte[]_, _Byte[]_, _char[]_, _Character[]_, enums, and any other type
+that implements _Serializable_.footnote:[Mapping of
 java.time.LocalDate, java.time.LocalTime, java.time.LocalDateTime,
 java.time.OffsetTime, and java.time.OffsetDateTime types to columns
 other than those supported by the mappings defined by Appendix B of the
@@ -1825,7 +1825,7 @@ defined by the _GenerationType_ enum:
 
 [source,java]
 ----
-public enum GenerationType { TABLE, SEQUENCE, IDENTITY, AUTO };
+public enum GenerationType { TABLE, SEQUENCE, IDENTITY, UUID, AUTO };
 ----
 
 The _TABLE_ generator type value indicates
@@ -1840,13 +1840,17 @@ and IDENTITY are not portable across all databases.]
 The further specification of table generators
 and sequence generators is described in <<a16164>> and <<a16256>>.
 
+The _UUID_ value indicates that the persistence provider should
+assign an RFC 4122 Universally Unique IDentifier.
+
 The _AUTO_ value indicates that the
 persistence provider should pick an appropriate strategy for the
 particular database. The _AUTO_ generation strategy may expect a
 database resource to exist, or it may attempt to create one. A vendor
 may provide documentation on how to create such resources in the event
 that it does not support schema generation or cannot create the schema
-resource at runtime.
+resource at runtime. In the case of a field or property of type
+_java.util.UUID_, the _AUTO_ strategy is equivalent to _UUID_.
 
 This specification does not define the exact
 behavior of these strategies.
@@ -1907,11 +1911,10 @@ The _Id_ annotation specifies the primary key
 property or field of an entity. The _Id_ annotation may be applied in an
 entity or mapped superclass.
 
-The field or property to which the _Id_
-annotation is applied should be one of the following types: any Java
-primitive type; any primitive wrapper type; _java.lang.String_;
-_java.util.Date_; _java.sql.Date_; _java.math.BigDecimal_;
-_java.math.BigInteger_footnote:[Primary keys using
+The field or property to which the _Id_ annotation is applied should be
+one of the following types: any Java primitive type; any primitive wrapper
+type; _java.lang.String_; _java.util.UUID_; _java.util.Date_; _java.sql.Date_;
+_java.math.BigDecimal_; _java.math.BigInteger_footnote:[Primary keys using
 types other than these will not be portable. In general, floating point
 types should never be used in primary keys.]. See <<a132>>.
 
--- a/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
+++ b/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
@@ -617,6 +617,11 @@ protected String name;
 protected String getName() { return name; }
 ----
 
+If the persistence provider stores a value of type _java.util.UUID_ in
+a column of type _VARCHAR_ or equivalent, the value must be stored in
+its canonical representation, unless the application explicitly indicates
+that some other representation is preferred.
+
 ==== Cacheable Annotation
 
 The _Cacheable_ annotation
@@ -1854,6 +1859,12 @@ _java.util.UUID_, the _AUTO_ strategy is equivalent to _UUID_.
 
 This specification does not define the exact
 behavior of these strategies.
+
+However, if the persistence provider stores a value generated according
+to the _UUID_ strategy in a column of type _VARCHAR_ or equivalent, the
+value must be stored in its canonical representation, unless the
+application explicitly indicates that some other representation is
+preferred.
 
 [source,java]
 ----
--- a/api/src/main/resources/jakarta/persistence/orm_3_1.xsd
+++ b/api/src/main/resources/jakarta/persistence/orm_3_1.xsd
@@ -1,989 +1,32 @@
-//
-// Copyright (c) 2017, 2020 Contributors to the Eclipse Foundation
-//
-
-== XML Object/Relational Mapping Descriptor [[a16944]]
-
-The XML object/relational mapping descriptor
-serves as both an alternative to and an overriding mechanism for Java
-language metadata annotations.
-
-=== Use of the XML Descriptor
-
-The XML schema for the object
-relational/mapping descriptor is contained in <<a17175>>. The root element of this
-schema is the _entity-mappings_ element. The absence or present of the
-_xml-mapping-metadata-complete_ subelement contained in the
-_persistence-unit-defaults_ subelement of the _entity-mappings_ element
-controls whether the XML object/relational mapping descriptor is used to
-selectively override annotation values or whether it serves as a
-complete alternative to Java language metadata annotations.
-
-If the _xml-mapping-metadata-complete_
-subelement is specified, the complete set of mapping metadata for the
-persistence unit is contained in the XML mapping files for the
-persistence unit, and any persistence annotations on the classes are
-ignored.
-
-If _xml-mapping-metadata-complete_ is
-specified and XML elements are omitted, the default values apply. These
-default values are the same as the corresponding defaults when
-annotations are used, except in the cases specified in <<a16953>> below. When the
-_xml-mapping-metadata-complete_ element is specified, any
-_metadata-complete_ attributes specified within the _entity_,
-_mapped-superclass_, and _embeddable_ elements are ignored.
-
-If the _xml-mapping-metadata-complete_
-subelement is not specified, the XML descriptor overrides the values set
-or defaulted by the use of annotations, as described below.
-
-The mapping files used by the application
-developer must conform to the XML schema defined in <<a17175>>
-or to the previous version of the XML schema, __orm_2_2.xsd__,
-defined in a previous version of this specification <<a19493>>
-in accordance with the version specified by the _orm.xml_ file.
-
-The Jakarta Persistence persistence provider
-may support use of older versions of the object/relational mapping
-schema as well as the object/relational mapping schema defined in
-<<a17175>>, whether singly or
-in combination when multiple mapping files are used.
-
-=== XML Overriding Rules [[a16953]]
-
-This section defines the rules that apply
-when the XML descriptor is used to override annotations, and the rules
-pertaining to the interaction of XML elements specified as subelements
-of the _persistence-unit-defaults_, _entity-mappings_, _entity_,
-_mapped-superclass_, and _embeddable_ elements.
-
-==== persistence-unit-defaults Subelements
-
-===== schema
-
-The _schema_ subelement applies to all
-entities, tables, secondary tables, join tables, collection tables,
-table generators, and sequence generators in the persistence unit.
-
-The _schema_ subelement is overridden by any
-_schema_ subelement of the _entity-mappings_ element; any _schema_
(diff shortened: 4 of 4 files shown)


gavinking:
Note: I changed only `orm_3_0.xsd`. I'm unclear on whether this should also affect `orm_2_2.xsd`.

api/src/main/resources/jakarta/persistence/orm_3_0.xsd
```diff
@@ -1057,14 +1057,15 @@
...
-        public enum GenerationType { TABLE, SEQUENCE, IDENTITY, AUTO };
+        public enum GenerationType { TABLE, SEQUENCE, IDENTITY, UUID, AUTO };
 
       </xsd:documentation>
     </xsd:annotation>
     <xsd:restriction base="xsd:token">
       <xsd:enumeration value="TABLE"/>
       <xsd:enumeration value="SEQUENCE"/>
       <xsd:enumeration value="IDENTITY"/>
+      <xsd:enumeration value="UUID"/>
```
beikov: I guess this needs a new schema since this is JPA 3.1

api/src/main/java/jakarta/persistence/GenerationType.java
```diff
@@ -42,7 +42,14 @@
...
      */
-    IDENTITY, 
+    IDENTITY,
+
+    /**
+     * Indicates that the persistence provider must assign
+     * primary keys for the entity by generating a RFC 4122
+     * Universally Unique IDentifier.
+     */
+    UUID,
```
lukasj, line 52: Adding a value to enum in java is always backwards incompatible change. Given the strict requirement on semantic versioning for spec projects, we would need to bump spec version to 4.0 - requires new project plan or wait till there is progress review for current version and bump the spec version during that.

api/src/main/resources/jakarta/persistence/orm_3_0.xsd
lukasj: Yes, new schema is required. 2.2/3.0 (and all other already released versions) should be taken as written in the stone with no modifications allowed

lukasj (commented):
In general, the idea and the change is good and goes in the right direction. The only problem I see is the way it is done is not backwards compatible. Would it be possible to make the change in a backwards compatible way?

api/src/main/resources/jakarta/persistence/orm_3_0.xsd
gavinking: OK, I wasn't sure.

gavinking:
> Would it be possible to make the change in a backwards compatible way?

Well, yes, there's two reasonable ways I can see.

The first is a minimal way, but sacrifices functionality: instead of introducing a new enum value, we could just use `AUTO`, and infer UUID generation based on the Java type of the field. The issues with this are:

- that some people might like to use fields of type `String` to hold their UUIDs, and
- asymmetry: UUID is an implicitly second-class citizen compared to the other generation strategies.

The second way is a much bigger change, but addresses a more significant flaw in the current design. Currently `@GeneratedValue` has two things:

- a `strategy`, which is an enumerated value, and
- a `generator`, which is a string-based name of a generator defined elsewhere.

These two things overlap in the sense that you never need to specify both of them.

The thing is that `@GeneratedValue` was designed before any of us had any experience at all programming with annotations, and we got it wrong:

- There's not just three "strategies". Rather, there's an open-ended list of algorithms for generating identifiers, where providers support things beyond the spec, and users often implement their own.
- Unfortunately, using string-based names to refer to generators is untypesafe. And this approach is quite uncomfortable in other ways. For example, I often feel like I don't have a good place to put the `@SequenceGenerator` annotation.

In summary, the current design just doesn't offer any truly clean way for JPA implementors and users to integrate their own custom id generation strategies. And now, apparently, it's even a problem for the JPA spec itself!

A better approach, it seems to me, would be to give `@GeneratedValue` a new member of type `Class`. This is more in line with the approach we introduced in later specs like CDI. A reference to a class is typesafe, and unlike an enum, it models an open-ended list. I'm not sure whether, for this problem, the reference would be to an annotation class or an interface, or to a class class&mdash;we would need to think that through. However, something approximately like the following would be reasonable, it seems to me:

1. Add a new member to `@GeneratedValue`, say `source` or `strategyClass`.
2. Introduce the types `UUID` and `Identity` to represent the built-in UUID and identity column generation strategies.
3. Allow users to specify a type annotated `@SequenceGenerator` or `@TableGenerator` that represents their sequence or table generator.
4. Allow users to specify a type that's interpreted by the JPA provider in some vendor-specific way, giving them a more comfortable way to integrate their own custom id generation strategies, or extensions provided by the vendor.

This solution looks perhaps a little baroque at first, but it actually solves multiple sources of discomfort in a way that's rather future-proof.
 
So one could write:

```java
@Id @GeneratedValue(source=UUID.class)
String id;
```

Or:

```java
@Id @GeneratedValue(source=MyIdTable.class)
long id;

....

@TableGenerator(table="my_ids", allocationSize=100)
interface MyIdTable {}
```

Or

```java
@Id @GeneratedValue(source=MyCustomStategy.class)
MyId id;
```

(In principle it can be made more typesafe by having a marker interface for types representing id generators. Not sure if that's worth doing.)

Thoughts?

gavinking:
A variation of the second option, also worth considering, would be to introduce an `@Generator` meta-annotation, and instead of writing `@GeneratedValue(source=UUID.class)` `@GeneratedValue(source=MyCustomStategy.class)`, you would just write `@UUID` or `@MyCustomStategy`.

Actually perhaps that's better.

rbygrave:
Just a thought in case it helps regarding

> The first is a minimal way, but sacrifices functionality: instead of introducing a new enum value, we could just use AUTO, and infer UUID generation based on the Java type of the field. The issues with this are:

>    that some people might like to use fields of type String to hold their UUIDs, and

As I see it people can already achieve this via a named custom generator. Maybe not super "nice" but it can be done.

>    asymmetry: UUID is an implicitly second-class citizen compared to the other generation strategies.

If I understand this correctly we are saying that because it's being implied via AUTO it isn't an equal with the other strategies that can all be defined nice and explicitly with their own enum value - yes hmmm, not ideal long term.

gavinking:
@rbygrave yes I agree it's possible, I just don't see it as a complete solution, and we would end up having to come back to it in the future.

gavinking:
> Actually perhaps that's better.

Yeah, so scratch what I said earlier, I think it's cleaner to do it via a meta-annotation.

So we would have to:

- introduce a new meta-annotation `@Generator`
- add `@UUID` and `@Identity`, both bearing this annotation 
- weaken the annotation constraint on `@SequenceGenerator` and `@TableGenerator` to allow these annotations to be used as meta-annotations

Then users would annotate their generated fields with a generator annotation instead of `@GeneratedValue`. This is or course just the pattern we used everywhere in CDI. (I should have reached straight for this hammer.)

gavinking:
Excuse me, it would have to be called `@GeneratedUUID` or something like that to avoid the name collision. (Thinking out loud here.)

lukasj:
`@UUID` and `@Identity` => `@UUIDGenerator` and `@IdentityGenerator` to have the naming consistent with `@SequenceGenerator` and `@TableGenerator`?

I like meta-annotation approach

gavinking:
@lukasj Well yeah I considered that option but then didn't like it because the new annotations are describing the generated field, the old (meta-)annotations are describing the generator itself. 

I'm not sure if I'm being clear but what I mean is `@UUIDGenerator UUID id` reads a bit wrong because the field isn't a generator. That's why I suggested `@UUIDGenerated UUID id`. Do you get me?

lukasj:
Yep, I see your point and it makes sense

sebersole:
We are getting a split discussion now between here and the original issue...

As I mentioned there, we have 2 pieces of information we need:

1) What is the database representation we should use
2) How do we generate these values

(1) is relevant whether the UUID is an identifier or a simple basic value.  (2) is ofc only relevant for identifiers.

IMO applying a `@GeneratedValue` to a `UUID`-typed id attribute is pretty straight-forward.  And in fact we might want to strictly specify how these values get generated for portability.

Its not as straight-forward if we want to allow UUID-based generation for a String id attribute.  I mean in Hibernate we already handle this, but I mean in a general way.

Personally I suggest just adding a new `@UUIDGenerator` annotation to be used in conjunction with @GeneratedValue.  E.g.

```
@Id
@GeneratedValue
UUID id;
```

```
@Id
@GeneratedValue
@UUIDGenerator
UUID id;
```
```
@Id
@GeneratedValue
@UUIDGenerator
String id;
```

On the issue, I even suggest a new `@UUID` annotation to control the storage if you want specific handling on the db.

To be clear and fwiw, I am not at all against some form of Hibernate's `IdentifierGenerator` contract making its way in to the spec as well.  That would allow much more flexibility in id values can get generated and allow users to do that in a portable way.  But I also think UUID is universal enough to warrant easy solution

lukasj:
I think it's better to postpone this to 4.0 and have it done properly. What do you think? Also, since this is open here for quite long time already, where to merge this to make it easier for others to experiment with this? Should we branch out 3.x from master and use master for 4.0 this OR should a branch for 4.0 be created and merge changes from there to master at the right time? Note that master is protected (at least one reviewer is required, while other branches are not), so using extra branch for the time being may speed thing up a bit...

gavinking:
Well, I guess I see it like this. There's things we could do for 3.x and there's things which we could do for 4.0. Some of the stuff I've described here would not really fit in a minor release. So let's set all that aside.

What _could_ make sense in a minor release?

1. In principle we could add support for the _most_ minimal thing, requiring support for `@Id @GeneratedValue UUID id` in 3.x. But how much value does this bring? Implementations can already support this if they want (e.g. Hibernate already does). And it would feel pretty strange that there's a `GenerationType` for each of the other built-in strategies, but not for `UUID`.  And it sucks that you can't use `String` to represent your UUIDs. And so it feels like half a solution.
2. The changes I proposed in the original PR are what I would think of as the "minimal thing that feels whole". They have a really low impact, are easy to implement, but put UUID generation on the same footing as the other built-in strategies. Of course, the PR adds `GenerationType.UUID` which in principle breaks binary compatibility.

So it's really a question for you. How inflexibly are we applying the rules around binary compatibility?

My intuition is that adding an enum value to `GenerationType` is a change that's going to break approximately zero programs, and it's something I wouldn't hesitate to do in a minor release of our product. 

If we thought that client programs did a lot of `switch (generationType)`, then sure, adding an enum value would be a nasty breaking change. But I don't think they do. I think the nature of how this `enum` is used suggests that this isn't actually a breaking change.

gavinking:
Note that _nothing_ in the current PR precludes any future extensions that "open up" `@GeneratedValue` to acknowledge the notion that custom id generators are a thing that exist. We can still do that in 4. (And we shouldn't do it in 3.)

lukasj:
> How inflexibly are we applying the rules around binary compatibility?

I'm personally open to changes having minimal impact on clients and preserving binary compatibility (which [is the case here](https://docs.oracle.com/javase/specs/jls/se11/html/jls-13.html#jls-13.4.26)) and therefore I would allow the change in `GenerationType` - the only ones affected clients I can think of here can be some user written code generators and/or IDEs which may be iterating through values of the enum values.

What does @jgrassel or @dazey3 think  about this?

lukasj:
wrt orm schema update - old versions have to stay untouched and new version - 3.1 - needs to be created

lukasj:
@dazey3 or @jgrassel can you share your view on inclusion of this `as is` in 3.1? I've already asked @Tomas-Kraus to prepare a TCK test for this ...

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -160,19 +160,19 @@ accessor methods. However, portable applications must not override the
...
-The persistent fields or properties of an
-entity may be of the following types: Java primitive types,
-_java.lang.String_, other Java serializable types (including wrappers
-of the primitive types, _java.math.BigInteger_, _java.math.BigDecimal_
-, _java.util.Date_, _java.util.Calendarfootnote:[Note that an instance
-of Calendar must be fully initialized for the type that it is mapped to.]_,
+The persistent fields or properties of an entity may be of the following
+types: Java primitive types, _java.lang.String_, other Java serializable
+types (including wrappers of the primitive types, _java.math.BigInteger_,
+_java.math.BigDecimal_, _java.util.UUD_, _java.util.Date_,
```
jgrassel: UUID, not UUD right?

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -1840,13 +1840,17 @@ and IDENTITY are not portable across all databases.]
...
+
 The _AUTO_ value indicates that the
 persistence provider should pick an appropriate strategy for the
 particular database. The _AUTO_ generation strategy may expect a
 database resource to exist, or it may attempt to create one. A vendor
 may provide documentation on how to create such resources in the event
 that it does not support schema generation or cannot create the schema
-resource at runtime.
+resource at runtime. In the case of a field or property of type
+_java.util.UUID_, the _AUTO_ strategy is equivalent to _UUID_.
```
jgrassel, line 1858: I'm wondering about the value of adding a UUID generator type enum value.  It would only be applicable to persistent fields of type UUID, and of little value to integer types.  If, for UUID types, AUTO == UUID, and with AUTO the default already for @GeneratedValue, then I would think that for the purpose of a generated value for a UUID type, just annotating it with @GeneratedValue would be sufficient.

sebersole: Nothing is conceptually incorrect about 

```
@Id
@GeneratedValue
@Uuid
String id;
```

jgrassel:
All of the persistent field types supported by JPA which are not (serializable) application-custom types are mapped by the JDBC specification to a SQL type, like int -> INTEGER, java.time.LocalDate -> DATE, etc.  Looking at the JDBC 4.3 spec, nothing has been settled for java.util.UUID, which kind of makes the mapping a bit of a wild west exercise.  Has there been any discussion about its type mapping among the JDBC standards group?  Because JDBC 4.3 doesn't map the type, were there issues with standardizing a mapping that they couldn't resolve?  I can see there being camps debating whether a UUID should be stored as a String, as a large numeric (if the database supports 128 bit numeric types), or as a raw binary.

I agree with Gavin's assertions about the existing @GeneratedValue annotation.  It is not particularly flexible and future friendly.  Adding a new field to the @GeneratedValue annotation, one of type Class (where the Class implements an Interface, or is loaded with annotations itself) would open a lot of possibilities.  It also opens the door to a lot of extra complexities, like how to provide access to a persistence context's underlying JDBC resources (something an application, if they want to instrument their own version of table or sequence generation, or perhaps invoke a database stored procedure, would need), EntityManager/Persistence Context reentrancy, whether the custom generator should be able to inject CDI resources, etc.  How to deal with generator failures and timeouts (like if the custom generator calls on a restful service to generate an identity and the request fails or times out).

I also wonder just what a "generic" UUID auto-generated value would look like?  Would it just be a call to UUID.randomUUID()?  I don't see that as a terribly valuable generator since that can conceivably create DuplicateKeyExceptions.  I would think that parts of the UUID are application specific, which really necessitates placing the burden of generation as something specific to the application.

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
sebersole: Or, along the lines of what the PR suggests

```
@Id
@GeneratedValue(UUID)
String id;
```

sebersole:
> I also wonder just what a "generic" UUID auto-generated value would look like? (we all know what identity|sequence|table-generated values for integers look like, it's just id++.) Would it just be a call to UUID.randomUUID()? I don't see that as a terribly valuable generator since that can conceivably create DuplicateKeyExceptions.

People absolutely use UUID ids in the real world and there is no real good reason they cannot be generated.  I think we all agree on that.  So the question about how the generation is done is much more of a vendor value-add question in my opinion.

sebersole:
> It also opens the door to a lot of extra complexities

Absolutely agree.  I don't think it makes any sense for the actual generation contract to be defined by the spec (aka, having a new interface)

jgrassel:
> People absolutely use UUID ids in the real world and there is no real good reason they cannot be generated. I think we all agree on that. So the question about how the generation is done is much more of a vendor value-add question in my opinion.

Oh, I know people use UUIDs, that's not what I was talking about.  It was about what kind of algorithm would generate it. 
 UUID.randomUUID() generates a value from the pseudo-random number generator.  All of the supported generators (identity, sequence, table) pretty much guarantee that the primary key generated when the new entity is inserted into the database will not fail because a row with the same identity doesn't already exist in the database.  Even though UUID's namespace is a tremendous 128 bits, a collision between two entities when a pseudo-random number generator is used is a non-zero possibility.  Should that generate a PersistenceException, or generate new random value and retry?

sebersole:
Previously I had said 

> To be clear and fwiw, I am not at all against some form of Hibernate's IdentifierGenerator contract making its way in to the spec as well. That would allow much more flexibility in id values can get generated and allow users to do that in a portable way. But I also think UUID is universal enough to warrant easy solution

So to clarify, its not so much the actual how-do-i-generate-a-value method I have a concern about.  It's the "supporting methods(s)" that would also be needed - how does a sequence-generator register a SEQUENCE for export?  etc

But since they are tied together, I believe we should just not include one

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Thanks, nice catch.

gavinking:
> I was talking about. It was about what kind of algorithm would generate it.

@jgrassel as stated in the proposed spec:

> an **RFC 4122** Universally Unique IDentifier

sebersole:
> All of the supported generators (identity, sequence, table) pretty much guarantee...

That's not up to the spec to decide in my opinion.  Random is fine for probably well over 85% of users.

And if it is not, and your particular provider can't support other algorithms... you can either find a new provider or implement a generator contract that I'm sure every provider has

gavinking:
Of course, implementations might provide additional options beyond RFC 4122. But RFC 4122 is a spec that exists and it makes sense for that to be the default and the thing that the JPA spec talks about.

jgrassel:
> That's not up to the spec to decide in my opinion. Random is fine for probably well over 85% of users.

I would have to wonder, if the go-to algorithm is a random number generator, then what value does a JPA generator offer?  Like I said, all of the JPA generators guarantee a unique id because they either get the new id value from the database itself (identity), or a value from a block of numbers acquired from the sequence/table generator -- values the application can't easily (or possibly) calculate itself, which is why the persistence provider is called to generate them.  If the generator is a simple call to UUID.randomUUID() (which doesn't depend on the contents of the database), then the application can easily do that itself.

jgrassel:
Anyways, the exact algorithm used to generate a UUID aside, there's two items that I'm thinking more about:

1) Database representation of UUID, since the JDBC 4.3 spec doesn't define a mapping, is this something we should pressure them to do?

2) Custom generators -- I think that there is a lot of value to this, well beyond UUID generators.  This could open the door to applications providing their own auto generators, like calling on a restful service or calling a stored procedure (a generator type I think we overlooked when basic stored procedure support was added to JPA).  This should be its own issue separate from UUID, given how far reaching it could become.

gavinking:
> Database representation of UUID, since the JDBC 4.3 spec doesn't define a mapping, is this something we should pressure them to do?

I don't see that as necessary at all. (To be clear, I don't see how it would help.)

What I guess we should do though, that is missing from this PR, is clearly state that when a UUID is persisted to a `varchar` column, then it should be persisted in the canonical representation, that is:

    xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx

jgrassel:
> I don't see that as necessary at all. (To be clear, I don't see how it would help.)

I think it's fairly relevant.  If I'm not mistaken, UUID will be the first Java Type to Database Mapping that is not governed by the JDBC spec.  Even the JDK 8 Java Time types, which we added support for in 2.2, have a JDBC Specification defined mapping.  I agree that varchar is probably the best representation, though we should make sure most, if not all, JPA spec stakeholders are on board with that.

rbygrave:
> I think it's fairly relevant.

I believe it is less relevant than you'd initially think and that is because a bunch of databases have a native UUID type and this is already being done without official JDBC UUID type support. That is, JPA providers will implement this as necessary depending on the database to support the use of database native UUID. They will do that as using database native UUID type is by far preferable over `varchar` or `binary` storage. That is, in some sense varchar and binary storage options are somewhat a fallback option rather than the preference depending on the database.

Said another way, people already have implementations of this without needing special JDBC support for UUID. The actual implementation of this does not rely on the JDBC spec officially supporting UUID (just like JPA implementations didn't _actually_ need support for java time types in JDBC either). 

Proper support in JDBC would be really nice and make life easier but is not strictly required for implementation today. 

> I agree that varchar is probably the best representation

As I read it, Gavin suggested we needed an _agreed representation_ for `varchar` storage and I totally agree with his comment.  Just noting that as I read it I don't think he suggested it was the _"best representation"_.  For what it's worth in terms of already having an implementation there is the logical entity model of java.util.UUID (or java.lang.String), the generator algorithm (V4, V1 etc) and the database representation (native uuid type, varchar, binary, binary optimised [1]).  Generally (99.9% of the time) the generator algorithm and database type are global configuration and depend on the target database.  

[1] binary optimised as per https://www.percona.com/blog/2014/12/19/store-uuid-optimized-way/

With Ebean the database representation option has _explicit_ options (irrespective of whether the database supports uuid natively) or _auto_ options (which prefers database native uuid and has a fallback for databases without native uuid).  So that ends up as:

```java
  /**
   * Specify how UUID is stored.
   */
  public enum DbUuid {

    /**
     * Store using native UUID in H2 and Postgres and otherwise fallback to VARCHAR(40).
     */
    AUTO_VARCHAR(true, false, false),

    /**
     * Store using native UUID in H2 and Postgres and otherwise fallback to BINARY(16).
     */
    AUTO_BINARY(true, true, false),

    /**
     * Store using native UUID in H2 and Postgres and otherwise fallback to BINARY(16) with optimized packing.
     */
    AUTO_BINARY_OPTIMIZED(true, true, true),

    /**
     * Store using DB VARCHAR(40).
     */
    VARCHAR(false, false, false),

    /**
     * Store using DB BINARY(16).
     */
    BINARY(false, true, false),

    /**
     * Store using DB BINARY(16) with optimized packing.
     */
    BINARY_OPTIMIZED(false, true, true);

    boolean nativeType;
    boolean binary;
    boolean binaryOptimized;

    DbUuid(boolean nativeType, boolean binary, boolean binaryOptimized) {
      this.nativeType = nativeType;
      this.binary = binary;
      this.binaryOptimized = binaryOptimized;
    }

```

gavinking:
> As I read it, Gavin suggested we needed an agreed representation for varchar storage and I totally agree with his comment. Just noting that as I read it I don't think he suggested it was the "best representation". 

Yes, that's right. If the db has a dedicated `UUID` column type, then that's great. If it *doesn't*, then the "fallback plan" to `varchar` should be well-defined, so that different JPA implementations are interoperable.

(By the same token, I it seems to me that specifying the algorithm is unnecessary, because the UUID format itself encodes the variant, and so different algorithms interoperate seamlessly.)

lukasj:
Draft of the TCK test for this change is available at https://github.com/eclipse-ee4j/jakartaee-tck/pull/793

api/src/main/resources/jakarta/persistence/orm_3_0.xsd
lukasj: @gavinking can you address this, please?

lukasj:
Are we good with the current state of the proposal or is there anything pending to be resolved before merge of this?

jgrassel:
Did you see my review comment on https://github.com/eclipse-ee4j/jakartaee-tck/pull/793 ?

lukasj:
no, your comment is not visible there. You should submit it somehow. Or just add a regular comment there

lukasj:
yes, it is visible. Should that block this or is it a separate thing to resolve within TCK? I think it is the latter

jgrassel:
I wouldn't call it a blocker.

gavinking:
@lukasj sorry, I've been  busy, and have not kept up with this. Is there anything you need from me here?

gavinking:
> @lukasj sorry, I've been busy, and have not kept up with this. Is there anything you need from me here?

(For example: I think I promised some language to specify how these things are represented in a archer column, right?)

lukasj:
The most important thing is to fix changes to orm schema - no changes to existing ones and create the new one for version 3.1; the diff between 3.0 and 3.1 is supposed to contain licence header, version and the UUID value in the enumeration.

I'll double check it one more time in the morning to see if there is anything else. I'd like to be able to produce RC2 build with these changes to allow completion of related tck test and adoption of this change in implementations. Whatever wording related can be tweaked later through additional PR. Unless there's anyone who disagrees with this flow...

gavinking:
OK, so I'll take a look at that tomorrow, OK?

lukasj:
Sounds good

lukasj:
> What I guess we should do though, that is missing from this PR, is clearly state that when a UUID is persisted to a varchar column, then it should be persisted in the canonical representation [..]

Agree, this needs to be done. As @rbygrave pointed out, some DB vendors do provide native support for this and persistence providers should be free to use such mapping should they want to or are told so by user; so the wording should really be sth like _If UUID is persisted to VARCHAR, then it should be done in UUID's canonical representation_.

gavinking:
Yup.

gavinking:
@lukasj updated, please review.

lukasj (approved):
LGTM (there are few minor things in the schema which I fix in the next PR)

Thank you all for the work!

gavinking:
Yay, excellent, thanks @lukasj!

beikov:
What is the exact expectation now for a persistent attribute like the following with respect to the used SQL type?

```java
@Id
@GeneratedValue(strategy = GenerationType.UUID)
UUID id;
```

Does the JPA spec require the use of a VARCHAR datatype, or is the JPA provider allowed to use a different DDL type like `binary` or a native `uuid` type? The TCK change https://github.com/eclipse-ee4j/jakartaee-tck/commit/feb60a95d4e7133b194efbaae5238fe9186cf277 seems to suggest VARCHAR is expected, but I would like clarification, because Hibernate currently represents such a mapping as native `uuid` if possible and falls back to `binary`. I know that we can change the DDL for TCK runs, but was there a conclusion about a DDL type expectation?

gavinking:
@beikov As far as I know the JPA spec never mandates any particular column type for generated DDL, and that's no different here.

beikov:
Ok, just wanted to make sure since there was talk about cross provider compatibility.

PR #339, opened 2021-12-06
 #264: avoid transitive dependency on java.instrument - v2

lukasj:
Simplified change for #264 

wrt change in the exception: this uses the new TransformerException which extends PersistenceException as I think that is the best approach. The generic `Exception` or `Throwable` can be used as well for better compatibility, or existing `PersistenceException` to just reuse what is available without the need of introducing new class. Which solution do you prefer? Thx!


m0mus (approved):
LGTM

m0mus (approved):
LGTM

api/src/main/java/jakarta/persistence/spi/TransformerException.java
```diff
@@ -11,11 +11,13 @@
...
 package jakarta.persistence.spi;
 
+import jakarta.persistence.PersistenceException;
+
 /**
  *
  * @since 3.1
  */
-public class TransformerException extends Exception {
+public class TransformerException extends PersistenceException {
```
jgrassel: I'm wondering if PersistenceException is the right Exception for the job here.  If a failed transform causes a PersistenceException to be thrown, will it mark the transaction for rollback?  I'm wondering if maybe making the parent of TransformException to be RuntimeException instead might be better.

jgrassel: (saw the PersistenceExecption javadoc update after re-review).  So whether to use RuntimeException instead is my only outstanding question.

jgrassel (changes requested):

api/src/main/java/jakarta/persistence/spi/TransformerException.java
jgrassel: Thinking about it, RuntimeException might not be the best either (burden of dealing with unchecked exceptions with the transformer caller).  But I am rather iffy on using PersistenceException, since its general use is from EntityManager operations, queries, or interactions with the persistence context.  I can't really think of any time PersistenceException is thrown outside of those scenarios.

lukasj: Ok, updated TransformerException to regular (not runtime) ex

jgrassel (approved):
LGTM

PR #340, opened 2021-12-06
bump src level to 11, update build plugins

lukasj:
Apart from updating build plugins, this moves binary level to SE 11 (and drops 8/9 compatibility) as is recommended by the Platform. If there is a reason to NOT move the binary level of the api artifact to SE 11 and keep it at 8/9, speak up, please, and I'll update related part of the change. I'd like to merge this change by Thu Dec 9. Thx


m0mus (approved):
LGTM

PR #341, opened 2021-12-07
Issue 238: Clarify mixing input parameter types for queries

### LINKED ISSUE #238: JPA Spec contradiction on mixing input parameter types
JPA Spec, Section 3.10.11 **Parameter Objects**:
```
The mixing of [javax.persistence.]Parameter objects with named or positional parameters is undefined.
```

JPA Spec, Section 3.10.12 **Named Parameters**:
```
The mixing of named and positional parameters is undefined.
```

JPA Spec, Section 3.10.17.1 **Named Stored Procedure Queries**:
```
If parameter names are not specified, it is assumed that positional parameters are used. 
The mixing of named and positional parameters is undefined
```

JPA Spec, Section 4.6.4: **Input Parameters**:
```
Either positional or named parameters may be used. Positional and named parameters must not be mixed in a single query.
```

---

Sections 3.10.11, 3.10.12, & 3.10.17.1, which govern the _Query_ & _StoredProcedureQuery_ APIs, seem to describe that the mixing of named and positional parameters is not defined by the spec and is open to implementation by the persistence provider.

Section 4.6.4, which governs the JPA query language, seems to describe that mixing positional and named parameters cannot be done in a single query.

These sections seem to be in contradiction. You could read this as positional and named cannot be mixed within a conditional expression, but is undefined anywhere else... but that also seems counter intuitive and messy. 

---

JDBC Spec, 4.1, Section 13.3.2: **Setting Parameters**:
```
It is not possible to combine setting parameters with ordinals and with names in the same statement. 
If ordinals and names are used for parameters in the same statement, an SQLException is thrown.
```

Considering the JDBC Spec prohibits mixing indexed and named parameters in the same statement, I submit that the JPA spec should be updated to prohibit it as well. Alternatively, Section 4.6.4 of the JPA spec should be updated to be in line with the rest of the spec.

dazey3:
fixes #238

Signed-off-by: Will Dazey <dazeydev.3@gmail.com>

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -5426,8 +5426,7 @@ Implementations may support the use of
 _Parameter_ objects for native queries, however support for _Parameter_
 objects with native queries is not required by this specification. The
 use of _Parameter_ objects for native queries will not be portable. The
-mixing of parameter objects with named or positional parameters is
-undefined.
+mixing of parameter objects with named or positional parameters is invalid.
 
 Portable applications should not attempt to
 reuse a _Parameter_ object obtained from a _Query_ or _TypedQuery_
@@ -5443,7 +5442,7 @@ queries that support named parameters.
 Named parameters follow the rules for
 identifiers defined in <<a4760>>.
 Named parameters are case-sensitive. The mixing of named and positional
-parameters is undefined.
+parameters is invalid.
 
 A named parameter of a Jakarta Persistence query
 language query is an identifier that is prefixed by the " _:_ " symbol.
@@ -5918,7 +5917,7 @@ If parameter names are used, the parameter name is used to bind the
 parameter value and to extract the output value (if the parameter is an
 INOUT or OUT parameter). If parameter names are not specified, it is
 assumed that positional parameters are used. The mixing of named and
-positional parameters is undefined.
+positional parameters is invalid.
 
 ===== Dynamically-specified Stored Procedure Queries
 


dazey3:
I decided to go with changing "undefined" with "invalid" because the EntityManager API for `createQuery(String qlString)` states the following:

```
    @throws IllegalArgumentException if the query string is found to be invalid
```

I tested current behavior with OpenJPA, EclipseLink, and Hibernate as well. Both OpenJPA/EclipseLink providers currently throw appropriate IllegalArgumentException from EntityManager.createQuery(...) calls when the query is parsed mixing input parameter types. 
Hibernate though doesn't fail and creates a query. This seems to violate the current `Section 4.6.4: Input Parameters` of the specification, but they aren't the reference implementation.

OpenJPA:
```
org.apache.openjpa.persistence.ArgumentException: Query "SELECT a FROM Story a WHERE a.name = :name AND a.id = ?1" Contains both named and positional parameters this is not allowed by the JPA specification. Detected parameters "[name, 1]".
	at org.apache.openjpa.kernel.jpql.JPQLExpressionBuilder.validateParameters(JPQLExpressionBuilder.java:2462)
...
	at org.apache.openjpa.persistence.EntityManagerImpl.createQuery(EntityManagerImpl.java:102)
```
NOTE: `org.apache.openjpa.persistence.ArgumentException` extends `java.lang.IllegalArgumentException`

EclipseLink:
```
java.lang.IllegalArgumentException: An exception occurred while creating a query in EntityManager: 
Exception Description: Syntax error parsing [SELECT a FROM Story a WHERE a.name = :name AND a.id = ?1]. 
[37, 42] Named and positional input parameters must not be mixed in a single query.
[54, 56] Named and positional input parameters must not be mixed in a single query.
	at org.eclipse.persistence.internal.jpa.EntityManagerImpl.createQuery(EntityManagerImpl.java:1620)
```

beikov:
I don't know how you tested this with Hibernate, but there are compatibility configurations that you must set in order for Hibernate to behave fully spec compliant. I'm pretty sure that if you set those setting, Hibernate will throw an exception as well.

dazey3:
@beikov 

> There are compatibility configurations that you must set in order for Hibernate to behave fully spec compliant. I'm pretty sure that if you set those setting, Hibernate will throw an exception as well.

Hibernate is knowingly not spec compliant by default and you must instead set properties to be JPA spec compliant? I have not heard of this before. Mind sharing a link to the documentation for this?

dazey3:
Regardless, if all providers fail, that adds even more weight that the specification should in fact describe mixing of input parameter types as "invalid" rather than just "undefined"

beikov:
Here are the various JPA compliance related configurations: https://docs.jboss.org/hibernate/orm/5.6/userguide/html_single/Hibernate_User_Guide.html#configurations-jpa-compliance

dazey3:
@beikov 

Testing with the Hibernate "compliance" properties from the documentation you linked did not have a noticeable change in behavior for the simple query test. Hibernate succeeds in creating a query regardless of any mixing of input types.

I understand this Hibernate issue is a side note to this JPA Spec issue, but does Hibernate have the ability to process the property `<property name="hibernate.jpa.compliance" value="true"/>`? It would be convenient to simple say "I want to run with a JPA compliant provider requiring only 1 provider specific property".

sebersole:
> @beikov
> 
> > There are compatibility configurations that you must set in order for Hibernate to behave fully spec compliant. I'm pretty sure that if you set those setting, Hibernate will throw an exception as well.
> 
> Hibernate is knowingly not spec compliant by default and you must instead set properties to be JPA spec compliant? I have not heard of this before. Mind sharing a link to the documentation for this?

Is it really that odd that a provider does things better by default?  I mean really?

sebersole:
> I understand this Hibernate discussion is a side note to this JPA Spec issue, but does Hibernate have the ability to process the property `<property name="hibernate.jpa.compliance" value="true"/>`? It would be convenient to simple say "I want to run with a JPA compliant provider requiring only 1 provider specific property".

Yes, I just added this for 6.0 - https://github.com/hibernate/hibernate-orm/blob/main/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java#L2051

dazey3:
> > @beikov
> > > There are compatibility configurations that you must set in order for Hibernate to behave fully spec compliant. I'm pretty sure that if you set those setting, Hibernate will throw an exception as well.
> > 
> > 
> > Hibernate is knowingly not spec compliant by default and you must instead set properties to be JPA spec compliant? I have not heard of this before. Mind sharing a link to the documentation for this?
> 
> Is it really that odd that a provider does things better by default? I mean really?

"Better" is a matter of opinion. Also, all I said was that I had not heard of this before for Hibernate and asked for a link to the documentation. Did I say something wrong?

sebersole:
> "Better" is a matter of opinion.

Fair enough.  I'll let usage numbers tell the tale of what the general opinion is.

Hibernate existed for many years before JPA.  In fact, most parts of JPA were influenced by Hibernate.  So some of this is "legacy".  Well one - `hibernate.jpa.compliance.list` is strictly a backwards compatibility setting.  Hibernate not only predates JPA, it also predates annotations, and initially there was only XML mapping.  Sometimes that XML bias still comes through.  This List handling is one.  

But honestly, for the most part Hibernate simply does do things better by default in these particular areas.

sebersole:
And we really should not b e having this conversation about Hibernate on the spec GitHub group.  Feel free to ping us one Zulip, etc as outlined at https://hibernate.org if you want to continue discussing

lukasj:
> Is it really that odd that a provider does things better by default? I mean really?

could we leave this sort of comments away from this place and stick to topic, please? IMHO it doesn't matter what the provider does "by default" as long as it satisfies [the TCK rules](https://www.eclipse.org/lists/jakartaee-tck-dev/msg01456.html)

lukasj (approved):
The change looks good to me

sebersole:
> could we leave this sort of comments away from this place and stick to topic, please? IMHO it doesn't matter what the provider does "by default" as long as it satisfies [the TCK rules](https://www.eclipse.org/lists/jakartaee-tck-dev/msg01456.html)

I was answering a direct question asked here.  But, 

> And we really should not be having this conversation about Hibernate on the spec GitHub group...

dazey3:
@lukasj 
Thank you for the review

PR #351, opened 2022-01-18
introduce several "missing" numeric functions

gavinking:
This adds `sign()`, `exp()`, `power()`, `ln()`, `round()`, `floor()`, `ceiling()` to the list of functions portably supported in JPQL. Each of these functions (or something very similar) is available on essentially every SQL database, and are needed in any language which is intended to be used for any sort of nontrivial data analysis. In particular, the natural logarithm and functions for rounding are pretty essential.

I have (against my own preference) *not* included trigonometric functions.

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
+++ b/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
@@ -542,6 +542,14 @@ public interface CriteriaBuilder {
 	
 
     //numerical operations:
+
+    /**
+     * Create an expression that returns the sign of its
+     * argument.
+     * @param x expression
+     * @return sign
+     */
+    Expression<Integer> sign(Expression<? extends Number> x);
 	
     /**
      * Create an expression that returns the arithmetic negation
@@ -558,7 +566,22 @@ public interface CriteriaBuilder {
      * @return absolute value
      */
     <N extends Number> Expression<N> abs(Expression<N> x);
-	
+
+    /**
+     * Create an expression that returns the ceiling of its
+     * argument.
+     * @param x expression
+     * @return ceiling
+     */
+    <N extends Number> Expression<N> ceiling(Expression<N> x);
+
+    /**
+     * Create an expression that returns the floor of its
+     * argument.
+     * @param x expression
+     * @return floor
+     */
+    <N extends Number> Expression<N> floor(Expression<N> x);
     /**
      * Create an expression that returns the sum
      * of its arguments.
@@ -702,7 +725,51 @@ public interface CriteriaBuilder {
      */	
     Expression<Double> sqrt(Expression<? extends Number> x);
 
-	
+    /**
+     * Create an expression that returns the exponential
+     * of its argument.
+     * @param x expression
+     * @return exponential
+     */
+    Expression<Double> exp(Expression<? extends Number> x);
+
+    /**
+     * Create an expression that returns the natural logarithm
+     * of its argument.
+     * @param x expression
+     * @return natural logarithm
+     */
+    Expression<Double> ln(Expression<? extends Number> x);
+
+    /**
+     * Create an expression that returns the first argument
+     * raised to the power of its second argument.
+     * @param x base
+     * @param y exponent
+     * @return the base raised to the power of the exponent
+     */
+    Expression<Double> power(Expression<? extends Number> x, Expression<? extends Number> y);
+
+    /**
+     * Create an expression that returns the first argument
+     * raised to the power of its second argument.
+     * @param x base
+     * @param y exponent
+     * @return the base raised to the power of the exponent
+     */
+    Expression<Double> power(Expression<? extends Number> x, Number y);
+
+    /**
+     * Create an expression that returns the first argument
+     * rounded to the number of decimal places given by the
+     * second argument.
+     * @param x base
+     * @param n number of decimal places
+     * @return the rounded value
+     */
+    <T extends Number> Expression<T> round(Expression<T> x, Integer n);
+
+
     //typecasts:
     
     /**
--- a/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
+++ b/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
@@ -545,7 +545,9 @@ public interface CriteriaBuilder {
 
     /**
      * Create an expression that returns the sign of its
-     * argument.
+     * argument, that is, {@code 1} if its argument is
+     * positive, {@code -1} if its argument is negative,
+     * or {@code 0} if its argument is exactly zero.
      * @param x expression
      * @return sign
      */
@@ -569,7 +571,8 @@ public interface CriteriaBuilder {
 
     /**
      * Create an expression that returns the ceiling of its
-     * argument.
+     * argument, that is, the smallest integer greater than
+     * or equal to its argument.
      * @param x expression
      * @return ceiling
      */
@@ -577,7 +580,8 @@ public interface CriteriaBuilder {
 
     /**
      * Create an expression that returns the floor of its
-     * argument.
+     * argument, that is, the largest integer smaller than
+     * or equal to its argument.
      * @param x expression
      * @return floor
      */
@@ -692,7 +696,8 @@ public interface CriteriaBuilder {
 	
     /**
      * Create an expression that returns the modulus
-     * of its arguments.
+     * (remainder under integer division) of its
+     * arguments.
      * @param x expression
      * @param y expression
      * @return modulus
@@ -701,7 +706,8 @@ public interface CriteriaBuilder {
 	
     /**
      * Create an expression that returns the modulus
-     * of its arguments.
+     * (remainder under integer division) of its
+     * arguments.
      * @param x expression
      * @param y value
      * @return modulus
@@ -710,7 +716,8 @@ public interface CriteriaBuilder {
 
     /**
      * Create an expression that returns the modulus
-     * of its arguments.
+     * (remainder under integer division) of its
+     * arguments.
      * @param x value
      * @param y expression
      * @return modulus
@@ -727,7 +734,8 @@ public interface CriteriaBuilder {
 
     /**
      * Create an expression that returns the exponential
-     * of its argument.
+     * of its argument, that is, Euler's number <i>e</i>
+     * raised to the power of its argument.
      * @param x expression
      * @return exponential
      */
--- a/spec/src/main/asciidoc/ch04-query-language.adoc
+++ b/spec/src/main/asciidoc/ch04-query-language.adoc
@@ -314,18 +314,19 @@ which the method _Character.isJavaIdentifierPart_ returns true. The
 question mark (_?_) character is reserved for use by the Jakarta
 Persistence query language.
 
-The following are reserved identifiers: ABS,
-ALL, AND, ANY, AS, ASC, AVG, BETWEEN, BIT_LENGTHfootnote:[BIT_LENGTH,
-CHAR_LENGTH, CHARACTER_LENGTH, POSITION, and UNKNOWN are not currently
-used: they are reserved for future use.],
-BOTH, BY, CASE, CHAR_LENGTH, CHARACTER_LENGTH, CLASS, COALESCE, CONCAT,
-COUNT, CURRENT_DATE, CURRENT_TIME, CURRENT_TIMESTAMP, DELETE, DESC,
-DISTINCT, ELSE, EMPTY, END, ENTRY, ESCAPE, EXISTS, EXTRACT, FALSE, FETCH,
-FROM, FUNCTION, GROUP, HAVING, IN, INDEX, INNER, IS, JOIN, KEY, LEADING,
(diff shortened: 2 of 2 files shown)


spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -314,18 +314,19 @@ which the method _Character.isJavaIdentifierPart_ returns true. The
...
+AVG, BETWEEN, BIT_LENGTH, BOTH, BY, CASE, CEILING, CHAR_LENGTH,
+CHARACTER_LENGTH, CLASS, COALESCE, CONCAT, COUNT, CURRENT_DATE,
+CURRENT_TIME, CURRENT_TIMESTAMP, DELETE, DESC, DISTINCT, ELSE, EMPTY,
+END, ENTRY, ESCAPE, EXISTS, EXP, EXTRACT, FALSE, FETCH, FLOOR, FROM,
+FUNCTION, GROUP, HAVING, IN, INDEX, INNER, IS, JOIN, KEY, LEADING,
+LEFT, LENGTH, LIKE, LN, LOCATE, LOWER, MAX, MEMBER, MIN, MOD, NEW,
+NOT, NULL, NULLIF, OBJECT, OF, ON, OR, ORDER, OUTER, POSITION, POWER,
+ROUND, SELECT, SET, SIGN, SIZE, SOME, SQRT, SUBSTRING, SUM, THEN,
+TRAILING, TREAT, TRIM, TRUE, TYPE, UNKNOWN, UPDATE, UPPER, VALUE,
+WHEN, WHERE.
```
gavinking: As a tangential comment, I think it's much better to _not_ include function names in the list of keywords. It's perfectly possible to parse JPQL while treating the function names as regular identifiers, just as you would do in any normal programming language.

gavinking: And I don't really know how or why `BIT_LENGTH`, `CHAR_LENGTH`, `CHARACTER_LENGTH`,
 `POSITION`, and `UNKNOWN` got onto this list (perhaps it was guff left over from EJB-QL pre 3.0), but I think they should simply be removed.

lukasj: since they are just reserved for the future use, let's keep them there for now

lukasj (commented):
Would it make sense to add these functions to the `CriteriaBuilder` as well?

gavinking:
> Would it make sense to add these functions to the `CriteriaBuilder` as well?

Yes, of course, indeed it does, and I'll add that to the PR.

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: I don't really care, just mentioning it because in a couple of cases I can't imagine what the future use might be: `CHAR_LENGTH`/`CHARACTER_LENGTH` in SQL are always just synonyms for JPA's `LENGTH` AFAIK.

And sure, we might in principle one day add a `POSITION` function, even though it's *really* similar to the existing `LOCATE`, but why is `POSITION` called out explicitly in that list and not `OVERLAY`?

And ... `UNKNOWN`? I can't even really speculate what that's doing there or what it might be useful for.

Anyway, it's not an important point.

lukasj: so perhaps mark them for the removal in the next update/release? I mean simply removing them right away does not look as the right thing to do.

gavinking: OK, your call.

Just removing them would not break anything though: there's no way any code can possibly be relying on them being illegal identifiers, right?

lukasj: Correct, yet it's still removal.

gavinking: Well, it's removal of a _limitation_.

gavinking:
> > Would it make sense to add these functions to the `CriteriaBuilder` as well?
> 
> Yes, of course, indeed it does, and I'll add that to the PR.

Done.

gavinking:
@lukasj do you want me to also add `extract()` to `CriteriaBuilder`? It's _slightly_ more complex because I'll have to add an enum with the supported field types, but other than that, easy enough.

(I would do it in a separate PR though.)

gavinking:
>  It's slightly more complex because I'll have to add an enum with the supported field types, but other than that, easy enough.

Hrm, in fact, it's a little tricky because not all the field types have the same result type, but it's still doable.

pzygielo:
Would it be possible to have method names consistent with the ones from `java.lang.Math`?

gavinking:
> (I would do it in a separate PR though.)

See https://github.com/eclipse-ee4j/jpa-api/pull/356/files for a rough draft.

gavinking:
> Would it be possible to have method names consistent with the ones from `java.lang.Math`?

I mean, it's certainly *possible*, though I'm unclear on why that would be advantageous.

Speaking for myself, as  a general principle I always avoid abbreviations unless the abbreviated usage is really completely ingrained, as is the case with `avg()`, for example. 

It's not that case that `pow()` and `ceil()` are "ingrained" abbreviations: `pow()` isn't used in any dialect of SQL, and `ceil()` is used in only a minority of dialects.

In particular, I very deliberately avoided `log()` instead of `ln()` because in many dialects of SQL, and in engineering, and often even in _physics_, `log()` means the base-10 logarithm. Now, I was educated as a mathematician so I write the natural logarithm as `log()`, but even in math it's recognized as ambiguous. So I think it's much better to use `ln()` here.

gavinking:
(I should add, because perhaps it's not completely obvious, that the list of functions defined by JPQL is inspired much more by SQL than by Java.)

pzygielo:
> > Would it be possible to have method names consistent with the ones from `java.lang.Math`?
> 
> I mean, it's certainly _possible_, though I'm unclear on why that would be advantageous.

> And on the other hand, I very deliberately avoided `log()` instead of `ln()` because in many dialects of SQL, and in engineering, and often even in _physics_, `log()` means the base-10 logarithm. Now, I was educated as a mathematician so I write the natural logarithm as `log()`, but even in math it's recognized as ambiguous. So I think it's much better to use `ln()` here.

By methods I meant methods in `CriteriaBuilder`.
Is its implementation expected to be used to write native SQL queries or rather be used in Java program?
If the second - I'd just prefer to not re-learn new names.

-----
As for `log` only - it's not just in  `j.l.Math`:

C:
```
NAME
       log, logf, logl - natural logarithmic function

SYNOPSIS
       #include <math.h>

       double log(double x);
       float logf(float x);
       long double logl(long double x);

DESCRIPTION
       These functions return the natural logarithm of x.
```

Python:
```

math.log(x[, base])

    With one argument, return the natural logarithm of x (to base e).
```

Ruby:
```
 log(x) → Float
log(x, base) → Float

Returns the logarithm of x. If additional second argument is given, it will be the base of logarithm. Otherwise it is e (for the natural logarithm).
```

Haskell, ES6 - the same.

`bc` doesn't use `ln` neither.

Perhaps there are programming languages that do.

-----

Maybe it's time and place to go with `ln`. Or have methods inconsistent with `j.l.Math`.
Just saying - I don't like it.

gavinking:
> Is its implementation expected to be used to write native SQL queries or rather be used in Java program?

The philosophy of criteria queries in JPA is very much to be a faithful reification of JPQL in Java objects. The methods of `CriteriaBuilder` have the same names as the built-in JPQL functions.

Since JPQL is substantially based on _SQL_ (not Java!), the functions are modeled after SQL functions, and so that's where the methods of `CriteriaBuilder` ultimately derive from.

Since `log()` is quite ambiguous in SQL (for example, `log()` in PostgreSQL means the base-10 logarithm), I was careful to pick a name that doesn't carry any possibility of confusion.

pzygielo:
@gavinking - thank you for details/reasoning provided.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -702,7 +725,51 @@
      */	
     Expression<Double> sqrt(Expression<? extends Number> x);
 
-	
+    /**
+     * Create an expression that returns the exponential
+     * of its argument.
```
Tomas-Kraus: Put a note, that base is `e` so it computes `e^x`.

gavinking: See: 6b986e8ccb20e808ac52783f3e4ba645a20ffe25

I've also clarified the semantics of `mod()`, since "modulus" is also used to mean the absolute value.

Tomas-Kraus: OK, thanks.

Tomas-Kraus:
Also we may need few JPA TCK tests to verify those API changes. :) Are there any volunteers?

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -702,7 +732,52 @@
...
+
+    /**
+     * Create an expression that returns the first argument
+     * rounded to the number of decimal places given by the
+     * second argument.
+     * @param x base
+     * @param n number of decimal places
+     * @return the rounded value
+     */
+    <T extends Number> Expression<T> round(Expression<T> x, Integer n);
```
Tomas-Kraus, line 778: This can be added too:
```
<T extends Number> Expression<T> round(Expression<T> x);
<T extends Number> Expression<T> round(Expression<T> x, Expression<Integer> n);
```

gavinking: I mean, it can certainly add it, but I'm kinda struggling I bit to see what someone would use it for...

Tomas-Kraus: At least 1st one makes sense. I have some issues with testing this on PostgreSQL - https://www.postgresql.org/docs/9.5/functions-math.html
There are 2 prototypes:
- round(dp or numeric) - this is the only one that supports double precision
- round(v numeric, s int) - numeric values only, throws `Internal Exception: org.postgresql.util.PSQLException: ERROR: function round(double precision, integer) does not exist` for Java Double in entity.

Databases usually have ROUND function with single argument.

2nd prototype makes sense when both values will come from Entity/table, like in my jUnit for POWER:
```
public class NumberEntity {
    @Id
    private Integer id;
    private Long longValue;
    private Double doubleValue;
   ...
}

        CriteriaBuilder cb = em.getCriteriaBuilder();
        CriteriaQuery<Double> cq = cb.createQuery(Double.class);
        Root<NumberEntity> number = cq.from(NumberEntity.class);
        cq.select(cb.power(number.get("doubleValue"), number.get("longValue")));
        cq.where(cb.equal(number.get("id"), id));
        return em.createQuery(cq).getSingleResult();
```
but maybe this use-case is useless.

gavinking: > At least 1st one makes sense. 

Oh, sorry, I was responding to the email notification before your edit. I agree that the first one is reasonable.

> I have some issues with testing this on PostgreSQL

OK, I believe a reasonable implementation is to translate it to something like:

    cast(d+0.5 as decimal(17,2))

(There might be a better way, I'm not certain.)

_UPDATE_: this isn't the best way: `floor()` is better.

> but maybe this use-case is useless.

I mean, that would be my _gut feeling_, but I'm not hung up on it...

gavinking: Actually perhaps `cast(cast(d+0.5 as decimal(20,2)) as double precision)` is better in some contexts.

gavinking: Or you can use `floor()`.

gavinking:
> * Calling ln(n) for n<=0 returns null on some databases (MySQL) and throws an exception on others (Derby,Postgres)

Interesting, I had not noticed this. 

In principle we could mandate the MySQL-like behavior, by translating `ln(x)` to `ln(case when d<0 then null else d end)` on Postgres and Derby, but I'm not sure it's worth the effort.

I would class this difference in behavior as fairly harmless, but I'm interested to hear what others think.

> Derby DB does not have `POWER` and `ROUND` functions so they should not be mandatory

I'm *totally* fine with making these things non-mandatory, but FYI, in our implementation we emulate 

- `power()` with `exp()` and `ln()`, 
- and `round()` using `floor()`
- 
-  on Derby.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
gavinking: Alright, so I just tested this across all platforms we support, and this is indeed an issue very specific to Postgres. I think the most reasonable solution is to emulate `round(x,n)` using `floor(x*10^n+0.5)/10^n` on Postgres, at least when the argument is a double precision type.

lukasj (approved):
LGTM, thanks!

Tomas-Kraus:
@gavinking  Yes, `floor(x*10^n+0.5)/10^n` works fine on Postgres.

PR #352, opened 2022-01-18
introduce local date/time special functions

gavinking:
This finally puts support for the `LocalDate`, `LocalTime`, and `LocalDateTime` types defined by `java.time` on an equal footing to support for the bad old `Date`,`Time`,`Timestamp` classes in `java.sql`.

It also clarifies the types of `current_date` and friends, which were left undefined until now, but which were interpreted by implementations to mean the `java.sql` types.


lukasj (approved):
LGTM

gavinking:
Thank you!

PR #354, opened 2022-01-30
fix the definition of EXTRACT(SECOND...) 

gavinking:
`EXTRACT(SECOND...)` should return a floating point value because truncating fractional seconds isn't useful.


lukasj (approved):
LGTM

gavinking:
Thanks @lukasj, and apologies for messing up here.

PR #379, opened 2022-09-08
draft spec for record support

gavinking:
Allows `@IdClass and` `@Embeddable` types to be records.


lukasj (approved):

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1130,12 +1132,17 @@ values. Embedded objects belong strictly to their owning entity, and are
...
+Embeddable classes must be annotated as
 _Embeddable_ or denoted in the XML descriptor as such. The access type
 for an embedded object is determined as described in <<a113>>.
 
+An embeddable class may be a regular Java class which adheres to the
+requirements specified in <<a18>> for entities, with the exception that
+an embeddable class is not annotated as _Entity_.
+
+Alternatively, an embeddable class may be a top-level Java record type,
+in which case the restrictions that apply to entity classes are relaxed.
```
gavinking: Change to:

> Aternatively, an embeddable class may be any Java record type.

lukasj (commented):
records are compiled into static classes, so there is no reason to limit their usage to top-level definitions.

it would be also good to reword the _restrictions that apply to entity classes are relaxed_ part to make it clear what is or is not applicable (probably enough to just remove this part)

gavinking:
> records are compiled into static classes, so there is no reason to limit their usage to top-level definitions.
> 
> it would be also good to reword the _restrictions that apply to entity classes are relaxed_ part to make it clear what is or is not applicable (probably enough to just remove this part)

I've made both these changes. I suppose this is ready to merge.

lukasj (approved):

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1130,12 +1132,16 @@ values. Embedded objects belong strictly to their owning entity, and are
...
 
-Embeddable classes must adhere to the
-requirements specified in <<a18>> for entities with the exception that embeddable classes
-are not annotated as _Entity_. Embeddable classes must be annotated as
+Embeddable classes must be annotated as
 _Embeddable_ or denoted in the XML descriptor as such. The access type
 for an embedded object is determined as described in <<a113>>.
 
+An embeddable class may be a regular Java class which adheres to the
+requirements specified in <<a18>> for entities, with the exception that
```
trajano, line 1140: @gavinking  not sure where a18 is but does it specify that it needs to be `Serializable`?  Entities are not required to be `Serializable`.

PR #386, opened 2022-10-22
Fixes the example codes in AttributeOverrides javadoc

hantsy:
Add missing `name` attribute to the example codes in AttributeOverrides javadoc.

### DIFF
--- a/api/src/main/java/jakarta/persistence/AttributeOverrides.java
+++ b/api/src/main/java/jakarta/persistence/AttributeOverrides.java
@@ -32,10 +32,10 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  *
  *    &#064;Embedded
  *    &#064;AttributeOverrides({
- *            &#064;AttributeOverride(name="startDate", 
- *                               column=&#064;Column("EMP_START")),
- *            &#064;AttributeOverride(name="endDate", 
- *                               column=&#064;Column("EMP_END"))
+ *            &#064;AttributeOverride(name="startDate",
+ *                               column=&#064;Column(name="EMP_START")),
+ *            &#064;AttributeOverride(name="endDate",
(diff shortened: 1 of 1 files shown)


dmatej (approved):

dmatej:
Interesting, it was correct in AttributeOverride, but the name was missing in AttributeOverrides annotation.

lukasj:
would the one merging this mind to fix the build? See https://www.eclipse.org/lists/jpa-dev/msg00160.html for details. Thanks

hantsy:
@lukasj It seems to need to update the copyright year.

lukasj:
right. I'd expect either the author of this PR or the person who merged this to fix it (read as "provides the PR fixing this").

dmatej:
> right. I'd expect either the author of this PR or the person who merged this to fix it (read as "provides the PR fixing this").

I hear you, that always happens when I believe it is a sure simple thing so I don't need to run the build locally :-D
Or better build system should verify pull requests ...

hantsy:
@dmatej Why not consider Github actions to build the project, thus it will forbid PR to merge if there are build error in the build workflow.

dmatej:
> @dmatej Why not consider Github actions to build the project, thus it will forbid PR to merge if there are build error in the build workflow.

It seems that it is already configured, but as you were not a committer of the project, the build did not start automatically and nothing warned me before the merge that I should have initiated the build.
For my pr it started automatically.
Next time we will do that better 🙂

lukasj:
the problem is in a way the gh action - the checkout part there in particular - does work. The GF copyright plugin sees *all* files as modified in a year at which the checkout has been done, even those unchanged by the recent commit. To avoid this false alarms, there is `-Dcopyright.ignoreyear=true` option in the GH action build for now. Unfortunately I haven't found a way to get around this problem (yet?).

Issue #408, opened 2023-05-18
concat() is a variadic function, but declared binary by CriteriaBuilder 


gavinking:
In `CriteriaBuilder` we have:

```java
Expression<String> concat(Expression<String> x, Expression<String> y);
Expression<String> concat(Expression<String> x, String y);
Expression<String> concat(String x, Expression<String> y);
```

we should add:

```java
Expression<String> concat(Expression<String>... strings);
```

PR #412, opened 2023-05-19
add variadic overload of concat() to CriteriaBuilder

gavinking:
See #408.


api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -1074,6 +1074,15 @@ public interface CriteriaBuilder {
...
     Predicate notLike(Expression<String> x, String pattern, char escapeChar);
 
+    /**
+     *  Create an expression for string concatenation.
+     *  If the given list of expressions is empty, returns
+     *  an expression equivalent to {@code literal("")}.
+     *  @param expressions  string expressions
+     *  @return expression corresponding to concatenation
+     */
+    Expression<String> concat(Expression<String>... expressions);
```
lukasj: this triggers `Possible heap pollution from parameterized vararg type jakarta.persistence.criteria.Expression<java.lang.String>` compile-time warning - can we do something about it?

gavinking: Unfortunately not. It's the world's most useless compiler warning, and unfortunately you can't put `@SafeVarargs` on an interface method :-(

gavinking: I suppose my least worst option is to change this to:

    Expression<String> concat(List<Expression<String>> expressions);

so people can use `concat(List.of(....))`

lukasj: I'd consider usage of `@SuppressWarnings` to be sufficient in this particular case

gavinking: Unfortunately that doesn’t eliminate the warning on the *caller side*, which is the one I’m much more bothered by, because that’s in user code.

zimmi: > I suppose my least worst option is to change this to:
> 
> ```
> Expression<String> concat(List<Expression<String>> expressions);
> ```
> 
> so people can use `concat(List.of(....))`

+1 to the `List` overload from me. The Criteria API is mostly used to construct queries programmatically. As soon as the number of expressions is only known at runtime (not unlikely considering the use case of the API) users will have a `List` anyway.

gavinking: @zimmi Yes, that's true. In fact it can be a PITA to pass an array instead of a `List`.

lukasj: Yes, let's move on with the `Expression<String> concat(List<Expression<String>> expressions);` signature

gavinking: Done.

gavinking:
> * change to `List<Expression<String>>`
> * add a note to changelog

I've made both changes. I believe this is ready to merge.

lukasj (approved):

Issue #410, opened 2023-05-19
add methods for managing Session/Transaction lifecycle to SessionFactory


gavinking:
The transaction management idiom in JPA is not so trivial:

```java
var entityManager = factory.createEntityManager();
var transaction = entityManager.getTransaction();
try {
    transaction.begin();
    // do some work...
    transaction.commit();
}
catch (Exception e) {
    if (transaction.isActive()) {
        try {
            transaction.rollback();
        }
        catch (Exception x) {
            e.addSuppressed(x);
        }
   }
   throw e; 
}
finally {
    entityManager.close();
} 
```

I think we should add a `withTransaction()` method to `EntityManagerFactory` declared as follows:

```java
<R> R withTransaction(Function<EntityManager,R> work);
```

The `work` would be performed within the scope of a resource-local transaction.

rbygrave:
I presume the alternative of using try-with-resources blocks is not workable?  Seems it could be done for EntityManager but problematic to get a transaction.close() in place?

gavinking:
Folks, any feedback on https://github.com/jakartaee/persistence/pull/414?

One issue here is whether we need two flavors of `withTransaction()`, `void`, and one returning a value.

I think we _do_, but I don't know what to call them.

gavinking:
@hantsy @Tomas-Kraus @lukasj and everyone else:

1. Are there any votes for renaming to `executeInTransaction()` + `computeInTransaction()`?
2. Alternatively, would anyone be in favor of just calling these methods `run()` and `call()` or `execute()` and `compute()`, and letting you pass in a `TxType.REQUIRED`, `TxType.REQUIRES_NEW`, etc.

So according to 2, you would write:

```java
entityManagerFactory.execute(TxType.REQUIRED, entityManager -> {
    entityManager.persist(book);
});
```

which is obviously quite a lot harder to type than:

```java
entityManagerFactory.executeInTransaction(entityManager -> {
    entityManager.persist(book);
});
```

But it's not _actually_ that much worse to look at, and it's quite a lot more flexible.

Another argument agains 2 is that the way `java.transaction` actually defines this enum is IMO pretty nasty:

- for some reason I cannot for the life of me understand, they defined the enum `TxType` as an inner type of the annotation type `@Transactional`, making it much less naturally-reusable, and
- `TxType` doesn't read well at all ... not to mention that it's not a "transaction type", but a transaction _propagation_ type.

[...]

PR #414, opened 2023-05-19
Add EntityManagerFactory.runInTransaction()/callInTransaction()

gavinking:
See #410.

This one is just a rough draft.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
+++ b/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
@@ -11,12 +11,16 @@
  */
 
 // Contributors:
+//     Gavin King      - 3.2
 //     Linda DeMichiel - 2.1
 //     Linda DeMichiel - 2.0
 
 package jakarta.persistence;
 
 import java.util.Map;
+import java.util.function.Consumer;
+import java.util.function.Function;
+
 import jakarta.persistence.metamodel.Metamodel;
 import jakarta.persistence.criteria.CriteriaBuilder;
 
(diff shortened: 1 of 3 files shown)


dmatej:
Perhaps yet add a documentation "copy pasted" from the getTransaction method about the JTA:
```
     * @throws IllegalStateException if invoked on a JTA entity manager
```

gavinking:
@dmatej do we really want to *require* implementations to do that though?

I'm not so sure about that.

gavinking:
I've updated the PR with an overloaded version of `withTransaction()`.

Tomas-Kraus:
Hi Gavin, I like this PR. But it may be done a bit better. :) I was designing identical pattern for Oracle Helidon just few weeks ago. The only difference between my and yours proposal are 2 more methods.
Check transaction methods in https://github.com/Tomas-Kraus/helidon/blob/helidon-data/data/data/src/main/java/io/helidon/data/HelidonData.java
and POC implementation code in https://github.com/Tomas-Kraus/helidon/blob/helidon-data/data/jpa/runtime/src/main/java/io/helidon/data/jpa/runtime/JpaHelidonData.java
The only difference between Callable and Function/Consumer is how the transaction is handled. Callable does it for you (rollback on any Throwable, commit otherwise), in Function/Consumer it's user response to handle the transaction.

So 1st note: Function/Consumer have

lukasj:
would it make sense to make these additional methods `default`?

Tomas-Kraus:
Here is sample I made to verify this change in Eclipselink: https://github.com/Tomas-Kraus/jpa-api/blob/emf_transaction/api/src/main/java/jakarta/persistence/EntityManagerFactory.java#L227
So default implementation may not be a problem.

gavinking:
@Tomas-Kraus so the essential differences with your proposal are that:

1. we also accept work that throws a checked exception, and
2. we also give you the option of receiving a reference to the `EntityTransaction` in the work lambda?

Is that right?

I'm not sure what I thin about 2. If we're managing the transaction for you, what are you going to do with the reference to the  `EntityTransaction`? It seems like you would not be allowed to call any of its methods?

gavinking:
(P.S. Apologies for not noticing the comments on the PR. For some reason I did not get notified.)

Tomas-Kraus:
@gavinking Looks like my link is broken, here is another one: https://github.com/Tomas-Kraus/jpa-api-gavin/blob/af4248c1cdbc12fc0f72d59d84795846485944b6/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
There is default implementation in the methods.

1. Transaction for `TransactionVoidWork`/`TransactionWork` is handled by JPA so there shall be checked exception and no user access to transaction handler is needed. Maybe `catch (Throwable e)` would be better in the  implementation to rollback on any `Throwable` that may happen.

2. Transaction for `BiConsumer`/`BiFunction` is **not** handled by JPA, only `begin()` is called before running user code. that's why EntityTransaction instance is passed to the user. It's user's responsibility to do commit/rollback so JPA does not need to know about any checked exceptions.

I added both options because there are use cases when users would like to have control over the transaction.

lukasj (commented):
Add a definition also for JTA EMs to handle:
* no transaction
* existing transaction

gavinking:
> Add a definition also for JTA EMs to handle:
> 
> * no transaction
> * existing transaction

I have updated the proposal with this and other fixes. Please take a look.

Tomas-Kraus:
I had some conversation with Lukas about his reviews while I was on vacation so I'm synced with current status. It looks good now.

gavinking:
Awesome, thanks @Tomas-Kraus!

lukasj (approved):
LGTM

Issue #383, opened 2022-09-08
typesafe way to pass options and an EntityGraph to find()


gavinking:
Currently, setting the `CacheStoreMode` or `CacheRetrieveMode` requires the use of string typing, the `jakarta.persistence.cache.retrieveMode` or `jakarta.persistence.cache.storeMode` properties.

This is obviously bad, and therefore:

- We should add `setCacheStoreMode()` and `setCacheRetrieveMode()` to `EntityManager` and to `Query`.
- It might even be worth introducing a superinterface of `CacheStoreMode`, `CacheRetrieveMode`, and `LockModeType`, calling it `FindOption`, for example, along with new overloads of `find()` and `refresh()` with signature `T find(Class<T> entityClass, Object primaryKey, FindOption... options)`.

gavinking:
There's a similar, related problem with `jakarta.persistence.loadgraph` and `jakarta.persistence.fetchgraph`, though those are a little trickier.

gavinking:
> It might even be worth introducing a superinterface of CacheStoreMode, CacheRetrieveMode, and LockModeType, calling it FindOption, for example, along with new overloads of find() and refresh() with signature T find(Class<T> entityClass, Object primaryKey, FindOption... lockMode).



> * It might even be worth introducing a superinterface of `CacheStoreMode`, `CacheRetrieveMode`, and `LockModeType`, calling it `FindOption`, for example, along with new overloads of `find()` and `refresh()` with signature `T find(Class<T> entityClass, Object primaryKey, FindOption... lockMode)`.

Actually, a better design would probably be a `FindOptions` class which lets you specify:

- the cache modes,
- the lock mode, 
- the `jakarta.persistence.lock.timeout` and `PessimisticLockScope`, and/or
- a load graph or fetch graph,

and then add an overload with the signature:

    T find(Class<T> entityClass, Object primaryKey, FindOptions options)

So this `FindOptions` class would just be a typesafe replacement for the `Map` argument of `find()`.

gavinking:
So, after quite a lot of reflection on this, I'm now inclined to go for something much more like how Hibernate does this, but with cleaned-up naming.

I generally prefer `new`ing objects to "fluent" APIs, but I think in this case the fluent approach has advantages.

You see, the tricky part of this is passing the `EntityGraph` in a typesafe way, without adding verbosity and/or complexity, and the fluent approach actually handles that really nicely.

Consider:

```java
[...]

PR #415, opened 2023-05-25
Typesafe operations for controlling find() queries

gavinking:
See #383.

- Added the `Find` interface.
- Also added a way to set the cache modes for a `Query`, and the `EntityGraph` for a `TypedQuery`.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -13,6 +13,7 @@
 // Contributors:
 //     Linda DeMichiel - 2.1
 //     Linda DeMichiel - 2.0
+//     Gavin King - 3.2
 
 
 package jakarta.persistence;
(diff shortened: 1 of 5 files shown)


trajano:
`with...` as long as it is fluent

gavinking:
@trajano The argument against that is that we should keep names consistent across `Query` and `Finder`. So I'm uncertain which way to go on this.

trajano:
I would prefer we mark the non-fluent `set` methods as deprecated and replace with `with` if it is fluent.  Convention-wise `with` we expect it to `return this` whereas `set` we don't.

gavinking:
@trajano after discussing with @lukasj last night, we have decided to go in a completely different direction.

trajano:
Can you link to the different direction @gavinking

gavinking:
https://github.com/jakartaee/persistence/issues/383#issuecomment-1673920733

https://github.com/jakartaee/persistence/pull/454

Issue #406, opened 2023-05-08
"entity local" @SequenceGenerator/@TableGenerator 


gavinking:
The spec says that a `@SequenceGenerator` and `@TableGenerator` annotation defines a named generator whose name is global to the whole persistence unit.

But a very common idiom is to specify one of these annotations directly on the `@Id` attribute or `@Entity` class, and in that case it makes much more sense for the generator to be seen as local to the entity.

I'm not completely sure how we should "fix" this, but one idea would be to make the `name` annotation member optional, and treat  `@SequenceGenerator` and `@TableGenerator` annotations without a name as local to the entity, and used by default. For example:

```java
@Entity 
class Book {
    @Id @GeneratedValue
    @SequenceGenerator(sequenceName="seq_book", allocationSize=10)
    Long id;
}
```

Or:

```java
@SequenceGenerator(sequenceName="seq_book", allocationSize=10)
@Entity 
class Book {
    @Id @GeneratedValue
    Long id;
}
```

would now be accepted, and the `@SequenceGenerator` would be visible only to the `Book` entity, and implicitly used as the generator for that entity.

This would avoid the need for nastiness like:

```java
@Entity 
class Book {
    @Id @GeneratedValue(strategy=SEQUENCE, generator="bookSeq")
    @SequenceGenerator(name="bookSeq", sequenceName="seq_book", allocationSize=10)
    Long id;
}
```

and is more typesafe to boot (one less stringly-typed name).

PR #416, opened 2023-05-25
Make the name member of @XxxxGenerator annotations optional

gavinking:
See #406.

### DIFF
--- a/api/src/main/java/jakarta/persistence/GeneratedValue.java
+++ b/api/src/main/java/jakarta/persistence/GeneratedValue.java
@@ -11,6 +11,7 @@
  */
 
 // Contributors:
+//     Gavin King      - 3.2
 //     Linda DeMichiel - 2.1
 //     Linda DeMichiel - 2.0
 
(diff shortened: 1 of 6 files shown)


lukasj (approved):

lukasj (commented):
the deafult name should be defined as: _SEQUENCE_ENTITYNAME_ for the class level use annotation and _SEQUENCE_ENTITYNAME_FIELDNAME_ for the field level use of the annotation

lukasj:
add a line to appendixes.adoc/changelog

gavinking:
>  the deafult name should be defined as: SEQUENCE_ENTITYNAME for the class level use annotation and SEQUENCE_ENTITYNAME_FIELDNAME for the field level use of the annotation

@lukasj after reviewing this and re-thinking it through properly this time, I realized that what we talked about isn't quite right, and I actually got it right the first time.

We need the `generator` name inferred for `@GeneratedValue` to line up with the name inferred for a `@SequenceGenerator` or `@TableGenerator`. Since the `@GeneratedValue` doesn't know what kind of generator it's looking for, nor where it's declared, the plain entity name is the right thing to use here.

If we want to nail down the actual inferred name of the database sequence or table when `sequenceName` or `table` is missing, that's something we could do, but it's an orthogonal question, since the spec currently doesn't currently constrain that at all. (It just says _"Defaults to a name chosen by persistence provider"_.)

So I think we should merge this as-is.

Sorry for the confusion.

gavinking:
> add a line to appendixes.adoc/changelog

Done.

lukasj (approved):

Issue #424, opened 2023-07-03
incorrect generic typing in operations of EntityGraph


gavinking:
There are several problems in the `EntityGraph` interface:

1. Calling `addAttributeNodes(Attribute<T, ?>... attribute)` produces the infamous *"Unchecked generics array creation for varargs parameter"* compiler warning.
2. `addAttributeNodes(Attribute<T, ?>... attribute)` should be `addAttributeNodes(Attribute<? super T, ?>... attribute)`
3.  `<X> Subgraph<X> addSubgraph(Attribute<T, X> attribute)` should be`<X> Subgraph<X> addSubgraph(Attribute<? super T, X> attribute)`
4. I'm not 100% sure, but I believe that the signature of `<X> Subgraph<X> addKeySubgraph(Attribute<T, X> attribute)` is simply wrong. I think it should be `<X> Subgraph<X> addKeySubgraph(MapAttribute<T, X, ?> attribute)`.

We need to put some thought into these.

gavinking:
Additionally, I'm very confused by this method:

```java
    /**
     * Add a node to the graph that corresponds to a managed
     * type with inheritance.  This allows for multiple subclass
     * subgraphs to be defined for this node of the entity
     * graph. Subclass subgraphs will automatically include the
     * specified attributes of superclass subgraphs. 
     *
     * @param attribute  attribute
     * @param type  entity subclass
     * @return subgraph for the attribute
     * @throws IllegalArgumentException if the attribute's target 
     *         type is not a managed type
     * @throws IllegalStateException if the EntityGraph has been 
     *         statically defined
     */
    public <X> Subgraph<? extends X> addSubgraph(Attribute<T, X> attribute, Class<? extends X> type);
```

The Javadoc is extremely cryptic, and the signature is weird. Is it supposed to be doing the equivalent of a `treat()`? If so, shouldn't the signature be:

    public <X, Y extends X> Subgraph<Y> addSubgraph(Attribute<T, X> attribute, Class<?Y> type);

because with the present signature it doesn't seem like you could do anything useful with it, assuming my interpretation is correct.

gavinking:
I'm also confused by this one:

```java
public <X> Subgraph<? extends X> addSubclassSubgraph(Class<? extends X> type);
```

which actually makes no sense even *without* knowing the semantics: Java has wildcard capture, so you don't need the wildcards there.

Perhaps it should be:

```java
[...]

PR #425, opened 2023-07-03
Proposal to fix signatures of operations of entity graphs

gavinking:
This is just a proposal to:

- fix the (I believe) incorrect signatures of some operations of `EntityGraph` and `SubGraph`, and
- introduce `AbstractGraph` as a place to declare common operations of `EntityGraph` and `SubGraph`.

See #424.

### DIFF
--- a/api/src/main/java/jakarta/persistence/AbstractGraph.java
+++ b/api/src/main/java/jakarta/persistence/AbstractGraph.java
@@ -1,0 +1,338 @@
+/*
+ * Copyright (c) 2011, 2020 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/persistence/AbstractGraph.java
```diff
@@ -0,0 +1,338 @@
...
+import jakarta.persistence.metamodel.Attribute;
+import jakarta.persistence.metamodel.MapAttribute;
+import jakarta.persistence.metamodel.PluralAttribute;
+
+import java.util.List;
+
+/**
+ * Declares operations common to {@link EntityGraph} and {@link Subgraph}.
+ *
+ * @since 6.3
```
lukasj, line 27: 6.3 is wrong version, I think

lukasj (commented):
split this work into multiple parts:
* 1st create _Graph_ interface (instead of AbstractGrapth)/refactor code
* 2nd PR pr change/additions to review

gavinking:
superseded by #430

Issue #432, opened 2023-08-08
easy access to the JDBC connection associated with the EM


gavinking:
JPA currently provides no easy way to just run some JDBC code within the transaction/connection associated with the `EntityManager`. This isn't such a big deal when using JPA inside a container, but it's a pretty big limitation when using it in standalone mode.

Hibernate has an API like this, as part of our `Session`:

```java
	/**
	 * Perform work using the {@link java.sql.Connection} underlying by this session.
	 *
	 * @param work The work to be performed.
	 *
	 * @throws HibernateException Generally indicates wrapped {@link java.sql.SQLException}
	 */
	void doWork(Work work) throws HibernateException;

	/**
	 * Perform work using the {@link java.sql.Connection} underlying by this session,
	 * and return a result.
	 *
	 * @param work The work to be performed.
	 * @param <T> The type of the result returned from the work
	 *
	 * @return the result of calling {@link ReturningWork#execute}.
	 *
	 * @throws HibernateException Generally indicates wrapped {@link java.sql.SQLException}
	 */
	<T> T doReturningWork(ReturningWork<T> work) throws HibernateException;
```

I propose we add something very similar, or even almost _identical_ to `EntityManager`.

hantsy:
I would like use `withConnection` or `doWithConnection` directly, the run/call will lead developers to think it will accept a Runnable or Callable as parameter at the first sight.

gavinking:
Well, actually I quite deliberately chose "run" and "call" to make you think of the difference between `Runnable` and `Callable`. The difference between these two operations is very much analogous.

I'm not very keen on your suggested convention because:

1. the names of the two operations are asymmetric, and
2. "with" vs "doWith" isn't suggestive at all: I *suppose* you mean to stick "do" on the one which *doesn't* return a value, but even that's not 100% obvious.

Do you have a different suggestion?

gavinking:
Note that "executeWith" and "computeWith" work and are very clear, but they're also quite verbose. (I could live with them if that's what others prefer.)

hantsy:
I just see some examples like these, eg. Micronaut Data use `withXXX` in the `XXXOperations` or `TransactionsOperator`  and Hibernate Reactive SessionFactory itself contains withSession/withTransaction, https://github.com/hantsy/quarkus-sandbox/blob/master/resteasy-reactive-hibernate/src/main/java/com/example/PostRepository.java.

gavinking:
[...]

PR #433, opened 2023-08-08
add EntityManager.runWithConnection()/callWithConnection()

gavinking:
see #432

### DIFF
--- a/api/src/main/java/jakarta/persistence/ConnectionConsumer.java
+++ b/api/src/main/java/jakarta/persistence/ConnectionConsumer.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2008, 2021 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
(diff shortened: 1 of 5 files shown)


gavinking:
> This is JPA so this looks like a h4ck3rs tool to bypass JPA API and execute whatever user wants directly trough java.sql. :)

Well, yeah, that's exactly what it is :-) 

But it's reasonably common for people to need to write this sort of l33tc0de even when they're using JPA for most things.

> Also EntityManager has `unwrap` for years 

So typically we use `unwrap()` to do things which are _not_ portable between providers.

My goal here is to let people do hardcoded JDBC calls in a way which _is_ portable. And for that `unwrap` kinda lacks type-safety.

And so you might ask, well, why didn't I just propose adding a `getConnection()` method?

And the answer is that that actually doesn't work very well or very portably because it interferes with eager releasing of connections, etc. It's better to pass the `Connection` to a lambda, since the provider can then release the connection according to whatever connection-management strategies it's using internally.

> So let's make all those users happy.

👍

lukasj (approved):
LGTM

Issue #163, opened 2017-10-18
Support for java.time.Instant


lukasj:
Original issue #63 mentioned support for java.time.Instant but it's missing in the final 2.2 spec. I haven't found any reason for omitting it.

I feel it's important for supporting pure UTC data flow as described in the #156. Instant already stores epoch internally so it cleanly maps to a UTC timestamp and it would simplify such use case rather than converting back and forth between different (application, database, user) time zones.

I believe that Instant as a [machine-timeline view](http://blog.joda.org/2009/11/why-jsr-310-isn-joda-time_4941.html) should be the recommended approach for persisting time data over LocalDateTime, OffsetDateTime, ZonedDateTime which are human-timeline views.

Hibernate [supports Instant](http://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#basic-datetime) just fine already and allows changing the database timezone via hibernate.jdbc.time_zone.

sergeykad:
@psyklopz Unfortunately this approach does not work correctly with all databases. On Microsoft SQL Server for example you will get local time instead of UTC.

psyklopz:
@sergeykad Great point.  By convention all our servers are set to UTC.  When running on a local dev machine, we pass the time zone (UTC) as part of our connection string to the MySQL JDBC driver.  

Maybe I worry more than I should, but I don't trust that time zone conversions done in Java are always consistent with those done in the database.  

I also get uncomfortable that converting from UTC to local and back to UTC is not foolproof.  In the hour before (or after) daylight savings we have a wrinkle in time.

I wish do had no concept of time zones.  That belongs in the presentation layer and not in the business layers.  But I digress...

Use my sample code if and only if your database and your Java code are both running in UTC.

sharmasourabh:
I am in favor of adding support for Instant to JPA specs.

gavinking:
@lukasj I agree that we should totally do this.

gavinking:
I've proposed support for `Instant` and `Year` in https://github.com/jakartaee/persistence/pull/434.

Note that it's clear why `Instant` was left out of previous revisions of JPA: the JDBC spec does not mention it in Appendix B. But in fact there are other missing type mappings in the JDBC spec, so what I've done is also nailed down the missing type mappings.

ritonglue:
[...]

PR #434, opened 2023-08-09
add support for Instant and Year and clarify JDBC mappings for basic types

gavinking:
see #163

### DIFF
--- a/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
+++ b/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
@@ -538,16 +538,25 @@ _java.util.UUID_, _java.util.Date_, _java.util.Calendar_,
 _java.sql.Date_, _java.sql.Time_, _java.sql.Timestamp_,
 _java.time.LocalDate_, _java.time.LocalTime_, _java.time.LocalDateTime_,
 _java.time.OffsetTime_, _java.time.OffsetDateTime_,
+_java.time.Instant_ and _java.time.Year_,
 _byte[]_, _Byte[]_, _char[]_, _Character[]_, enums, and any other type
-that implements _Serializable_.footnote:[Mapping of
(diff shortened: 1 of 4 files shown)


spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -538,16 +538,25 @@ _java.util.UUID_, _java.util.Date_, _java.util.Calendar_,
...
-annotation is optional for persistent fields and properties of these
-types. If the _Basic_ annotation is not specified for such a field or
+that implements _Serializable_.
+
+For the types listed above, the persistence provider must support mappings
+to the column types listed in tables B-2 and B-4 of Appendix B of the
+JDBC 4.3 specification. See <<a19496>>. In addition, the provider must
+support mapping:
+
+- _java.time.Instant_ to the JDBC _TIMESTAMP_ type,
```
beikov: What does "support mapping" mean in this context? Note that in Hibernate we map `Instant` to our type `SqlTypes.TIMESTAMP_UTC` which maps to `java.sql.Types.TIMESTAMP_WITH_TIMEZONE` when supported by the database and otherwise to `java.sql.Types.TIMESTAMP`.

gavinking: It means that if you point the model at a database defined with a `timestamp` column, DML will work.

This is nothing to do with schema export or anything like that. The JPA spec does not address those issues. (Rightly, in my view.)

OTOH, perhaps I should add "or `TIMESTAMP_WITH_TIMEZONE`" here.

gavinking: > OTOH, perhaps I should add "or `TIMESTAMP_WITH_TIMEZONE`" here.

I've done that.

lukasj (approved):

Issue #402, opened 2023-04-18
deprecate support for Calendar, Date, Time, Timestamp


gavinking:
The classes `java.util.Calendar`, `java.util.Date`, `java.sql.Date`, `java.sql.Time`, `java.sql.Timestamp` are bad and nobody should ever use them.

We should make this explicit by:

1. stating that use of these types in entity classes is deprecated, and
2. deprecating methods like `Query.setParameter()` which accept them.

We might even be able to deprecate `TemporalType`.

lukasj:
the real meaning here is not to _deprecate support for.._ but to _deprecate usage of..._, right?

gavinking:
> the real meaning here is not to deprecate support for.. but to deprecate usage of... and promote usage of java.time API instead, right?

Well the idea is we want to tell *users* of JPA not to use these things in new code, and to use the `java.time` types instead.

Of course persistence providers will still be required to support them, and I'm not suggesting any sort of plan to remove that support.

PR #435, opened 2023-08-09
deprecate the use of java.util.Date and friends

gavinking:
see #402


lukasj (commented):
mark coresponding methods/temporals as deprecated as well

gavinking:
> mark coresponding methods/temporals as deprecated as well

Done. WDYT?

lukasj (approved):
yes, this is it

PR #436, opened 2023-08-09
add Finder interface

gavinking:
Draft proposal for a new `Finder` interface allowing typesafe passing of caching and locking options and an `EntityGraph`.

see #383

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -93,7 +93,18 @@ public interface EntityManager extends AutoCloseable {
      *         no transaction
      */    
     public void remove(Object entity);
-    
+
+    /**
+     * Obtain an instance of {@link Finder} for the given entity class.
+     * @param entityClass  entity class
+     * @return a {@link Finder} to find instances of the entity by primary key
+     * @throws IllegalArgumentException if the first argument does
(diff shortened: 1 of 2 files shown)


trajano:
What happens when both loadGraph and fetchGraph are provided?
If we only allow one then
```
with(EntityGraph eg, EntityGraphLoadType loadOrFetch)
```

gavinking:
[P.S. Please compare this to #415 which has slightly different naming conventions.]

trajano:
> > What happens when both loadGraph and fetchGraph are provided?
> 
> Well, what happens today if you specify both hints? The semantics should be the same, I guess.

Correct, that semantic should be defined or be made explicit.  It does not make sense to have both of them and the behaviour is presently undefined.  

I'd like to change the proposal though to reverse the parameter order
```
with(EntityGraphLoadType loadOrFetch, EntityGraph eg)
```

So when people code it
```
.with(FETCH, myEntityGraph)
.with(LOAD, myEntityGraph);
```

And explicitly state to prevent ambiguity that the later of the two `with` will win.

gavinking:
So that makes perfect sense, but personally I find it really quite a lot less ergonomic. IDEs do a somewhat *better* job of completing enum values than in the past, but even so I really find that they still get in the way and disrupt my "flow" when I'm typing this kind of code.

Issue #76, opened 2014-03-17
Allow specification for null handling in order-by expressions (JPQL and Criteria API)


lukasj:
§4.9 of the specification explictly states:

> SQL rules for the ordering of null values apply: that is, all null values must appear before all non-null values in the ordering or all null values must appear after all non-null values in the ordering, but it is not specified which.

However, pretty much all of the important JPA providers support a <tt>nulls first</tt>/<tt>nulls last</tt> clause. So while it is already possible to define the strategy, it would be cool if one could reliably use it on top of JPA.

odrotbohm:
Is there a chance, this get's prioritized for JakartaEE 10? Users of Spring Data JPA and downstream projects have been repeatedly and patiently asked for this for years, and it's a severe limitation compared to accessing other store types.

beikov:
If anyone is willing to work on this, Hibernate can serve as a compatible implementation.

opengeekv2:
How hard would it be to take this on as a new (and ocasional) contributor?

gavinking:
I have proposed a solution in #440.

PR #440, opened 2023-08-09
add support for 'nulls first' and 'nulls last' in queries

gavinking:
see #76.

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
+++ b/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
@@ -117,19 +117,37 @@ public interface CriteriaBuilder {
 	
     /**
      * Create an ordering by the ascending value of the expression.
-     * @param x  expression used to define the ordering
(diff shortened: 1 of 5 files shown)


api/src/main/java/jakarta/persistence/criteria/NullPrecedence.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Specifies the precedence of null values within query result sets.
+ *
+ * @see CriteriaBuilder#asc(Expression, NullPrecedence)
+ * @see CriteriaBuilder#desc(Expression, NullPrecedence)
+ *
+ * @since 3.2
+ */
+public enum NullPrecedence {
```
gavinking: I'm _almost_ inclined to just call this enum `Nulls`. WDYT?

lukasj: let's go with `Nulls`

lukasj (approved):

Issue #30, opened 2012-05-24
Case sensitivity in JPQL queries


lukasj:
In JPA 2.0, JPQL is case insensitive. This means that contrary to Java code, an identifier can not be same name as an entity with the first letter in lower case.

For instance in the following query an underscore has to be added to make the name legal:

```
SELECT
    _user
FROM
    User _user
```

Other typical workarounds for this are abbreviating the identifier to e.g. "u", "usr" etc. Such names are rarely a best practice and actually against common advice for good naming given by books like Clean Code, etc.

I would like to request support for case sensitivity in some way, or alternatively to lift the requirement that identifiers may not have the same name as Entity classes. This last approach is taken by some persistence providers, e.g. Hibernate, which silently accepts identifiers having the same name as Entities.

One possible solution would be to use a setting, either per query or globally for the entire persistence unit. E.g.

```
<named-query name="User.getAll" caseSensitive="true">
    <query>
        SELECT
            user
        FROM
            User user
    </query>
</named-query>
```

gavinking:
> In JPA 2.0, JPQL is case insensitive. This means that contrary to Java code, an identifier can not be same name as an entity with the first letter in lower case.

To be clear: keywords and identification variables are case-insensitive.

The spec is actually silent on the question of whether range variables and field references in path expressions are case-sensitive (oops!), but it's _strongly_ implied, and intended, that they are case-sensitive.

I'm very surprised to hear that a range variable like `User` can be confused with an identification variable like `user`, since there occur in syntactically distinguishable locations! Hibernate has always accepted `from User user` and I would have thought that other implementations also support this.

That said, I think it was a mistake to say that identification variables are case-insensitive, and I wouldn't mind fixing that. But I can't think of a good and elegant way to fix it that doesn't imply breaking backward compatibility.

So how about we just flatly state in the spec that `from User user` is acceptable, and that range variables live in a different namespace to identification variables? That's like a one-line fix. WDYT?

gavinking:
See proposed fix #449.

gavinking:
> I'm very surprised to hear that entity names like `User` can be confused with an identification variable like `user`, since there occur in syntactically distinguishable locations!

Well, that's perhaps overstating the case _slightly_. We also need to consider disambiguating `entity_type_literal`  from identification variables in the expression grammar.

[...]

PR #449, opened 2023-08-10
put entity names and identification vars in distinct namespaces

gavinking:
and clarify rules around case-sensitivity

see #30


lukasj (approved):
LGTM

Issue #65, opened 2013-08-31
Need another property to make lazy loading of attributes easy with entity graphs


lukasj:
There are currently two properties to use with entity graphs:

javax.persistence.fetchgraph -

This property accepts an entity graph to act as a complete override of all the attributes for the type. Attributes are dictated to be eager if included or lazy if excluded from the graph, regardless of how they are mapped.

javax.persistence.loadgraph -

This property offers a selective eager approach. One can simply add the attributes that one wants to be eagerly loaded and the rest are left as they are statically mapped.

The missing property would be something to allow more convenient selective lazy overriding without having to declare the entire attribute set for the type (as required by fetchgraph). So something like:

javax.persistence.lazygraph -

This property would offer a selective lazy approach. One would be able to add the attributes that one wants to be lazily loaded to the graph, with the rest being left as they are statically mapped.

Issue #64, opened 2013-08-27
Mutability of collections returned by EntityGraph API


lukasj:
The List/Map getter methods (e.g. getAttributeNodes()) on EntityGraph, AttributeNode, and Subgraph do not specify whether the List/Map returned is a mutable one or a copy. They should be specified as returning the exact collections so the collections can be mutated. If it is a copy then the ability to mutate an existing named entity graph using createEntityGraph(String) is quite limited. There would be no way to remove an attribute node or a subgraph, or for that matter even add a subgraph for an existing attribute node.

The alternative to returning the actual collections is to add methods to the API to enable the additional mutating operations.

gavinking:
> There would be no way to remove an attribute node or a subgraph, or for that matter even add a subgraph for an existing attribute node.

This is at least _partially_ addressed in #454.

PR #454, opened 2023-08-11
introduce FindOption interface and new overloads of EM.find()

gavinking:
Introduces:

- `FindOption` interface and new overloads of `EM.find()`
- `RefreshOption` interface and new overloads of `EM.refresh()`
- `LockOption` interface and new overloads of `EM.lock()`
- `Graph.removeAttributeNode()`
- `setCacheStoreMode()`/`setCacheRetreiveMode()` on `EntityManager` and `Query`

see #383 and #65. (Also #64.)

### DIFF
--- a/api/src/main/java/jakarta/persistence/Timeout.java
+++ b/api/src/main/java/jakarta/persistence/Timeout.java
@@ -22,7 +22,7 @@ package jakarta.persistence;
  *
  * @since 3.2
  */
-public class Timeout implements FindOption {
+public class Timeout implements FindOption, RefreshOption {
     private final int milliseconds;
 
     private Timeout(int milliseconds) {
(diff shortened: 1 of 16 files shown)


lukasj (approved):
LGTM

api/src/main/java/jakarta/persistence/Timeout.java
```diff
@@ -0,0 +1,68 @@
...
+
+    private Timeout(int milliseconds) {
+        this.milliseconds = milliseconds;
+    }
+
+    /**
+     * Specify a timeout in seconds.
+     * (Abbreviation of {@link #seconds(int)}.)
+     */
+    public Timeout s(int seconds) {
```
lukasj, line 36: whoops, another oversight - this (and few other methods) should be static, right now there is no way to create/get an instance of the `Timeout`

gavinking: OMG I suck. Sorry.

Issue #409, opened 2023-05-18
@ManyToOne defaults to eager fetching and this is almost always wrong


gavinking:
Currently we have:

```java
public @interface ManyToOne {
    ...
    FetchType fetch() default FetchType.EAGER;
}
```

This is actually a really bad default, and has a tendency to lead to N+1 selects. And of course it's also inconsistent with the default for `@OneToMany` and `@ManyToMany`.

I propose that we finally fix this. I've been trying to see how we could do that without breaking existing programs, and here's what I've come up with:

1. Introduce a new member to `FetchType`:
       
       public enum FetchType { LAZY, EAGER, DEFAULT }
       
2. Change the definition of `@ManyToOne` to
      ```java
      public @interface ManyToOne {
          ...
          FetchType fetch() default DEFAULT;
      }
      ```
   (And do the same to `@OneToOne`.)
3. Introduce a configuration property `jakarta.persistence.defaultFetchType` which accepts `EAGER` or `LAZY`. Specify that it defaults to `EAGER`.

One might object that this proposal looks like it breaks binary compatibility, but I would argue that it is effectively binary back-compatible, in that programs compiled against the older APIs continue to work.

WDYT?

gastaldi:
So that fetch type could be overriden globally regardless if used in a `@ManyToOne`, `@OneToMany`, `@ManyToMany` or `@Basic`? Wouldn't that mess with some queries that expect the default fetch type for each relationship?

gavinking:
@gastaldi not sure what you mean ... example?

gastaldi:
My concern is that introducing a property that impacts every association fetch type may introduce unexpected behavior, specially when your code queries a relationship assuming the default fetch type, like the `Book` x `Author` example:

```java
EntityManager em = emf.createEntityManager();
em.getTransaction().begin();
 
TypedQuery<Author> q = em.createQuery(
        "SELECT a FROM Author a",
        Author.class);
List<Author> authors = q.getResultList();
em.getTransaction().commit();
em.close();
 
for (Author author : authors) {
    List<Book> books = author.getBooks();
    log.info("... the next line should throw LazyInitializationException but it may not if the default fetch type is overridden");
    books.size();
}
```

Maybe it's not a legitimate concern but I think overriding this property may break existing legacy working code.

gastaldi:
Shouldn't the association be part of the property name then? Eg. `jakarta.persistence.manyToOne.defaultFetchType=LAZY`?

[...]

PR #455, opened 2023-08-11
add FetchType.DEFAULT and make it the default for @MenyToOne + @OneToOne

gavinking:
see #409

### DIFF
--- a/api/src/main/java/jakarta/persistence/FetchType.java
+++ b/api/src/main/java/jakarta/persistence/FetchType.java
@@ -42,10 +42,20 @@ package jakarta.persistence;
  * @since 1.0
  */
 public enum FetchType {
-
-    /** Defines that data can be lazily fetched. */
+    /**
+     * Specifies that data can be lazily fetched.
+     */
     LAZY,
 
-    /** Defines that data must be eagerly fetched. */
-    EAGER
+    /**
+     * Specifies that data must be eagerly fetched.
+     */
+    EAGER,
+
+    /**
+     * Specifies that the fetching strategy depends on the value
+     * of the property {@code jakarta.persistence.defaultFetchType}.
+     * @since 3.2
+     */
+    DEFAULT
 }
--- a/api/src/main/java/jakarta/persistence/ManyToOne.java
+++ b/api/src/main/java/jakarta/persistence/ManyToOne.java
@@ -103,12 +103,14 @@ public @interface ManyToOne {
 
     /** 
      * (Optional) Whether the association should be lazily 
-     * loaded or must be eagerly fetched. The EAGER
-     * strategy is a requirement on the persistence provider runtime that 
+     * loaded or must be eagerly fetched. The EAGER strategy
+     * is a requirement on the persistence provider runtime that
      * the associated entity must be eagerly fetched. The LAZY 
      * strategy is a hint to the persistence provider runtime.
+     * <p>The default fetching strategy depends on the value of
+     * the property {@code jakarta.persistence.defaultFetchType}.
      */
-    FetchType fetch() default FetchType.EAGER;
+    FetchType fetch() default FetchType.DEFAULT;
 
     /** 
      * (Optional) Whether the association is optional. If set 
--- a/api/src/main/java/jakarta/persistence/OneToOne.java
+++ b/api/src/main/java/jakarta/persistence/OneToOne.java
@@ -134,12 +134,14 @@ public @interface OneToOne {
 
     /** 
      * (Optional) Whether the association should be lazily 
-     * loaded or must be eagerly fetched. The EAGER 
-     * strategy is a requirement on the persistence provider runtime that 
+     * loaded or must be eagerly fetched. The EAGER strategy
+     * is a requirement on the persistence provider runtime that
      * the associated entity must be eagerly fetched. The LAZY 
      * strategy is a hint to the persistence provider runtime.
+     * <p>The default fetching strategy depends on the value of
+     * the property {@code jakarta.persistence.defaultFetchType}.
      */
-    FetchType fetch() default FetchType.EAGER;
+    FetchType fetch() default FetchType.DEFAULT;
 
     /** 
      * (Optional) Whether the association is optional. If set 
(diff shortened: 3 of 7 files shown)


gavinking:
I am hereby reviving this one, and moving it out of draft mode.

@sebersole @njr-11 @anija-anil Please review this and think about it.

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -516,6 +518,27 @@ public ValidationMode validationMode() {
...
     }
 
+    /**
+     * Specify the {@linkplain FetchType#DEFAULT default fetch type}
+     * for one-to-one and many-to-one associations.
+     * @param defaultFetchType the default fetch type
+     * @return this configuration
+     * @since 4.0
+     */
+    public PersistenceConfiguration defaultFetchType(FetchType defaultFetchType) {
```
njr-11: Disallow setting a default of DEFAULT.
```suggestion
     * for one-to-one and many-to-one associations. The given fetch type
     * value must be {@link FetchType#EAGER} or {@link FetchType#LAZY}.
     * @param defaultFetchType the default fetch type
     * @return this configuration
     * @since 4.0
     */
    public PersistenceConfiguration defaultFetchType(FetchType defaultFetchType) {
        if (defaultFetchType == FetchType.DEFAULT)
            throw new IllegalArgumentException("defaultFetchType: " + FetchType.DEFAULT);
```

api/src/main/java/jakarta/persistence/FetchType.java
```diff
@@ -31,13 +31,39 @@
...
+ * private Publisher publisher;
  * }
+ * <p>This example specifies that an {@link ElementCollection} is
+ * fetched eagerly:
+ * {@snippet :
+ * @ElementCollection(fetch = EAGER)
+ * Set<String> topics;
+ * }
+ * <p>The use of {@code fetch = EAGER} often leads to fetching of
+ * unwanted data, and is therefore discouraged. Eager fetching
```
njr-11:
```suggestion
 * unwanted data and is therefore discouraged. Eager fetching
```

njr-11 (commented):
Good idea here.  I added a suggestion to prevent someone from setting the default to DEFAULT.   I think we should give it a more precise name so that users won't believe they can use this to set a default of EAGER for other association types.  I'm not sure what to call it, though. Ideally, something that communicates  `setDefaultXToOneAssociationFetchType` but which isn't that long.

gavinking:
> I added a suggestion to prevent someone from setting the default to DEFAULT. 

+1. Applied.

> I think we should give it a more precise name so that users won't believe they can use this to set a default of EAGER for other association types. I'm not sure what to call it, though. 

Yes, I have the same problem. I did not come up with anything good.

gavinking:
> something that communicates setDefaultXToOneAssociationFetchType but which isn't that long.

How about `defaultEntityFetchType()`, to distinguish fetching an entity-values field from fetching a collection-valued field or basic field?

njr-11:
> > something that communicates setDefaultXToOneAssociationFetchType but which isn't that long.
> 
> How about `defaultEntityFetchType()`, to distinguish fetching an entity-values field from fetching a collection-valued field or basic field?

I could still see users misinterpreting it as applying to all fetched entities, whether fetching multiple or singular.  Although we could clarify that with

`defaultSingularFetchType`
or
`defaultSingleEntityFetchType`

gavinking:
I don't think "single" or "single-valued" work here because a `@Basic` field is single-valued and it has a `fetch`. (Which defaults to `EAGER`, and we definitely don't want to ever change that!)

So the only alternatives I can think of are `defaultEntityFetchType` and `defaultToOneFetchType`. I don't love either of them.

njr-11:
> So the only alternatives I can think of are `defaultEntityFetchType` and `defaultToOneFetchType`. I don't love either of them.

Of these, I would go with `defaultToOneFetchType` because the name will make it clear that the default's scope is limited to *-to-One associations.

sebersole:
I really like this introduction of `FetchType.DEFAULT` fwiw.  Selfishly, as a provider, I always hated that to-ones defaulted to EAGER and there was not a good compliant way to override that.


> > So the only alternatives I can think of are `defaultEntityFetchType` and `defaultToOneFetchType`. I don't love either of them.
> 
> Of these, I would go with `defaultToOneFetchType` because the name will make it clear that the default's scope is limited to *-to-One associations.

+1

gavinking:
OK, since there is a, um, "consensus" of 2/3, I have made that change. :-)

gavinking:
I believe this is good to merge now, given the 17 votes in favor on the issue description, and positive feedback here.

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -516,6 +518,30 @@ public ValidationMode validationMode() {
...
+     * Specify the {@linkplain FetchType#DEFAULT default fetch type}
+     * for one-to-one and many-to-one associations. The given fetch type
+     * value must be {@link FetchType#EAGER} or {@link FetchType#LAZY}.
+     * @param defaultToOneFetchType the default fetch type
+     * @return this configuration
+     * @since 4.0
+     */
+    public PersistenceConfiguration defaultToOneFetchType(FetchType defaultToOneFetchType) {
+        if (defaultToOneFetchType == FetchType.DEFAULT)
+            throw new IllegalArgumentException("defaultToOneFetchType: " + FetchType.DEFAULT);
```
njr-11: I missed this before. We should also disallow `null`,
```suggestion
        if (defaultToOneFetchType == FetchType.DEFAULT ||
            defaultToOneFetchType == null)
            throw new IllegalArgumentException("defaultToOneFetchType: " + defaultToOneFetchType);
```

njr-11 (approved):
Looks good. I suggested a minor improvement to my prior suggestion. We can also prevent users from setting it to `null`.

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
gavinking: @njr-11 applied.

gavinking:
Thanks guys for your feedback. Going to pull the trigger on this.

Issue #456, opened 2023-08-12
Use Named or Qualifier to replace the existing PersistenceContext/PersistenceUnit


hantsy:
When switching to CDI first, #377, I would like deprecate the existing `PersistenceContext/PersistenceUnit` uses `@Named` or `@Qualifier` with `@Inject` instead.

```java
@Inject @Named("customerDatabase")  EntityManager em;
@Inject @CustomerDatabase  EntityManager em;

@Inject @CustomerDatabase  EntityManagerFactory emf;
```

`@Qualifier` is preferred, declaring an EMF with the same qualifiers, see #358

gavinking:
Of course, if we address #377, that's the obvious thing to do. But look, we already have #377 and #309, I'm not sure we need to scatter the discussion across _three_ different issues.

Issue #377, opened 2022-09-04
Directly support CDI for injecting EntityManager and EntityManagerFactory


arjantijms:
Jakarta Persistence predates CDI, and uses its own annotations to inject the EntityManager and EntityManagerFactory.

In order to make Jakarta EE more consistent, these should be directly injectable using CDI with an optional qualifier to specify the persistence unit name if needed.

gavinking:
I'm not sure this is a real issue. The CDI spec (since the very first version) shows examples of using CDI annotations to inject the `EntityManager`.

In section 1.3.3, there is this example:

```java
 public class Databases {
      @Produces @PersistenceContext(unitName="UserData")
      @Users EntityManager userDatabaseEntityManager;

      @Produces @PersistenceUnit(unitName="UserData")
      @Users EntityManagerFactory userDatabaseEntityManagerFactory;

      @Produces @PersistenceContext(unitName="DocumentData")
      @Documents EntityManager docDatabaseEntityManager;
  }
```

And then:

```java
@Inject @Users EntityManager userDatabase;
```

Now, in principle, we could allow the CDI qualifier annotation to be specified in the persistence unit declaration in `persistence.xml`, but it's not completely clear to me that this would be much of an improvement.

jgrassel:
I agree, I always felt like this was a glaring omission in the JPA specification.

m-reza-rahman:
Would indeed be a good improvement. This could also take into account how transactions and concurrency is supposed to work outside of the better defined realm of EJB.

arjantijms:
> I'm not sure this is a real issue. 

This issue (337) must be real, as I've opened it and people are responding to it. Okay, jokes aside, there are obviously workarounds for this (as there's with pretty much everything, thank you for designing both CDI and JPA in such way to make this possible).

But the idea is to not need all these workarounds or examples anymore, but just support it directly and natively.

Put differently, if CDI had happened to exist before JPA was designed, would `@EntityManager` have been used in it like it's now?

gavinking:
> But the idea is to not need all these workarounds or examples anymore, but just support it directly and natively.

I guess what I don't quite understand is what about the current design is it that you're worried is not native.

I mean, as far as I'm aware, Java EE containers today already implement `@PersistenceContext` and `@PersistenceUnit` injection using CDI, in terms of native CDI constructs. (Please correct me if I'm wrong.)

[...]

Issue #309, opened 2021-04-17
Support @PersistenceContext injection in CDI SE and CDI/Servlet environment


hantsy:
Like Spring, add CDI decorations to provide injection for `EntityManager` and `EntityMangerFactory` for CDI SE and Servlet/CDI environment.

* CDI SE, the application is bootstrapped by a CDI container, even the embedded tomcat/jetty can be declared as CDI beans.
* Servlet/CDI, the application is war deployed into a standard servlet container that enabled CDI support via a servlet Listener.

I think there are some approaches already support this,  such as Apache DeltaSpike and  Quarkus, Helidon etc.

arjantijms:
I'd go one step further and not just add some "decorations", but make the CDI way the primary way and deprecate the existing annotations. See https://github.com/eclipse-ee4j/jpa-api/issues/377

gavinking:
> See https://github.com/jakartaee/persistence/issues/377

Yeah, I think this and #377 are essentially the same issue. Or at least the solution would be the same.

Issue #70, opened 2014-01-09
Allow two phase bootstrap approach to creating the EntityManagerFactory


lukasj:
Introduce a standard two phase persistence unit bootstrap and require that certain services are not accessed until the second phase. This will improve how JPA/CDI + JPA/@DataSourceDefinition work together.

In a perfect world, the first persistence unit bootstrap phase should not use the datasource or the CDI bean manager (instead wait for the second phase). Class enhancing/rewriting should also occur in the first phase (before application classes have been read).

This will help avoid use of the CDI bean manager too early (which otherwise can cause application entity classes to be loaded before they are enhanced/rewritten).

This will help avoid use of a datasource that is being deployed with the application (@DataSourceDefinition) but may not be available for use yet.

Also see discussion at [https://java.net/projects/jpa-spec/lists/jsr338-experts/archive/2013-06/message/0](https://java.net/projects/jpa-spec/lists/jsr338-experts/archive/2013-06/message/0).

gavinking:
@scottmarlow what is the status of this issue? Is it still relevant?

scottmarlow:
> @scottmarlow what is the status of this issue? Is it still relevant?

WildFly + Hibernate implementations have already solved this.  I think we could close this and reopen it if other implementations are still interested in having a **standard** way for bytecode enhancement to work consistently without contending with the CDI need to rewrite application classes.  

WebSphere/WebLogic teams expressed interested years ago (just missed the JPA 2.1 release train in 2013) but I haven't heard any feedback recently.  

@lukasj @Emily-Jiang have either of you heard any recent feedback on whether this issue is still relevant?

gavinking:
The underlying need here might have been substantially already solved by #650.

gavinking:
In fact, yes, re-reading, I believe they are dupes. Closing this one too.

PR #460, opened 2023-08-12
define how a CDI container should provide native support for JPA

gavinking:
Draft spec for direct CDI integration.

see #377, #456, #309


arjantijms:
Note that under Eclipse, we renamed JPA to "Jakarta Persistence" and updated all references in Jakarta EE 9.

gavinking:
Ah yeah, good point, thanks.

lukasj (approved):
From the persistence spec point of view, this looks good.

But since this imposes requirement on the CDI, it would be good to get an opinion on this change from them as well

arjantijms:
> But since this imposes requirement on the CDI, it would be good to get an opinion on this change from them as well

I may have missed something, but what are the requirements this imposes on CDI? Isn't this just using portable extensions of CDI and the CDI SPI and API in general?

gavinking:
> I may have missed something, but what are the requirements this imposes on CDI? Isn't this just using portable extensions of CDI and the CDI SPI and API in general?

This is a really interesting question.

I mean yeah, absolutely this could be implemented as a portable extension. But .... we're being asked to impose a requirement on someone here, and if it's just a portable extension, exactly _who_ is being required to provide that?

It can't be the _persistence provider_ which supplies it, IMO, which is why I was skeptical at first that this even really belongs in the JPA spec.

So @arjantijms perhaps you think I should adjust that wording to say it's the Jakarta EE container which is required to provide this integration (possibly via a portable extension). I guess I'm OK with that. I was trying to make it a bit more general-purpose I guess.

gavinking:
> I'm curious, why can't the persistence provider (Jakarta Persistence implementation) not supply this?

Well, first, because that would be a pretty major change to the division of responsibilities between persistence provider and container. The spec was written so that the persistence provider is a simple library that really knows very little about container integration, so that injection, etc, was always the responsibility of the container.

Second because unlike the `SecruityContext`, `FacesContext`, or `UserTransaction`, creating an `EntityManagerFactory` requires a very complex archive scanning process that the persistence provider is simply physically incapable of performing without being passed a lot of context from the container.

So:

- the persistence provider has to wait until it gets called with a `PersistenceUnitInfo` by the container before it creates an `EMF`, but
- it has to register portable extensions which are specific to those persistence units in a call to the `Extension`.

This means that the JPA spec would have to start micromanaging a multi-phase interaction between the container and the persistence provider. Now, that's not impossible, but in my opinion the JPA spec is a _very poor_ place to start writing down this sort of stuff.

By contrast, if registering beans is the responsibility of the _container_, then the spec can remain silent about _how_ that occurs, and the container is free to do it whatever way works.

arjantijms:
> The spec was written so that the persistence provider is a simple library that really knows very little about container integration, so that injection, etc, was always the responsibility of the container.

Wasn't that another time / era? I guess that was once a good idea so that not every spec had to duplicate an injection mechanism. Now that there is a general "injection service" available, this concern (if that indeed ever was the concern) is not really applicable anymore.

Jakarta Transactions (JTA at the time) was also written where the container had to have different (mostly more) responsibilities, yet it did evolve in a way such that `UserTransaction`, but also interceptors such as `Transactional` were able to be delivered by a portable CDI extension (see e.g. https://github.com/OmniFish-EE/omni-transact/tree/main/jta/src/main/java/ee/omnifish/transact/jta/cdi)

> This means that the JPA spec would have to start micromanaging a multi-phase interaction between the container and the persistence provider.

Is that really all necessary? Why not just let things proceed as they do today, and then provide a bean that essentially wraps the EntityManager and when called by user code resolved the underlying entity manager in a way it normally would do programmatically.

I dabbled a little into that approach in Piranha:

https://github.com/piranhacloud/piranha/blob/current/extension/eclipselink/src/main/java/cloud/piranha/extension/eclipselink/EntityManagerProducer.java

arjantijms:
p.s.

> creating an EntityManagerFactory requires a very complex archive scanning process that the persistence provider is simply physically incapable of performing without being passed a lot of context from the container.

Does that mostly concerns scanning of EARs? As in the WAR case it does seem the persistence provider is capable of doing this (as evidenced by e.g. adding Hibernate or EclipseLink to Tomcat, but also by the fact that in Piranha I use the scanning process as being done by EclipseLink)

gavinking:
> Is that really all necessary? Why not just let things proceed as they do today, and then provide a bean that essentially wraps the `EntityManager` and when called by user code resolved the underlying entity manager in a way it normally would do programmatically.

Yes, I believe it is, because of the requirement to scan archives.

There is information in the `PersistenceUnitInfo` (a class loader, a `ClassTransformer`, the root URL of the persistence unit, the URLs of container jar files) which simply isn't available to a CDI `Extension`.

I guess @scottmarlow is the expert on this.

gavinking:
> Does that mostly concerns scanning of EARs? 

I'm not sure. Possibly?

arjantijms:
> There is information in the PersistenceUnitInfo (a class loader, a ClassTransformer, the root URL of the persistence unit, the URLs of contained jar files) which simply isn't available to a CDI Extension.

I think those don't have to be available to the CDI extension. The CDI extension can return a delegating entity manager, that delegates to the actual entity manager. The actual entity manager can be obtained via "traditional" means.

Traditional can then be done via something like e.g.

```java
private EntityManager getNonTxScopedEntityManager() {
        return
            CDI.current()
               .select(NonTxEntityManagerHolder.class)
               .get()
               .computeIfAbsent(() ->
                   getEntityManagerFactory().createEntityManager(synchronizationType, properties));
    }
```

and

```java
public EntityManagerFactory getEntityManagerFactory() {
        if (entityManagerFactory == null) {
            entityManagerFactory = CDI.current()
                    .select(EntityManagerFactoryCreator.class)
                    .get()
                    .get(unitName);
        }

        return entityManagerFactory;
    }
```

Where eventually `EntityManagerFactoryCreator` is a class provided by the persistence provider, which knows how to create its own `EntityManagerFactory` implementation (doing that in the same way it would do without any CDI in the loop).

gavinking:
> The CDI extension can return a delegating entity manager, that delegates to the actual entity manager. 

But how does it even know what persistence units exist in the deployment? We would have to specify the the portable extension is started *after* the persistence units are already created.

So we could be imposing a lifecycle on the container.

I'm not saying it can't work. I'm just saying that I don't see why the JPA spec should be imposing such things on container implementors when it could in principle, _just say nothing on the topic_.

I mean, look, if I'm implementing a container I feel like I want to have control over this stuff. I don't think I want the persistence provider to be doing it's own crap that imposes a particular lifecycle on me.

gavinking:
I think you think that making the persistence provider provide an `Extension` makes this easier for the container implementor. 

That's not my intuition: I think it makes things harder for *both* the Jakarta EE container implementor, *and* for the persistence provider implementor.

gavinking:
P.S. I'm not sure how much this discussion relates to issue #70, but it might be a similar problem.

arjantijms:
> But how does it even know what persistence units exist in the deployment?

It doesn't have to know that. It just collects what appears at injecting sites within the application. E.g.

```java
public class EntityManagerProducer {

    /**
     *
     * @param injectionPoint the injectionPoint
     * @return EntityManager
     */
    @Produces
    public EntityManager produce(InjectionPoint injectionPoint) {
        Annotated annotated = injectionPoint.getAnnotated();
        if (annotated instanceof AnnotatedParameter<?> annotatedParameter) {
            annotated = annotatedParameter.getDeclaringCallable();
        }

        PersistenceContext persistenceContext = annotated.getAnnotation(PersistenceContext.class);

        return new PiranhaEntityManager(
                    persistenceContext.unitName(),
                    persistenceContext.type(),
                    persistenceContext.synchronization(),
                    propertiesToMap(persistenceContext.properties()));

    }

    private Map<Object, Object> propertiesToMap(PersistenceProperty[] properties) {
        return Arrays.stream(properties)
                     .collect(toMap(
                        PersistenceProperty::name,
                        PersistenceProperty::value));
    }
}
```

arjantijms:
>  I think it makes things harder for both the Jakarta EE container implementor, and for the persistence provider implementor.

As a Jakarta EE container implementor for both Piranha Cloud and GlassFish, I *think* it makes my life easier, and not just a little but to a very large degree. 

Maybe I'm just not seeing something still, which may well be the case. 

But I've integrated EclipseLink in Piranha, and even having to still write the extension myself, it was pretty easy. If that extension already existed it would be trivial.

That said, I must admit that on Piranha I haven't even tried to run the Jakarta Persistence TCK, so maybe I've overlooked something very important.

arjantijms:
> WDYT folks, is that better?

My own preference, which I hope would be possible, is more something along the lines of:

"In a Jakarta EE environment, the Jakarta Persistence implementation MUST provide a build-in unqualified CDI bean implementing the `EntityManager` interface that represents the default entity manager. Additional build-in qualified CDI beans MUST be made available for every defined entity manager"

gavinking:
>  It doesn't have to know that. It just collects what appears at injecting sites within the application. > E.g.
>
>
>     public class EntityManagerProducer {
>
>        @Produces
>        public EntityManager produce(InjectionPoint injectionPoint) {

Hahahahaha OMG section 1.3.5 of the CDI spec&mdash;I wrote that but I had completely forgotten it existed.

(Anyway it's pretty nice and it's nice that some people know it exists.)

Yeah OK, so I admit that this _almost_ works. But if you have multiple providers, each with their own `Extension`, won't they "fight" over the persistence units?

And, worse, I fear you won't get proper startup-time messages from the bean container about unsatisfied dependencies on `EntityManager`s. Because every dependency on an EM is satisfies by the producer. Or am I wrong? Clearly I don't know CDI as well as I thought I did 😂

sebersole:
The situation as @scottmarlow and I played with between Hibernate and WildFly was that a well-defined extension would be beneficial.  

I don't think security, JTA, etc are good corollaries for this situation because here we have a bidirectional dependency - JPA expects that the `BeanManager` be available and ready to use when the `EntityManagerFactory` is being booted and CDI expects the `EntityManagerFactory` to be available (at least for reference injection) when the `BeanManager` is being booted.  

In Hibernate we ended up handling with through delayed access to the BeanManager.  We specifically have 2 delayed access strategies -

1. The container passes us the `BeanaManager` and we simply agree to not use it until after the `EntityManagerFactory` is fully booted.  In effect, this is a "first use" strategy.  
2. Scott and I developed an `ExtendedBeanManager` contract that is a delayed reference to the `BeanManager` from the container, which is used to register a callback for when the `BeanManager` is really ready to use.  

Personally, I think the second approach is better because its a known point where we expect all the beans to be available and we can perform the CDI lifecycle at one point.  The "first use" strategy also delays bean wiring exceptions, which is not particularly user-friendly in my opnion.

sebersole:
> Yeah OK, so I admit that this almost works. But if you have multiple providers, each with their own Extension, won't they "fight" over the persistence units?

If I understand the point you are trying to make, it was THE reason why CDI's built-in listener stuff would not work and why we ended up creating our `ExtendedBeanManager` contract - it allows scoping

sebersole:
> The "first use" strategy also delays bean wiring exceptions, which is not particularly user-friendly in my opnion.

I wanted to clarify I point here.  

With immediate access, when the JPA provider attempts to resolve the bean and there is a problem, that problem can become an exception which triggers a failure in the deployment.  Failing the deployment is ideal because we have a state where that bean will never be usable.

However, by themselves, both delayed approaches mean that we lose the ability.  The `ExtendedBeanManager` approach at least gives the container the option of making that happen so long as the `ExtendedBeanManager` callback is made during the deployment phase - a 2 phase process:

1.  Create a `ExtendedBeanManager`
2. Create the `EntityManagerFactory` passing along the `ExtendedBeanManager`
3. Ceate the `BeanManager`
4. Trigger callback on the `ExtendedBeanManager` passing along the usable `BeanManager`

where each of those steps happen during deployment.

gavinking:
OK, so I have run this issue past our CDI and ORM people here internally, and the feedback I got was:

1. They are happy with the (updated) language on this PR, where it's clear that we're making this functionality a requirement on the Java EE container.
2. There are several different, all imperfect, ways to implement integration between CDI and JPA, ranging from some sort of phased lifecycle for CDI initialization, to lazy resolution of JPA beans, similar to what is outlined by @arjantijms above. Our preference has _not_ been to implement this via lazy resolution since that essentially circumvents CDI's dependency validation (detection of ambiguous/unsatisfiable dependencies).

The insight from @sebersole was that there's a real chicken/egg problem here: the EMF needs beans from CDI for stuff like injection into entity listeners, but CDI beans need to inject EMFs. And that there's a hierarchy of solutions to this problem which range from "cleaner" to "messier", with the cleaner solutions having the disadvantage that CDI is bypassed and JPA does its own ad hoc dependency validation.

It seems to me that the core problem here is that we have two different "containers", the bean manager, and the EMF, each with their own "components" and cross-linked dependencies. Actually the problem is slightly worse, because we can have multiple providers each with their own EMFs. This is _very different_ from the situation faced by other specs such as security and JTA!

So I would analogize _our_ situation to the problem faced by the Java compiler and annotation processors. There, the solution is to process in "rounds", where at each round, the processing of a single component is either a success or a failure, meaning it will be reprocessed by the next round. I imagine that a "complete" solution to this problem would involve CDI adopting a similar multiphase lifecycle for bean registration. But that's a problem best solved in CDI, not here.

I'm sensitive to the fact that @arjantijms likes his producer method based on `InjectionPoint`, and I quite like it too, but it seems to me that it works for him _precisely because_ there's only one `Extension` registering JPA beans, and that if, as he suggests, we were to move this responsibility to the providers, then I predict that his solution immediately breaks. Now, I might be wrong, and I know there's ways to make it work, but they're not quite so clean.

Anyway, in my opinion, _the very fact that this discussion has even gone this far_ indicates to me that the "how to implement" part of this is not something that's ripe for standardization. Indeed, I think if we were to impose this requirement on providers (instead of on the EE container), I would be very fearful of all the complaints we would get that such-and-such persistence provider _P_ works in container _C_ but not in _D_, and could we please clarify what _C_ and _D_ are supposed to do in tricky corner case _X_.

Let's not go there. At least not now. One advantage of saying nothing about how this requirement is _implemented_, is that we can always decide to say something later, when we understand the problem better. Who knows, perhaps CDI will eventually offer something like Steve's and Scott's `ExtendedBeanManager`, or like my `javac`-inspired round-based processing.

arjantijms:
> Actually the problem is slightly worse, because we can have multiple providers each with their own EMFs. 

Just out of curiosity, by your understanding or knowledge, does anyone ever actually use that? The feature is intended I understand that e.g. JBoss EAP can ship with both Hibernate and EclipseLink, and subsequently define two persistence contexts; one for Hibernate, and one for EclipseLink.

Applications would then arbitrarily (based on their requirements) persists a given entity with either Hibernate or EclipseLink.

I have been using JPA in production since before it was officially released (yes, I know), and I don't think I've ever seen any customer use that.

arjantijms:
> Our preference has not been to implement this via lazy resolution since that essentially circumvents CDI's dependency validation (detection of ambiguous/unsatisfiable dependencies).

I've been thinking about this indeed, and you are of course right. Still, I wonder, can't we do validation as soon as the EntityManagerFactory / persistence context comes online?

E.g. in a condensed timeline:

```
t=1 -> CDI container boots
t=2 -> Extensions add a delegating entity manager Bean (e.g. via producer or custom Bean<t>)

t=3 -> Jakarta Persistence boots
t=4 -> EntityManagerFactory / persistence context comes online / becomes available 

t=5 -> Validation of previously added entity manager Bean(s) take place

t=6 -> Servicing of requests start
```

arjantijms:
> Extension registering JPA beans, and that if, as he suggests, we were to move this responsibility to the providers, then I predict that his solution immediately breaks. Now, I might be wrong, and I know there's ways to make it work, but they're not quite so clean.

Just thinking out loud, and I haven't investigated, but... 

There can only ever be air most one default persistence context (entity manager factory) and persistence unit (entity manager). This should be true regardless of how many Jakarta Persistence implementations there are on the class path.

All the other usages have to be qualified. For the persistence unit that is only the `name` (`unitName` in @PersistenceUnit). Names need to be unique already, so a Jakarta Persistence implementation can only ever add a qualified bean for a unique name. Should a duplicate name be added, it blows up. As intended.

gavinking:
> Just out of curiosity, by your understanding or knowledge, does anyone ever actually use that?

I mean, not explicitly, but I imagine it would be a pretty reasonable thing to do if you're in the process of migrating from one provider to another. And it's certainly a core feature of JPA.

> Still, I wonder, can't we do validation as soon as the EntityManagerFactory / persistence context comes online?

I mean, I imagine there are ways to do it. But that's pretty different to saying that I know this can be done cleanly with no tradeoffs in every single EE container, and that it's an approach I want the spec to _mandate_.

> Names need to be unique already

In CDI we don't inject on the basis of name. We inject by matching type + qualifiers.

gavinking:
(To be clear, I think disambiguating injection points using `unitName` is bad and untypesafe.)

arjantijms:
> (To be clear, I think disambiguating injection points using unitName is bad and untypesafe.)

I don't disagree, but what I meant is that any qualifier eventually must reference a name (unitName), which must ultimately be unique. So wouldn't that make clashes between different implementations unlikely, or even impossible?

arjantijms:
> I mean, not explicitly, but I imagine it would be a pretty reasonable thing to do if you're in the process of migrating from one provider to another. And it's certainly a core feature of JPA.

I dunno, I would migrate in one go really. It feels quite odd to migrate and then persist say entity `User` via Hibernate, but still persist `Invoice` using OpenJPA. Is that even possible if `Invoice` referenced `User`?

I also wonder why only Jakarta Persistence has this. Never heard the wish for say Faces to have both Mojarra and MyFaces at the same time active, and then render one part of a page with Mojarra, and another part with MyFaces.

arjantijms:
> I mean, I imagine there are ways to do it. But that's pretty different to saying that I know this can be done cleanly with no tradeoffs in every single EE container, and that it's an approach I want the spec to mandate.

Perhaps in the simplest case, we can add a spec demand for the container / jakarta persistence implementation to also validate `@Inject EntityManager` at the same it also validated `@PersistenceUnit EntityManager`?

I mean, it now already has to validate `@PersistenceUnit EntityManager`. What exactly would hold whatever does that validation now, to optionally, also validate `@Inject EntityManager`?

lukasj:
> > I mean, not explicitly, but I imagine it would be a pretty reasonable thing to do if you're in the process of migrating from one provider to another. And it's certainly a core feature of JPA.
> 

> I also wonder why only Jakarta Persistence has this. Never heard the wish for say Faces to have both Mojarra and MyFaces at the same time active, and then render one part of a page with Mojarra, and another part with MyFaces.

It is NOT only Persistence who has this. Counter example is XML Binding (formerly JAXB) - even GF has jaxb-ri _and_ MOXy where both do implement the same spec for quite a long time. Not saying there was another runtime in the JDK (up to 10). And everything has been available through one API, even though it may be tricky to get the right implementation (rest endpoint with moxy and soap endpoint with jaxb-ri in one war is valid and expected to be supported use-case). 

The motivation and sort of a rule behind this built-in support - as I remember it (...but I can be wrong) - is - If a spec targets Java SE - and Persistence as well as XML Binding (and few others) do - then it must allow and support co-existence of multiple independent implementations.

arjantijms:
> even GF has jaxb-ri and MOXy where both do implement the same spec for quite a long time.

Well...   

```xml
        <!-- Provides conflicting JAXB provider.
             See https://github.com/eclipse-ee4j/eclipselink/issues/1380
        <dependency>
            <groupId>org.eclipse.persistence</groupId>
            <artifactId>org.eclipse.persistence.moxy</artifactId>
            <exclusions>
                <exclusion>
                    <groupId>*</groupId>
                    <artifactId>*</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
        -->
```

arjantijms:
>  then it must allow and support co-existence of multiple independent implementations in one JVM.

Just like with the machinery behind EARs, I wonder if we didn't jump the shark here with such features.

scottmarlow:
> > then it must allow and support co-existence of multiple independent implementations in one JVM.
> 
> Just like with the machinery behind EARs, I wonder if we didn't jump the shark there with such features.

For those that didn't watch the TV show Happy Days and haven't heard the term of `jumping the shark`, defining reference is https://en.wikipedia.org/wiki/Jumping_the_shark which describes how the Happy Days TV show ended when character Arthur Fonzarelli jumped over a shark. :-)

arjantijms:
> ended when character Arthur Fonzarelli jumped over a shark. :-)

Slightly off topic, but I loved watching that show, must have seen every episode. The shark episode was...troublesome though :| ;)

starksm64:
So where integration requirements are defined and where the associated TCK tests are produced has been an ongoing discussion that is coming to head in EE 11. In discussions at the platform group level, the feeling is that this needs to done as part of a platform profile or the platform spec and associated TCK. I'll take this work as the basis for a new Persistence/CDI integration spec section in the Web profile with associated TCK tests. It seems possible that there need not be any changes to the Persistence specification, but that needs to be verified.

I'll also propose that we update the specification operation process to require that any integration specs and TCKs have a sign off from the component specification team(s) that are impacted.

arjantijms:
> I'll also propose that we update the specification operation process to require that any integration specs and TCKs have a sign off from the component specification team(s) that are impacted.

I still find it very troublesome that the container / runtime needs to have such intimate knowledge about Jakarta Persistence that only the container can create an `EntityManager`. 

Jakarta Persistence implementations should really be able to do this themselves. If they can't, then I strongly suspect the problem is elsewhere, and handing this problem off to the container is just some kind of hack.

For comparison again, in Jakarta Faces, the implementation can create a `FacesContext`. It doesn't need the container for this. In Jakarta Security, the implementation can create an `HttpAuthenticationMechanism` it doesn't need the container for this, etc for quite a number of other specifications.

mnriem:
I like delegating approach simply because for most applications this would be sufficient and thus enables developer productivity. One could narrowly define this way of integration only scoped for applications with just one persistence unit. Thoughts?

mnriem:
@gavinking What is missing in the public EE API surface so CDI integration can be done in a vendor agnostic way?

sebersole:
> the EMF needs beans from CDI for stuff like injection into entity listeners, but CDI beans need to inject EMFs

To be clear, the EMF needs the ability to get the entity listeners themselves from CDI.  Minor distinction, but the cause of the problem.

> I like the delegating approach

Which is the "delegating approach"?

> What is missing in the public EE API surface so CDI integration can be done in a vendor agnostic way?

In the solution between Hibernate and Wild Fly this boiled down to the persistence provider being able to get a *scoped* callback that the `BeanManager` is ready to use.  The scoped part is important to be able to restrict this between singular EMF and `BeanManager` instances.

gavinking:
> @gavinking What is missing in the public EE API surface so CDI integration can be done in a vendor agnostic way?

Hi, I'm not quite sure what you're asking.

mnriem:
> > I like the delegating approach
> 
> Which is the "delegating approach"?

The approach where the injected EntityManager is merely a proxy to the real thing. 
> 
> > What is missing in the public EE API surface so CDI integration can be done in a vendor agnostic way?
> 
> In the solution between Hibernate and Wild Fly this boiled down to the persistence provider being able to get a _scoped_ callback that the `BeanManager` is ready to use. The scoped part is important to be able to restrict this between singular EMF and `BeanManager` instances.

So is the ServletContainerInitializer API sufficient to accomplish this or is there a need for more API surface?

arjantijms:
> To be clear, the EMF needs the ability to get the entity listeners themselves from CDI. Minor distinction, but the cause of the problem.

Could you explain this in more detail? Why is this exactly a problem?

```java
public class HibernateCdiExtension implements Extension {

    public void afterBean(final @Observes AfterBeanDiscovery afterBeanDiscovery, BeanManager beanManager) {

        afterBeanDiscovery.addBean()
                .scope(...)
                .types(EntityManager.class)
                .id("some id")
                .createWith(cc -> HibernateFactory.create(...).createEntityManager());
}
```

And suppose we have:

```java
@EntityListeners(UserListener.class)
@Entity
public class User {
    //...
}
```

and

```java
public class UserListener {
    
    @Inject
    Foo foo;
    
    @PrePersist   
    private void beforeAnyUpdate(User user) {
        foo.check(user);
    }
```

In terms of these classes, where do things exactly fail? 

In Jakarta Security for instance it also happens that a bean added in `AfterBeanDiscovery` executes code in some path that itself needs the CDI API, either programmatically using the beanManger lookup methods or otherwise.

sebersole:
Sorry for the late reply, I've been traveling.

>  Consider the following pseudo extension code:

So long as the `createWith` callback happens only once the BeanManager is ready to use, then that should work.  I'm not familiar enough with CDI at that level to say.

sebersole:
Oh wait.. I (may) take that back.  

Is the expectation of your pseudo code to do that for each bean that wants EMF injection?  If so, no that's not really feasible.

arjantijms:
> So long as the createWith callback happens only once the BeanManager is ready to use, then that should work.

Of course, that's how CDI works. CDI will not create beans with CDI before CDI itself is ready to use.

> Is the expectation of your pseudo code to do that for each bean that wants EMF injection? If so, no that's not really feasible.

No, the create() method is only executed once per scope. Within that scope all beans will use the same instance of the entity manager (not EMF in this example) here.

It's only pseudo code. The `EntityManagerFactory` itself, I guess, will be either statically available, or reside within application scope.

Updated example:

```java
public class HibernateCdiExtension implements Extension {

    public void afterBean(final @Observes AfterBeanDiscovery afterBeanDiscovery, BeanManager beanManager) {

        afterBeanDiscovery.addBean()
                .scope(ApplicationScoped.class)
                .types(EntityManagerFactory.class)
                .id("some id")
                .createWith(cc -> HibernateFactory.create(...));

         afterBeanDiscovery.addBean()
                .scope(RequestScoped.class)
                .types(EntityManager.class)
                .id("some id")
                .createWith(cc -> 
                        CdiUtils.getBeanReference(EntityManagerFactory.class)
                                     .createEntityManager()
                );
}
```

beikov:
The problem is with enhancement of JPA entity classes. Hibernate/JPA has to register a `ClassTransformer` before any user code can run, which is the first phase of the Hibernate bootstrap. Then, after CDI booted completely i.e. when the `AfterDeploymentValidation` event is fired and the `BeanManager` can be used to retrieve beans, the second phase of the Hibernate bootstrap has to be initiated, which is to actually eagerly build an `EntityManagerFactory`.

scottmarlow:
Am I correct that as the title mentions, this is for the EE/JPA container to implement?  Or is it better for each Persistence Provider to implement as per more recent comments?

arjantijms:
> Or is it better for each Persistence Provider to implement as per more recent comments?

To my mind it would be much better if the Persistence Provider (Jakarta Persistence implementation) implemented this. E.g. Hibernate, OpenJPA and EclipseLink should be able to implement this. 

If this *has* to be implemented by the EE container's internal integration code (e.g. that of WildFly, TomEE, GlassFish) then IMHO it would mean the Jakarta EE APIs (Jakarta CDI, Jakarta Persistence, etc) are missing something.

scottmarlow:
> > Or is it better for each Persistence Provider to implement as per more recent comments?
> 
> To my mind it would be much better if the Persistence Provider (Jakarta Persistence implementation) implemented this. E.g. Hibernate, OpenJPA and EclipseLink should be able to implement this.
> 

I'm not exactly sure of how Hibernate, EclipseLink, OpenJPA, DataNucleus would deal with the various chicken/egg timing issues that require specialized EE/JPA container code for doing what the EE containers have been doing for a long time in terms of deployment (WildFly does work better with Hibernate due to supported extensions as Steve mentioned earlier).  Thanks for helping me to remember this. :-)

arjantijms:
> I'm not exactly sure of how Hibernate, EclipseLink, OpenJPA, DataNucleus would deal with the various chicken/egg timing issues that require specialized EE/JPA container code for doing what the EE containers have been doing for a long time 

Well, for one, I'm still not convinced that such "specialized EE/JPA container code for doing what the EE containers have been doing for a long time" should be absolutely needed. What is that specialised code *really* doing when you take a closer look?

I know we have it in GlassFish, but when reimplementing integration in Piranha I found there barely was a need for it. The Jakarta Persistence implementation (EclipseLink in our case) could do everything just as well. The only small detail was that it defaults to SE mode, where the transactional settings are different by default. A simple patch lets it return the persistence unit in EE mode, without any involvement of tons of Piranha internal code.

But actually, that all is another issue; it's why containers historically needed (or think they need?) tons and tons of container specific code.

The issue here is much smaller than that. 

This issue doesn't seek to get rid of all that integration code. It just asks why Jakarta Persistence implementations can't take care of the `@Inject EntityManager` case.

scottmarlow:
> Well, for one, I'm still not convinced that such "specialized EE/JPA container code for doing what the EE containers have been doing for a long time" should be absolutely needed. What is that specialised code _really_ doing when you take a closer look?
> 

In addition to parsing and dealing with the persistence.xml to prepare for calling https://jakarta.ee/specifications/persistence/3.1/apidocs/jakarta.persistence/jakarta/persistence/spi/persistenceprovider#createContainerEntityManagerFactory(jakarta.persistence.spi.PersistenceUnitInfo,java.util.Map) and dealing with https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#container-responsibilities, WildFly is coordinating the right time for certain other things that Steve mentioned above that there are separate issues for standardizing (or not if others aren't interested which was mostly the case 8 years ago except for a few vendors).  The extensions deal with calling a Hibernate callback after the CDI Bean Manager is started and available for use.  As well as coordinating with application defined datasource (e.g. [WFLY-2727](https://issues.redhat.com/browse/WFLY-2727)) initialization in a way that ensures that bytecode enhancement occurs.  

> I know we have it in GlassFish, but when reimplementing integration in Piranha I found there barely was a need for it. The Jakarta Persistence implementation (EclipseLink in our case) could do everything just as well. The only small detail was that it defaults to SE mode, where the transactional settings are different by default. A simple patch lets it return the persistence unit in EE mode, without any involvement of tons of Piranha internal code.

IMO, that would be mostly a SE mode type integration that works with JTA which is interesting but not exactly a new thing as EE applications can do that via SE mode bootstrapping, except you are dealing with the injection but not in a EE compatible way.

> 
> But actually, that all is another issue; it's why containers historically needed (or think they need?) tons and tons of container specific code.

You are comparing EE vs SE application requirements but still there isn't `tons and tons of container specific code`, but there is a contract that defines the EE/Persistence container contract versus the persistence provider contract that is important for EE application compatibility, which I think we would lose if we follow what you pushing for here.

> 
> The issue here is much smaller than that.
> 
> This issue doesn't seek to get rid of all that integration code. It just asks why Jakarta Persistence implementations can't take care of the `@Inject EntityManager` case.

I think applications would lose the benefits of https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#container-responsibilities which means their applications won't work the same.  Application compatibility is as I understand it, very important.

arjantijms:
> In addition to parsing and dealing with the persistence.xml

I'm not really sure why the container has to do this. Why would an implementation not be capable of parsing persistence.xml?

In Jakarta Faces, it's the implementation that does exactly this. E.g. Mojarra or MyFaces parses faces-config.xml.

arjantijms:
> IMO, that would be mostly a SE mode type integration that works with JTA which is interesting but not exactly a new thing as EE applications can do that via SE mode bootstrapping, except you are dealing with the injection but not in a EE compatible way.

"the injection but not in a EE compatible way.". Could you elaborate on that? 

In a Jakarta Faces implementation we also do injection in non-CDI beans (injecting CDI, EJB, Resource), but we don't need to have knowledge of Jakarta Faces in the container.

(we did had some SPIs that the container could provide implementations of for scanning and injection, but these are barely needed anymore these days)

arjantijms:
> I think applications would lose the benefits of https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#container-responsibilities which means their applications won't work the same.

I don't understand that at all.

First if `@Inject EntityManager` would only be a small wrapper over getting the EntityManager "the old way", why would anything change? Which benefits would applications lose, and why would they not work the same?

Injecting the EntityManager via CDI is a new feature, so how can anything change? That code would always be new, right?

And secondly, the "container-responsibilities" seem rather mundane anyway. With CDI, a transaction scope, and Interceptors around, why can't the Jakarta Persistence implementation not simply create and close an entity manager? Why is this something a container would only be able to do?

I've been scanning a little through the integration code in GlassFish, and most things seem to be from the early JPA days and focussed on EJB (the "container-responsibilities" text also uses EJB terminology).

If there might be timing or ordering issues with interceptors and transactions, should we not focus on fixing these? This is what I meant earlier with IFF something would only be possible at the container level, it might signal a bug or omission in the Jakarta APIs.

sebersole:
> It's only pseudo code. The `EntityManagerFactory` itself, I guess, will be either statically available, or reside within application scope.

I mean, if the expected "solution" is to hold the EMF in a static variable (singleton) or some such, I just don't see that as a solution.

sebersole:
> The problem is with enhancement of JPA entity classes. Hibernate/JPA has to register a `ClassTransformer` before any user code can run, which is the first phase of the Hibernate bootstrap. Then, after CDI booted completely i.e. when the `AfterDeploymentValidation` event is fired and the `BeanManager` can be used to retrieve beans, the second phase of the Hibernate bootstrap has to be initiated, which is to actually eagerly build an `EntityManagerFactory`.

That's a different topic.  

How the CDI integration works is what's being discussed here.  The problem there is not enhancement.  It's a question of handling the chicken-egg problem of CDI needs an EMF, but the EMF needs CDI

scottmarlow:
> @arjantijms I am referring to the current Persistence contracts which [I linked](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#container-responsibilities). It sounds like you are more in favor of applications using `application-managed entity manager` via `@Inject EntityManagerFactory` which gives the application the same full control over any created EntityManager as discussed in https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a11465

Responding to myself to state that I am favor of the current pull request which refers to `container-managed entity manager` being available for injection via CDI.  For applications that do not want a `container-managed entity manager` should inject a EntityManagerFactory instead so they can use a `application-managed entity manager`.

scottmarlow:
I'm giving my +1 approval of this pr and https://github.com/jakartaee/platform/pull/746 as I like the idea of the change, regardless of which pull request is used.

scottmarlow (approved):

arjantijms:
> I am referring to the current Persistence contracts which [I linked](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#container-responsibilities). It sounds like you are more in favor of applications using `application-managed entity manager` via `@Inject EntityManagerFactory` which gives the application the same full control over any created EntityManager as discussed

No, that's something else really.

I'm just wondering why the implementation code for the existing contracts HAS to reside within WildFly, GlassFish, etc and cannot reside in Hibernate, EclipseLink, etc.

I guess the "simple" reason is that when these contracts were designed, there were no public APIs in Java EE to do what they asked. So it had to be implemented with integration code.

But with today's transaction scope and CDI, it's much easier to create an entity manager once during a transaction upon first access. That is exactly what the create() method of a Bean<T> does. Same for closing it.

arjantijms:
> It's a question of handling the chicken-egg problem of CDI needs an EMF, but the EMF needs CDI

Is that really a chicken-egg problem?

If a Bean<T> creates an EntityManager using the EnitityManagerFactory, CDI is available at that point, and the EnitityManagerFactory can use it. The resulting EntityManager can then be returned.

In Jakarta Security a similar thing happens. CDI needs an AuthenticationMechanism, and the AuthenticationMechanism needs CDI:

```java
authenticationMechanismBean = new CdiProducer<HttpAuthenticationMechanism>()
                    .scope(ApplicationScoped.class)
                    .types(Object.class, HttpAuthenticationMechanism.class)
                    .addToId(FormAuthenticationMechanismDefinition.class)
                    .create(e -> {
                        FormAuthenticationMechanism authMethod = CdiUtils.getBeanReference(FormAuthenticationMechanism.class);

                        authMethod.setLoginToContinue(
                            LoginToContinueAnnotationLiteral.eval(formAuthenticationMechanismDefinition.loginToContinue()));

                        return authMethod;
                    });
```

arjantijms:
> Responding to myself to state that I am favor of the current pull request which refers to container-managed entity manager being available for injection via CDI.

It might be that you're mixing up thing here.

It's not about the difference between a container-manager entity manager and an application-managed entity managed.

The difference is where the implementation code to handle a container-manager entity manager should or can reside, and specifically if an implementation like Hibernate is capable of providing a container-manager entity manager via `@Inject`.

If it's really true that only special container code can add a Bean for `@Inject` to inject, and a component library such as Hibernate cannot use the CDI API for this, we really have other problems in our APIs.

But I simply don't think this is the case.

I think we're just misunderstanding each other here.

arjantijms:
> > It's only pseudo code. The `EntityManagerFactory` itself, I guess, will be either statically available, or reside within application scope.
> 
> I mean, if the expected "solution" is to hold the EMF in a static variable (singleton) or some such, I just don't see that as a solution.

It was just an example. It has to be kept "somewhere". Where do you keep the `EntityManagerFactory` currently?

arjantijms:
> > IMO, that would be mostly a SE mode type integration that works with JTA which is interesting but not exactly a new thing as EE applications can do that via SE mode bootstrapping, except you are dealing with the injection but not in a EE compatible way.
> 
> "the injection but not in a EE compatible way.". Could you elaborate on that?

@scottmarlow I feel this is still a very important question. What exactly is "injection in an EE compatible way" here?

scottmarlow:
> > > IMO, that would be mostly a SE mode type integration that works with JTA which is interesting but not exactly a new thing as EE applications can do that via SE mode bootstrapping, except you are dealing with the injection but not in a EE compatible way.
> > 
> > 
> > "the injection but not in a EE compatible way.". Could you elaborate on that?
> 
> @scottmarlow I feel this is still a very important question. What exactly is "injection in an EE compatible way" here?

Is it okay if we discuss on the Persistence mailing list as I think others may be interested in the discussion and I think we are getting off topic with regard to this issue IMO.

arjantijms:
> Is it okay if we discuss on the Persistence mailing list as I think others may be interested in the discussion and I think we are getting off topic with regard to this issue IMO.

Sure, looking forward to the discussion there :)

beikov:
The problem is that no JPA managed type class may be loaded until the JPA provider has called `PersistenceUnitInfo#addTransformer`, because that `ClassTransformer` is what does the bytecode enhancement on demand when loading classes.
This method is called from within `PersistenceProvider#createContainerEntityManagerFactory` which creates the entity manager factory and hence validates and boots up the JPA provider. During that boot process, the JPA provider must be able to access beans through the `BeanManager` for CDI integration to work.

Since CDI potentially loads JPA managed classes, it must be ensured that `PersistenceUnitInfo#addTransformer` is called before that happens, so the JPA bootstrap must happen first, before CDI does any class loading. 
This means the JPA bootstrap will run with a "partial" `BeanManager` and it is not possible to lookup beans. This is the reason why there exists a special integration in Wildfly and potentially also other containers.

mnriem:
My understanding here is that @arjantijms is asking for the extension / integration points to be formalized so how the integration is done is specified and well understood from the perspective of an integrator.

lukasj (approved):
Yes, this is good. I'll fix few minor things (like version of the schema in the xsd) in upcoming PR

gavinking:
Thanks man.

Issue #459, opened 2023-08-12
named queries, entity graphs, and result set mappings in static metamodel generation


gavinking:
A very nice&mdash;and extremely underused&mdash;feature of the static metamodel is the ability to write:

```java
@OneToMany(mappedBy=Item_.ORDER)
Set<Item> items;
```

which is a lot more typesafe than:

```java
@OneToMany(mappedBy="order")
Set<Item> items;
```

But we could take this further. We could have the static metamodel include string-valued constants for named queries, named entity graphs, and even SQL result set mappings.

So you could write, for example:

```java
em.createQuery(Book_.QUERY_ALL_BOOKS).getResultList();
```

Hibernate 6.3 already supports this.

gavinking:
Oh, wait, I messed up in describing this.

> A very nice—and extremely underused—feature of the static metamodel

Actually that "very nice" feature is a feature of our implementation, which is not actually mandated by section 6.2.1.1 of the specification.

But it should be!

gavinking:
Please check out this proposal: https://github.com/jakartaee/persistence/pull/461/commits/ff3f6f9328333cd75004254d4ec54b07c31dc22d

PR #461, opened 2023-08-12
specify static string generation for metamodel classes

gavinking:
and add reference to `ManagedType` to static metamodel

see #459, #475


lukasj (approved):
LGTM

Issue #47, opened 2013-02-11
custom value mappings for @Enumerated


lukasj:
In mapping a pre-existing schema, it is rarely possible to map a given set of constants using an @Enumerated:

*   Ordinals are non-sequential
*   Names are incompatible with Java Enum Constants
*   Names cannot be aliased to something more useful or conventional for Java Naming standards

Limitation is because specification states the name() will be used of an enum.
A typical example: Database uses 'N/A', which can never be mapped to an enum. Intuition will let us look for a solution to map this to 'NOT_AVAILABLE', both a valid name for an Enum Constant, and also conventional naming.

On top of this, an enum often has a representation in the Database, and a constant has usually other additional attributes that could be useful.
Better than using the name() of an @Enumerated, we should map the enumerated similar as an @Entity, and use the @ID.

gavinking:
So this very long discussion is quite old now, and doesn't seem to have ever reached much of a firm conclusion.
So let me give my take on this.

As pointed out by others, there's quite a lot of overlap with `AttributeConverter` here. _However_, there's one thing that `AttributeConverter`s aren't really set up to handle, and that's DDL generation. For a JPA `@Enumerated` field, we want to automatically generate a `check` constraint or a database-native `ENUM` type. To do that, we need to know upfront what values are possible.

So there's indeed a good case to me made for allowing explicit "remapping" of `enum` values.

Just off the top of my head, I can think of two ways to do this. The first is the one that seems to have been considered here:

```java
enum Status {
    OPEN, CLOSED, CANCELLED;

    @EnumeratedValue
    int toInt() {
        switch (this) {
            case OPEN: return 0;
            case CLOSED: return 1;
            case CANCELLED: return -1;
            default: throw new RuntimeException();
        }
    }
}
```

The problem with this is it makes the enumerated values implicitly dynamic. There's no way to force `toInt()` to return the same `int` value for a given enum value.

But there's a different way, which was mentioned in [this comment](https://github.com/jakartaee/persistence/issues/47#issuecomment-417722089):

```java
enum Status {
    @EnumeratedIntValue(0)
    OPEN, 
    @EnumeratedIntValue(1)
    CLOSED, 
    @EnumeratedIntValue(-1)
    CANCELLED;
}
```

That's shorter, and more "correct". It's the solution I've already proposed that we implement in Hibernate.

The downside to this approach is that we would need two annotations (`@EnumeratedIntValue` and `@EnumeratedStringValue` or whatever).

[...]

PR #463, opened 2023-08-13
introduce @EnumeratedValue and actually specify what ORDINAL and STRING mean

gavinking:
see #47

### DIFF
--- a/api/src/main/java/jakarta/persistence/EnumeratedValue.java
+++ b/api/src/main/java/jakarta/persistence/EnumeratedValue.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright (c) 2008, 2020 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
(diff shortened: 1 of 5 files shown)


api/src/main/java/jakarta/persistence/EnumeratedValue.java
```diff
@@ -0,0 +1,63 @@
...
+ *       }
+ *   }
+ * </pre>
+ *
+ * @see Enumerated
+ * @see EnumType
+ *
+ * @since 3.2
+ */
+@Target({FIELD})
```
beikov, line 60: IMO also a `METHOD` should be allowed, as I think most enums will have such a method anyway, whereas I'm not sure if it will have a field.

gavinking: Well the proposal (and it's still just a proposal) was to say that it has to annotate a final field, so that it's enforced to be a constant value for each instance of the enum. A method can, in principle, return a different value each time it's called.

beikov: >A method can, in principle, return a different value each time it's called.

Sure, but JPA also supports `Access.PROPERTY` which could suffer from the same problem. In reality, users can always shoot their foot, but trying to handle all such cases doesn't seem practical to me.

lukasj (approved):
LGTM

Issue #128, opened 2016-03-29
Support JOIN/ON for two root entities


lukasj:
Currently JPA does not allow root entities after various JOIN clauses which means that especially outer joins cannot be specified and executed without association path. In some cases it may be beneficial when we can join even when the relation is not mapped, e.g. to avoid potentially EAGER One- or @ManyToOne relationship (LAZY is only hint, not guaranteed).

EclispeLink currently supports such joins and they add a lot of options how to solve some problems with a minimal JPQL change. "EclipseLink also supports usage of the ON clause between two root level objects." See: [https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL#ON](https://wiki.eclipse.org/EclipseLink/UserGuide/JPA/Basic_JPA_Development/Querying/JPQL#ON)

Please, consider root entites as candidates for joins.

gavinking:
To be clear, we're talking about queries like this:

```sql
select book.title, publisher.name
from Book book
    join Publisher publisher
        on book.publisher = publisher
where book.title like :titlePattern
```

Correct?

If so, then this is something that both EclipseLink and Hibernate have supported for quite a long time, and it's something that should be quite trivial to add to the spec.

I would say we can easily do this for 3.2.

Issue #71, opened 2014-01-27
Add subquery(EntityType) to javax.persistence.criteria.CriteriaQuery


lukasj:
If using dynamic entities not defined by classes (as is the case when using EntityMode.MAP), there's no way to create a subquery. This is because javax.persistence.criteria.CriteriaQuery.subquery() takes an entity Class as an argument, and unlike javax.persistence.criteria.CriteriaQuery.from() there's no overloaded method to subquery by EntityType. This severely limits what can be done using entities mapped with EntityType for which there is no Java class.

gavinking:
This is a perfectly sensible and straightforward request. 

I've added the missing overload to `CommonAbstractCriteria` in #71.

lukasj:
done in #467

PR #467, opened 2023-08-14
add joins to entity types (range variables)

gavinking:
see #128,  #71

### DIFF
--- a/spec/src/main/asciidoc/ch04-query-language.adoc
+++ b/spec/src/main/asciidoc/ch04-query-language.adoc
@@ -303,9 +303,16 @@ identification_variable_declaration ::= range_variable_declaration {join | fetch
 
 range_variable_declaration ::= entity_name [AS] identification_variable
 
-join ::= join_spec join_association_path_expression [AS] identification_variable [join_condition]
+join ::= range_join | path_join
+
+range_join ::= join_spec range_variable_declaration [join_condition]
+
+path_join ::=
+    join_spec join_association_path_expression [AS] identification_variable [join_condition]
 
 fetch_join ::= join_spec FETCH join_association_path_expression
+
+join_spec ::= [INNER | LEFT [OUTER]] JOIN
 
 join_association_path_expression ::=
     join_collection_valued_path_expression |
@@ -316,8 +323,6 @@ join_association_path_expression ::=
 join_collection_valued_path_expression ::= identification_variable.{single_valued_embeddable_object_field.}*collection_valued_field
 
 join_single_valued_path_expression ::= identification_variable.{single_valued_embeddable_object_field.}*single_valued_object_field
-
-join_spec ::= [LEFT [OUTER] | INNER] JOIN
 
 join_condition ::= ON conditional_expression
 
@@ -433,7 +438,7 @@ query (or subquery) in which it is defined and is also visible to any
 subqueries within that query scope that do not define an identification
 variable of the same name.
 
-==== Range Variable Declarations
+==== Range Variable Declarations [[a4766]]
 
 The syntax for declaring an
 identification variable as a range variable is similar to that of SQL;
@@ -686,22 +691,43 @@ SELECT DISTINCT l.product
 FROM Order AS o JOIN o.lineItems l
 ----
 
-It is illegal to use a
-_collection_valued_path_expression_ other than in the FROM clause of a
-query except in an _empty_collection_comparison_expression,_ in a
-_collection_member_expression_, or as an argument to the SIZE operator.
+A _collection_valued_path_expression_ may only occur in:
+
+- the FROM clause of a query,
+- an _empty_collection_comparison_expression,_
+- a _collection_member_expression_, or
+- as an argument to the SIZE operator.
+
 See <<a5139>>, <<a5150>>, and <<a5284>>.
 
 ==== Joins
 
-An inner join may be implicitly specified by
-the use of a cartesian product in the FROM clause and a join condition
-in the WHERE clause. In the absence of a join condition, this reduces to
-the cartesian product.
+JPQL defines the following varieties of join:
 
-The main use case for this generalized style
-of join is when a join condition does not involve a foreign key
-relationship that is mapped to an entity relationship.
+- inner joins,
+- left outer joins,
+- right outer joins,
+- full outer joins.
+
+The semantics of each variety of join is identical to SQL, and the
+syntax is borrowed from ANSI SQL. Providers are not required to
+support right joins or full joins on databases which do not natively
+support these join types.
+
+Every join has a target, either:
+
+- an entity-valued path expression, or
+- an entity type (that is, range variable declaration, as already
+  specified in <<a4766>>).
+
+An inner join may be implicitly specified by the use of a cartesian
+product in the FROM clause and a join condition in the WHERE clause.
+In the absence of a join condition, this reduces to the cartesian
+product.
+
+The main use case for this generalized style of join is when a join
+condition does not involve a foreign key relationship mapped to an
+association between entities.
 
 Example:
 
@@ -710,17 +736,22 @@ Example:
 SELECT c FROM Customer c, Employee e WHERE c.hatsize = e.shoesize
 ----
 
-In general, use of this style of inner join
-(also referred to as theta-join) is less typical than explicitly defined
-joins over relationships.
+This style of inner join (sometimes called a "theta" join) is less
+typical than explicitly defined joins over relationships.
 
-The syntax for explicit join operations is as
-follows:
+The syntax for explicit join operations is given by:
 
 ----
-join ::= join_spec join_association_path_expression [AS] identification_variable [join_condition]
+join ::= range_join | path_join
+
+range_join ::= join_spec range_variable_declaration [join_condition]
+
+path_join ::=
+    join_spec join_association_path_expression [AS] identification_variable [join_condition]
 
 fetch_join ::= join_spec FETCH join_association_path_expression
+
+join_spec ::= [INNER | LEFT [OUTER]] JOIN
 
 join_association_path_expression ::=
     join_collection_valued_path_expression |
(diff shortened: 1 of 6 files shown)


spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -274,14 +274,21 @@ from_clause ::=
     FROM identification_variable_declaration
         {, {identification_variable_declaration | collection_member_declaration}}*
 
-identification_variable_declaration ::= range_variable_declaration {join | fetch_join}*
+identification_variable_declaration ::= range_variable_declaration join*
 
 range_variable_declaration ::= entity_name [AS] identification_variable
 
-join ::= join_spec join_association_path_expression [AS] identification_variable [join_condition]
+join ::= range_join | path_join | fetch_join
```
beikov: This change would allow fetch joins in a subquery, which is obviously wrong. See `subselect_identification_variable_declaration`, which uses the `join` rule.

spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -705,51 +734,93 @@ join_collection_valued_path_expression ::=
...
-<<a19497>>. It selects those customers of
+This is equivalent to the following query using the earlier IN
+construct, defined in <<a19497>>. It selects those customers of
 status 1 for which at least one order exists:
 
 [source,sql]
 ----
-SELECT OBJECT(c) FROM Customer c, IN(c.orders) o WHERE c.status = 1
+SELECT OBJECT(c)
+FROM Customer c, IN(c.orders) o
```
beikov, line 848: Can we please deprecate this syntax? It violates the isolation of "from clause spaces". SQL says that these "spaces", separated by a comma, are "isolated" i.e. you can't refer to aliases of a different space. The `collection_member_declaration` rule violates this constraint though and for IMO no good reason. People can just as well write an inner join instead.

spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -681,18 +705,23 @@ Example:
...
+join ::= range_join | path_join | fetch_join
+
+range_join ::= join_spec range_variable_declaration [join_condition]
+
+path_join ::=
+    join_spec join_association_path_expression [AS] identification_variable [join_condition]
 
 fetch_join ::= join_spec FETCH join_association_path_expression
 
+join_spec ::= [INNER | LEFT [OUTER] | RIGHT [OUTER] | FULL [OUTER]] JOIN
```
beikov: The new join types should also be added to the `jakarta.persistence.criteria.JoinType` enum.
While we're at it, we should IMO also add `cross join` so that people don't have to write `a join b on 1=1`.

We might have to adapt some texts that talk about portability of right joins e.g.

> Right outer joins and right outer fetch joins are not required
> to be supported. Applications that use <code>RIGHT</code> join
> types will not be portable.

Either we should add the FULL join type to this portability note or "specify" the behavior to make it a required feature.

beikov (commented):
Not sure if you want to add/propose the Criteria API changes in a separate PR or deliver later after initial discussion, but I wanted to mention for completeness that the API changes are missing. Note that cross and entity joins might need new interfaces similar to what we have in Hibernate, as well as some thoughts on what `From#getJoins` should return, given the fact that `Join` has a `getAttribute()` method. Maybe it's fine to say that `Join#getAttribute()` might return `null` for entity joins. For reference, in Hibernate `CrossJoin` and `EntityJoin` extend `From` to not have to deal with the `ON` clause and `Join#getAttribute()` issues.

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: I have no absolutely no objection to explicitly deprecating it. It was _effectively_ already deprecated by JPA 1.0 (it's a holdover from the older EJB-QL).

But let's make that a separate issue.

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: Ah, OK, thanks for catching that 👍

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: > We might have to adapt some texts that talk about portability of right joins

OK, good point.

gavinking: > The new join types should also be added to the `jakarta.persistence.criteria.JoinType` enum.

Interestingly, `JoinType` already has `RIGHT`, even though the spec never mentions right joins anywhere. But yes, need to add `FULL`.

gavinking: > we should IMO also add `cross join` so that people don't have to write `A a join B b on 1=1`

Why can't they just just write `A a, B b`?

I guess I'm not convinced we need to add `cross join`.

gavinking:
> Not sure if you want to add/propose the Criteria API changes in a separate PR

OK, so please take a look at the _very minimal_ change I just pushed, and let me know what you think is missing.

> Maybe it's fine to say that Join#getAttribute() might return null for entity joins.

That seems _completely_ fine to me.

But I wonder if we should pull `getModel()` up from `Root` to `From`. That's a bit unclear.

spec/src/main/asciidoc/ch04-query-language.adoc
beikov: >Why can't they just just write `A a, B b`?

If you want to be able to refer to `a` in a following join, then you must use a join between `a` and `b`, because the comma separator isolates from clause spaces i.e. it is illegal to write e.g. `from A a, B b join C c on b.x = c.x and a.y = c.y` because `a` is not visible within the second from clause space. Right now, people have to write `from A a join B b on 1=1 join C c on b.x = c.x and a.y = c.y` but I think it might be nicer to allow writing the cross join instead `from A a cross join B b join C c on b.x = c.x and a.y = c.y`.

spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -834,27 +931,27 @@ WHERE p.status = 'inStock'
...
-association or element collection as a side effect of the execution of a
-query.
+A FETCH JOIN clause in a query results in eager fetching of an association
+or element collection as a side effect of execution of the query.
 
-The syntax for a fetch join is
+The syntax for a fetch join is given by:
 
 ----
 fetch_join ::= [LEFT [OUTER] | INNER] JOIN FETCH join_association_path_expression
```
beikov, line 980: This grammar now differs from the BNF i.e. the BNF "allows" right and full join fetches as well. Maybe the BNF should be adapted to include a `fetch_join_spec`?

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: > If you want to be able to refer to a in a following join, then you must use a join between a and b, because the comma separator isolates from clause spaces

Ummmm, which section of the JPA spec says that?

beikov (commented):
There is one more section in chapter 6 talking about right joins.
https://github.com/jakartaee/persistence/blob/efbb3393f632508dcf71f5b0f0c36edcdfa77c49/spec/src/main/asciidoc/ch06-criteria-api.adoc?plain=1#L4122

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: > Maybe the BNF should be adapted to include a `fetch_join_spec`?

I wondered that, but then decided it wasn't necessary, since the text of section 4.4.5.3 says explicitly that right and full joins are disallowed.

But I'm not _against_ adding a separate rule.

spec/src/main/asciidoc/ch04-query-language.adoc
beikov: Actually, it seems the spec even permits this sort of usage due to 4.4.2:

>The identification variable declarations are evaluated from left to right in the FROM clause, and an identification variable declaration can use the result of a preceding identification variable declaration of the query string.

It's a bit of a bummer though, because SQL specifies that the from clause spaces are isolated, and since JPQL builds on SQL, I was assuming it follows this fundamental rule.

AFAIU, this means the comma separator must be rendered to a cross join in SQL to allow the desired semantics. I don't don't know the performance aspects of cross join vs comma separator, but it seems to me that a cross join would have an effect on join ordering, unless the optimizer tries to determine dependencies somehow.

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: > But I'm not against adding a separate rule.

Just especially for you, I made the change.

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: > AFAIU, this means the comma separator must be rendered to a cross join in SQL

Well, you only need to do this transformation in the case where there's a reference to the identification variable before the comma from a join occurring after the comma. Which is not going to be the common case, and seems like it's easy to detect.

gavinking: Well, "easy" might be a slight overstatement. But _possible_, surely.

beikov:
>But I wonder if we should pull `getModel()` up from `Root` to `From`. That's a bit unclear.

`Path`, which `From` extends from, already defines `Bindable<X> getModel()` I think that's good enough.

gavinking:
> > But I wonder if we should pull `getModel()` up from `Root` to `From`. That's a bit unclear.
> 
> `Path`, which `From` extends from, already defines `Bindable<X> getModel()` I think that's good enough.

Ah, cool. Missed that. I guess that's fine then.

spec/src/main/asciidoc/ch04-query-language.adoc
beikov: A quick test with PostgreSQL showed that the comma separator vs. cross join does indeed affect the join order and hence the execution plan might differ when 3+ tables are involved. Here is a nice overview showing the differences: https://stackoverflow.com/a/54738895/412446

It seems very hard (impossible?) to express the following query with the Criteria API

```
from A a cross join B b join C c on ...
```

e.g.

```
CriteriaQuery<?> cq = ...
Root<A> a = cq.from(A.class);
Root<B> b = cq.from(B.class);
Join<C> c = a.join(C.class);
```

unless the `CriteriaQuery` is supposed to have a fixed order of from clause items based on the order in which methods are called. In Hibernate, we currently only order joins within a `Root`.

I'm trying to understand if the cross join support we have in Hibernate is "useless" and that we should implement join ordering for the whole query instead of per `Root`, or if the JPA spec might need cross joins as well.
Right now, it seems to me that we should get rid of cross joins in Hibernate, but I don't yet understand if the join ordering implications of rendering a SQL cross join might have a negative effect on performance. Either way, the spec seems to imply that the SQL must have cross joins semantics, so it seems like a forced move.

gavinking:
@lukasj I have removed all references to right and full joins. Please review. Thanks!

lukasj (approved):
LGTM

PR #469, opened 2023-08-16
aesthetic cleanups to chapter 8

gavinking:
Chapter 8 (entity packaging) was quite difficult to read, mostly due to formatting of the text and property definition. I've cleaned up the formatting, and also made some improvements to language, etc.

There should be no changes to semantics in these edits.


spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -4,127 +4,129 @@
...
+  _persistence.xml_ file, or
+- any combination of the previous options.
+
+The managed persistence classes may be:
+
+- contained within the root of the persistence unit,
+- specified by reference—that is, by naming the classes, class archives, or
+  XML mapping files (which in turn reference classes) that are accessible on
+  the application classpath, or
+- specified by some combination of these means.
```
lukasj: _any_  is more appropriate word here; _some_ may trigger thinking that certain combinations are invalid without actually listing them

lukasj: I mean, based on https://dictionary.cambridge.org/grammar/british-grammar/any

_We use any to mean ‘it does not matter which or what’, to describe something which is not limited. We use this meaning of any with all types of nouns and usually in affirmative sentences._

and https://dictionary.cambridge.org/grammar/british-grammar/some

_We can use this strong form to refer to someone or something particular but unknown, especially with singular countable nouns_

gavinking: I agree, "any" is better here.

gavinking: Fixed.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -168,149 +170,163 @@ Examples:
...
+the entity manager factory for the persistence unit are JTA entity managers or
+resource-local entity managers. The value of this element must be _JTA_ or
+_RESOURCE_LOCAL_:
+
+- _JTA_ means that a JTA data source is provided—either as specified by the
+  _jta-data-source_ element, or by the container.
+- In a Jakarta EE environment, _RESOURCE_LOCAL_ usually means that a non-JTA
+  datasource is provided.
+
+See <<a12302>>.
```
lukasj: this should probably contain some description of the pointer to understand a context about why it is here. Right now it is unclear why should one go there and what one can find there. Something like _For more details about data sources, see ..._ or _See ... for data source configuration_. From a different angle - does the pointer have to be here? I'd say it is good to have it

gavinking: I agree.

gavinking: @lukasj is that better?

lukasj: yes, thanks

lukasj (approved):

Issue #472, opened 2023-08-19
add lockMode to @EntityResult


gavinking:
A native query can pessimistically or optimistically lock an entity in its result set.

It would be nice if the EM could know about this.

We could add:

    LockModeType lockMode() default OPTIMISTIC;

to `@EntityResult`.

Issue #471, opened 2023-08-19
allow nested result set mappings in @NamedNativeQuery


gavinking:
The `@SqlResultSetMapping` annotation defines a named mapping, which is useful when you're passing a SQL query as a string to `EntityManager.createNativeQuery()`.

But when you have a `@NamedNatiiveQuery` it's just annoying and untypesafe.

I therefore propose adding the following members to `@NamedNatiiveQuery`:

```java
    /** Specifies the result set mapping to entities. */
    EntityResult[] entities() default {};

    /** 
     * Specifies the result set mapping to constructors. 
     * @since 2.1
     */
    ConstructorResult[] classes() default {};

    /** Specifies the result set mapping to scalar values. */
    ColumnResult[] columns() default {};
```

Naturally, this is just a copy/paste job of the members of `@SqlResultSetMapping`.

PR #476, opened 2023-08-19
allow result set mappings to be embedded directly in @NamedNativeQuery

gavinking:
and add lockMode to `@EntityResult`.

see #471, #472


gavinking:
Example usage:

```java
@NamedNativeQuery(name = “some”Query,
	query = "select TABLE_NAME as t_name, %s as t_time from ALL_TABLES where TABLE_NAME = 'AUDIT_ACTIONS' “,
	entities = @EntityResult(entityClass = AllTables.class,
				fields = {
						@FieldResult(name = "tableName", column = "t_name"),
						@FieldResult(name = "daysOld", column = "t_time")
				}))
```

Issue #464, opened 2023-08-13
tiny clarification for entity listeners


gavinking:
I propose we clarify in section 3.5.2 that if an entity listener callback method deals with a specific entity type `E`, its parameter may be declared with that type, i.e. `void <METHOD>(E)` instead of `void <METHOD>(Object)`, avoiding a typecast.

PR #478, opened 2023-08-21
light rewriting of section on lifecycle callback methods

gavinking:
tiny clarification for entity listeners

see #464

@lukasj for you, please review

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -2463,39 +2463,41 @@ component.]
 
 ==== Lifecycle Callback Methods
 
-Entity lifecycle callback methods can be
-defined on an entity listener class and/or directly on an entity class
-or mapped superclass.
+Entity lifecycle callback methods can be defined on an entity listener
+class and/or directly on an entity class or mapped superclass.
 
-Lifecycle callback methods are annotated with
-annotations designating the callback events for which they are invoked
-or are mapped to the callback event using the XML descriptor.
+A lifecycle callback method must be either:
 
-The annotations (and XML elements) used for
-callback methods on the entity class or mapped superclass and for
-callback methods on the entity listener class are the same. The
-signatures of individual methods, however, differ.
+- annotated with annotations designating the callback events for which
+  it is invoked, or
+- mapped to a callback event type using the XML descriptor.
 
-Callback methods defined on an entity class
-or mapped superclass have the following signature:
+The same annotations (and XML elements) are used to declare:
 
+- callback methods of an entity class or mapped superclass, and
+- callback methods of an entity listener class.
+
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2507,24 +2509,20 @@ event callback methods of the corresponding types.
...
+* Lifecycle callbacks can invoke JNDI, JDBC, JMS, and enterprise beans.
+* A lifecycle callback method may modify the non-relationship state of
+  the entity on which it is invoked.
+* In general, the lifecycle method of a portable application should not
+  invoke _EntityManager_ or query operations, access other entity
+  instances, or modify relationships within the same persistence
+  contextfootnote:[Note that this caution applies also to the actions of
+  objects that might be injected into an entity listener]footnote:[The
+  semantics of such operations may be standardized in a future release of
+  this specification.].
```
lukasj: was the removal of the last sentence (*) here intentional? I think it's better to be explicit over staying silent here

* _A lifecycle callback method may modify the non-relationship state of the entity on which it is invoked._

gavinking: I didn't remove it, I moved it to its own bullet point tight above.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2507,24 +2509,20 @@ event callback methods of the corresponding types.
...
-specification.]. A lifecycle
-callback method may modify the non-relationship state of the entity on
-which it is invoked.
+* Lifecycle callback methods may throw unchecked/runtime exceptions.
+  A runtime exception thrown by a callback method that executes within
+  a transaction causes that transaction to be marked for rollback if
+  the persistence context is joined to the transaction.
+* Lifecycle callbacks can invoke JNDI, JDBC, JMS, and enterprise beans.
+* A lifecycle callback method may modify the non-relationship state of
+  the entity on which it is invoked.
```
gavinking, line 2518: 👆

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: By the way, I would remove the footnote:

> The semantics of such operations may be standardized in a future release of this specification.

That's been there since JPA 1.0, and 17 years later: we've gone through several major versions, we have not standardized this, we have no plan to do so, and I think it would be _extremely_ hard to do so.

lukasj: I've missed that move, sorry. And I agree with the removal of that footnote as, in the end, it won't change the actual meaning anyway :-)

lukasj (approved):
LGTM

PR #480, opened 2023-08-21
Update changelog,

lukasj:
update build plugins,
fixed copyrights

...the most interesting part here is updated change log - I have added what is missing so far and reordered items a bit


gavinking (approved):
LGTM

Issue #481, opened 2023-08-21
Cleanup compile-time warnings introduced through recent additions


lukasj:
ie `PersistenceConfiguration` is missing a constructor - ie do we want to have just default public one and/or rather a specific one taking some arguments?

gavinking:
@lukasj please review #482 which I think addresses your concerns and more.

Issue #139, opened 2016-12-03
Provide constants for properties


lukasj:
There are a lot of available properties, each one applicable to specific scenarios:

*   Persistence Unit level properties (persistence.xml).
*   Entity Manager level properties (EntityManager#setProperty()).
*   Query hints.

Those properties are scattered trough the spec and it's difficult to find all of them and to know which one applies to which scenario.

A simple yet very handful addition would be a class that groups all the properties and specifies its applicability in its JavaDoc.

Hibernate already does something like this: [https://github.com/hibernate/hibernate-orm/blob/0a2a5c622e3eb30724e80bc8661c0ac55ebfb2be/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java](https://github.com/hibernate/hibernate-orm/blob/0a2a5c622e3eb30724e80bc8661c0ac55ebfb2be/hibernate-core/src/main/java/org/hibernate/cfg/AvailableSettings.java)
#### Affected Versions
[2.1.1]

gavinking:
This would be a very good thing to do.

gavinking:
#482 adds a list of configuration properties to `PersistenceConfiguration`.

This addresses at least the first requested item, which is, IMO, the most acute need.

In other work (#383) we've already added typesafe methods for specifying timeouts, cache modes, and the "load" graph.

The only standard hint not covered by all the above, unless I've missed something is the infamous `jakarta.persistence.fetchgraph`, which, if it were up to me, I would just deprecate.

PR #482, opened 2023-08-22
improvements to PersistenceConfiguration

gavinking:
see #481, #139


api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -76,27 +198,35 @@ public class PersistenceConfiguration {
     private ValidationMode validationMode = ValidationMode.AUTO;
     private PersistenceTransactionType transactionType = PersistenceTransactionType.RESOURCE_LOCAL;
 
-    private List<Class<?>> managedClasses = new ArrayList<>();
-    private List<String> mappingFileNames = new ArrayList<>();
-    private Map<String,Object> properties = new HashMap<>();
+    private final List<Class<?>> managedClasses = new ArrayList<>();
+    private final List<String> mappingFileNames = new ArrayList<>();
```
lukasj, line 202: It might be better to have this defined as `List<java.nio.file.Path>` instead of strings which in the end the provider needs to convert to something usable for getting a stream out of it. With `Path` we'd only need to resolve it against "some" root, if it is not absolute, and then use sth like:

```
try (InputStream is = Files.newInputStream(mappingFile) {
    parse(is);
}
```

to read its content. This leaves necessary _String_ validation on the caller - and that is IMO not a bad thing (...well, at least not for us :-) )

gavinking: Ummm OK so I guess the question is: are these files loaded using `getResource()`, or using a `File`?

gavinking: So I'm checking our APIs. We have:

```java
	/**
	 * Read mappings as an application resourceName (i.e. classpath lookup).
	 *
	 * @param name The resource name
	 *
	 * @return this (for method chaining purposes)
	 */
	public MetadataSources addResource(String name) {
		...
		return this;
	}

	/**
	 * Read mappings from a particular XML file.
	 * <p>
	 * The given path is resolved using {@link File#File(String)}.
	 *
	 * @param path The path to a file
	 *
	 * @return this (for method chaining purposes)
	 *
	 * @see #addFile(File)
	 */
	public MetadataSources addFile(String path) {
		addFile( new File( path ) );
		return this;
	}

	/**
	 * Read mappings from a particular XML file.
	 *
	 * @param file The reference to the XML file
	 *
	 * @return this (for method chaining purposes)
	 */
	public MetadataSources addFile(File file) {
		...
		return this;
	}
```

So we let you add a `File`, a `String` interpreted as a resource, or a `String` interpreted as a file path.

gavinking: Do we need two methods here, one for resources and one for files?

lukasj: > Ummm OK so I guess the question is: are these files loaded using `getResource()`, or using a `File`?

good point. The spec as well as the schema explicitly state that mapping-files are loaded _as resources_ which makes it obvious that _String_  interpreted as a resource name is the only variant to be  available (unless we allow alternatives also directly in the schema/spec), my bad. The only restriction we currently have in place is about the exact location of _orm.xml_

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -166,14 +301,17 @@ public PersistenceConfiguration managedClass(Class<?> managedClass) {
...
      * @return all configured managed classes
      */
     public List<Class<?>> managedClasses() {
         return managedClasses;
     }
 
     /**
-     * Add an {@code orm.xml} mapping file name to the configuration.
+     * Add an XML mapping file name to the configuration.
      * @param name the file path of the mapping file
```
lukasj: the javadoc here should be made consistent with the schema and the spec where we say that _An orm.xml mapping file or other mapping file is loaded as a resource by the persistence provider_ Using the word _file_ here gives an impression of handling arbitrary string based paths (even absolute ones)

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
lukasj: let's not add File/Path variants and keep that on the providers for now

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -183,6 +321,7 @@ public PersistenceConfiguration mappingFile(String name) {
     }
 
     /**
+     * The configured XML mapping files.
      * @return all configured mapping file names
```
lukasj: dtto `s/file/resource` for consistency with the spec/schema to avoid confusion about how the _String_ is to be treated

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
gavinking: I see, good point.

gavinking: Is that better, DYT?

gavinking: Um, see https://github.com/jakartaee/persistence/pull/482/commits/9e9ec882ede8fc98acaea6b0a1a306caa670b96c

lukasj (approved):
LGTM

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
lukasj: the meaning is clear and consistent with other places now, thanks!

Issue #483, opened 2023-08-22
javax.sql dependence


gavinking:
#433 provided a _very_ convenient API and addressed a major point of lack-of-portability for JPA usage in the SE environment.

But _literally ten minutes_ after merging the PR, @lukasj and I realized that there's something we need to consider more carefully here, that _possibly_ justifies rolling back the PR.

Essentially, JPA 3.1 has only one SPI which depends on the module `java.sql`, and in principle that dependence could be removed fairly straightforwardly. But #433 bakes in a much harder dependence to `java.sql` that in its present form _could never be removed_.

Now, that's not a _disaster_, since `java.sql` is part of the JDK after all, and almost every implementation of JPA has that dependence, and so indeed it's quite difficult to decide how much importance we should place on this.

But it's indeed true that the new APIs in #433 are some sort of layer-breaker, and it seems that there is still one implementation of JPA based on non-JDBC datasources, so arguably this new API pollutes `EntityManager` with operations that aren't guaranteed to be available.

So what should we do here?

- leave it alone because it's so _useful_?
- revert it because we want to keep JPA "clean" of JDBC?
- or, a third option...

My third option is to change the signature of those methods to:

```java
<C> void runWithConnection(ConnectionConsumer<C> action);
<C,T> T callWithConnection(ConnectionFunction<C,T> function)
```

where we've removed the direct dependence on JDBC, at the expense of a loss of type safety by making the connection type generic and provider-specific.

Usage with JDBC would now look like:

```java
em.runWithConnection((Connection conn) -> {
    conn.prepareCall("blah blah").execute();
});
```

which is not a disaster.

Thoughts?

gavinking:
PR #484 implements the idea proposed above. It's not terrible, IMO. WDYT?

PR #484, opened 2023-08-22
eliminate dependence on JDBC from runWithConnection()

gavinking:
makes the connection type generic

see #483

### DIFF
--- a/api/src/main/java/jakarta/persistence/ConnectionConsumer.java
+++ b/api/src/main/java/jakarta/persistence/ConnectionConsumer.java
@@ -16,11 +16,11 @@
 
 package jakarta.persistence;
 
-import java.sql.Connection;
-import java.sql.SQLException;
-
 /**
- * An executable action which makes use of a {@linkplain Connection JDBC connection}.
+ * An executable action which makes use of a native database connection.
+ * The connection is usually a JDBC connection.
+ *
+ * @param <C> the connection type, usually {@code java.sql.Connection}
  *
  * @see ConnectionFunction
  * @see EntityManager#runWithConnection(ConnectionConsumer)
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/persistence/ConnectionConsumer.java
```diff
@@ -16,25 +16,26 @@
...
 	 * Execute the action using the given connection.
 	 *
 	 * @param connection the connection to use
 	 *
-	 * @throws SQLException if a problem occurs calling JDBC
+	 * @throws Exception if a problem occurs calling the connection,
+	 *                   usually a {@code java.sql.SQLException}
 	 */
-	void accept(Connection connection) throws SQLException;
+	void accept(C connection) throws Exception;
```
lukasj, line 40: `<C, E extends Exception>` with `accept(C connection) throws E;` - not sure if that isn't too much.. The benefit here would be that the client has a possibility to explicitly cover particular exception type instead of having to deal with everything. WDYT?

gavinking: ummmmmm well I was assuming that the `EM` would just catch any checked exception and rethrow it as a `PersistenceException`

But also: won't `E` just get inferred to whatever type the lambda throws, which may as well be `Exception` *anyway*? It's essentially unconstrained...

lukasj: grr, I've used bad example on my end :-/

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1162,36 +1162,42 @@ public interface EntityManager extends AutoCloseable {
...
-     * with a transaction, the action is executed in the context of the transaction.
-     * The given action should close any JDBC resources it creates, but should not
-     * close the connection itself, nor commit or roll back the transaction. If the
-     * given action throws an exception, the persistence provider must mark the
+     * Execute the given action using the database connection underlying this
+     * {@code EntityManager}. Usually, the connection is a JDBC connection, but a
+     * provider might support some other native connection type, and is not required
+     * to support {@code java.sql.Connection}. If this {@code EntityManager} is
+     * associated with a transaction, the action is executed in the context of the
+     * transaction. The given action should close any JDBC resources it creates, but
```
lukasj: _any resources it creates_ ? likely obvious, but...

gavinking: Yeah of course, thanks, fixed now.

lukasj (approved):
LGTM

Issue #477, opened 2023-08-21
deprecate PersistenceUnitTransactionType


gavinking:
After #465 we have the API `PersistenceTransactionType` and the SPI `PersistenceUnitTransactionType`.

We can deprecate `PersistenceUnitTransactionType` for removal.

gavinking:
@lukasj Two questions:

1. should we add `EntityManagerFactory.getTransactionType()`?
2. should we rename this enum to the overly-verbose-but-more-precise `PersistenceTransactionManagementType`?

gavinking:
Or just `PersistenceTransactionManagement`?

hantsy:
`PersistenceTransactionType` is ok, but it seems txType attribute of `@Transactional()` is not a good naming, will confuse developers. The Spring `@Transactional` `propagation` is more reasonable.

Issue #56, opened 2013-05-04
@Convert annotation's converter property should be Class<? extends AttributeConverter>, not Class (unsafe)


lukasj:
Currently, the <tt>converter</tt> property for the <tt>@Convert</tt> annotation is declared as follows:

```
Class converter() default void.class;
```

However, along with just generally being unsafe ("safe" use would be <tt>Class<?></tt>), this does not properly restrict the set of classes that can be specified. My understanding is that this MUST be a class that implements <tt>javax.persistence.AttributeConverter</tt>. Therefore, the <tt>converter</tt> property should be specified like so:

```
Class<? extends AttributeConverter> converter() default void.class;
```

With this change, the developer will know at compile time if he has specified an incorrect class. Without this change, the developer will not know until he gets a runtime error, which is seriously less desirable.

gavinking:
Fixed in #487.

PR #487, opened 2023-08-23
remove raw types from APIs and fix typesafety of @Convert annotation

gavinking:
see #477 and #56


api/src/main/java/jakarta/persistence/Convert.java
```diff
@@ -193,7 +193,7 @@
    * element must be specified if multiple converters would
    * otherwise apply.
    */
-  Class converter() default void.class;
+  Class<? extends AttributeConverter> converter() default AttributeConverter.class;
```
gavinking, line 196: @lukasj Note this change.

lukasj (approved):
LGTM

PR #497, opened 2023-08-25
updated changelog

lukasj:
* reflect recent updates
* reordered changelog so the latest is the first and oldest is the last (my thinking is that I want to learn what's new first)
* turned JPQL into Jakarta Persistence QL
* fixed copyrights
* added `@SuppressWarnings` to `Query` where raw types are expected


lukasj:
also removed redundant modifiers where applicable and added .git-blame-ignore-revs

gavinking (approved):
Great move to get rid of all those redundant `public` modifiers!

PR #500, opened 2023-08-25
finish deprecation of spi.PersistenceUnitTransactionType

### LINKED ISSUE #477: deprecate PersistenceUnitTransactionType

lukasj:
@gavinking we've missed part in PersistenceUnitInfo; the name of the new method is long, but do you have better idea?

fixes #477

### DIFF
--- a/api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
+++ b/api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2021 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2023 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -11,6 +11,7 @@
  */
 
 // Contributors:
+//     Lukas Jungmann  - 3.2
 //     Linda DeMichiel - 2.1
 //     Linda DeMichiel - 2.0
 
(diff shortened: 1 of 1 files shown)


gavinking:
@lukasj Well, actually I didn't miss it, I deliberately left it out with the plan in mind that we would simply change the return type of the older method in JPA 4.0. That seemed like a better solution than introducing an ugly-named method that would hang around after the original was removed. WDYT?

lukasj:
I'm thinking about possible transition path and I see two ways it "can" be offered.

Let's assume that everything marked for removal will get removed and v4 contains many incompatible changes

- one change for which one cannot prepare beforehand more or less  doesn't matter, this one is only about changing import (ie `sed` can take of it)
vs.
- opt-in to move now and possibly don't care in the future, just delete ~3 lines when it comes

I'm inclining to think 2nd approach is better for most, even though I personally don't use wildcard imports and so I can use the `sed` way.

In any case, the intent should be clearly communicated - be it this or that. Do you prefer me to update the javadoc instead?

gavinking:
> the intent should be clearly communicated

Yes, that's fair.

> Do you prefer me to update the javadoc instead?

That's my (weak) preference, but _only_ because I didn't want to be left with the longer, uglier method name after the migration.

But honestly I don't care much. This is an SPI anyway, it's not in the face of the end user.

lukasj:
who could we check this with.. @scottmarlow do you have any preference here (from the AS vendor POV)?

lukasj:
add a comment instead

gavinking (approved):

Issue #443, opened 2023-08-10
improve formatting of Javadoc


gavinking:
There's a *lot* we could do to improve the Javadoc, but something easy to do which carries no risk of accidentally changing the semantics is to start using `@link` and `@code` tags consistently everywhere.

gavinking:
This was done a while ago. Closing.

PR #506, opened 2023-08-26
Major cleanup of the javadoc

gavinking:
see #443


api/src/main/java/jakarta/persistence/AssociationOverride.java
```diff
@@ -29,46 +29,44 @@
...
- * join columns.
+ * <p>If the relationship mapping is a foreign key mapping, the
+ * {@link #joinColumns} element is used. If the relationship mapping
+ * uses a join table, the {@link #joinTable} element must be specified
+ * to override the mapping of the join table and/or its join columns.
  *
+ * <p>Example 1: Overriding the mapping of a relationship defined by a
+ * mapped superclass
  * <pre>
- *    Example 1: Overriding the mapping of a relationship defined by a mapped superclass
```
lukasj, line 70: I was to say that this should have been updated to use `<pre>{@code ...` but then I realised #508 is a better way

gavinking: Yeah, we need @snippet. Actually I find that `<pre>{@code` doesn't work well.

gavinking:
OK, I think I'm 'bout done.

WDYT, @lukasj? Esp. the top-level javadoc I just added to `EM` and `EMF`.

api/src/main/java/jakarta/persistence/Converter.java
```diff
@@ -22,38 +22,36 @@
...
+ * annotation may be used to override or disable auto-apply conversion
+ * on a per-attribute basis.
  *
- * <p>If <code>autoApply</code> is <code>false</code>, only those
- * attributes of the target type for which the <code>Convert</code>
- * annotation (or corresponding XML element) has been specified will
- * be converted.
+ * <p>If {@code autoApply = false}, only those attributes of the target
+ * type for which the {@link Convert} annotation (or corresponding XML
+ * element) has been specified will be converted.
```
lukasj: few lines above, _will not be converted_ was changed to _are not converted_, but here, _will be converted_ remains unchanged - I do not have strong opinion on which tense should be used, the only (sort of) requirement I have is to use the chosen one consistently

gavinking: OK, sorted. I got rid of lots of instances of "will be", which was very overused throughout the javadoc.

gavinking:
> Esp. the top-level javadoc I just added to `EM` and `EMF`.

Also please carefully review what I wrote in `@Converter` and `@Convert`.

api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -21,21 +21,21 @@
...
- * <p> 
- * The syntax of the <code>columnList</code> element is a 
- * <code>column_list</code>, as follows:
+ * <p>The syntax of the {@code columnList} element is given by
+ * the following simple BNF:
  * 
  * <pre>
- *    column::= index_column [,index_column]*
- *    index_column::= column_name [ASC | DESC]
+ *    column_List ::= index_column [,index_column]*
```
lukasj: this should correspond to the definition in the spec, which currently has:
```
----
column::= index_column [,index_column]*
index_column::= column_name [ASC | DESC]
----
```

...so should be updated there in spec, in section 11.2.24

api/src/main/java/jakarta/persistence/Entity.java
```diff
@@ -23,9 +23,67 @@
...
+ *     mapped by the entity,
+ * <li>an {@linkplain ElementCollection element collection} usually
+ *     maps to a separate {@linkplain CollectionTable collection table},
+ * <li>a {@linkplain ManyToOne many-to-one} association usually maps
+ *     to a {@linkplain JoinColumn foreign key column} or columns in
+ *     one of the tables mapped by the entity,
+ * <li>a {@linkplain OneToOne one-to-one} association usually maps
+ *     to a unique foreign key relationship (sometimes using a shared
+ *     primary key),
+ * <li>a {@linkplain ManyToOne many-to-many} association usually maps
```
lukasj: OneToMany?

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -29,26 +29,117 @@
...
+ * The client may force an immediate flush to occur by calling
+ * {@link #flush()}.
+ *
+ * <p>At any given moment, a persistence context might hold an
+ * optimistic or pessimistic <em>lock</em> on an entity instance.
+ * The full range of possible lock types is enumerated by
+ * {@link LockModeType}. Some operations of this interface,
+ * including the methods {@link #lock(Object, LockModeType)},
+ * {@link #refresh(Object, LockModeType)}, and
+ * {@link #find(Class, Object, LockModeType)}, accept an explicit
```
lukasj, line 128: consider adding find, refresh taking FindOption and RefreshOption here

api/src/main/java/jakarta/persistence/Convert.java
```diff
@@ -25,188 +25,230 @@
...
+   * <li>if an {@linkplain Embedded embedded field} is annotated
+   *     {@code @Convert(attributeName = "startDate")}, then the
+   *     converter is applied to the field or property named
+   *     {@code startDate} of the referenced {@linkplain
+   *     Embeddable embeddable} class, or
+   * <li>if an {@linkplain ElementCollection map collection}
+   *     whose key type is an embeddable type is annotated
+   *     {@code @Convert(attributeName="key.jobType")}, the
+   *     converter is applied to the field or property named
+   *     {@code jobType} if the map key class.
```
lukasj: s/if/of ?

lukasj (commented):
In general looks good to me with only few minor things to fix. I'd also consider revisiting all usages of `{@code _}`, `{@code .}` and similar to bring in consistency wrt having this wrapped in quotation marks

lukasj (approved):
LGTM

Issue #509, opened 2023-08-27
Holes in the specification of attribute converters


gavinking:
AFAICT, nowhere does the spec say how to determine the target type of an `AttributeConverter`.

UPDATE: in fact, the spec as written is not even clear _at all_ as to what is a legal "target" type, and what is a legal converted type.

gavinking:
Something else I ran into: in the javadoc it's lightly-implied that `@Convert` with no `converter` and no `disableConversion` may be used to apply a non-auto-apply converter, which looks useful and totally makes sense.

But then this is contradicted (clarified?) in the spec which says that in fact the behavior in this case is undefined.

I would formalize what is implied by the Javadoc. Annotating a field `@Convert` should look for a suitable converter and apply it. There's no need to explicitly specify the converter class unless there is an ambiguity.

gavinking:
I've found another hole here. The spec mentions three cases for use of `@Convert`:

1. _"applied to a basic attribute or to an element collection of basic type"_
2. _"applied to an embedded attribute or to a map collection attribute whose key or value is of embeddable type"_
3. _"applied to a map containing instances of embeddable classes, the attributeName element must be specified, and "key." or "value." must be used"_
4. _"applied to a map to specify conversion of a map key of basic type, "key" must be used"_
5. _"Convert annotation may be applied to an entity class that extends a mapped superclass"_

What about:

1. Can `@Convert` be used with non-map collections of embeddables?
2. For a map value of basic type can/must "value" be used?

PR #510, opened 2023-08-27
nail down holes in the semantics of @Converter/@Convert

gavinking:
- Actually define what a converter target type is!
- Clarify what happens with a plain `@Convert` with no `converter` and no `disableConversion` ... it was vaguely implied by the javadoc that this applies a non-auto-apply converter, but this was contradicted by a line in the spec saying that the behavior is undefined.
- Make it clear that the "target type" may be any arbitrary Java type, but that the converted type must be a basic type.
- Properly define all cases involving element collections.

see #509


lukasj (approved):
LGTM

Issue #511, opened 2023-08-28
fallback id generator


gavinking:
It would be very convenient to be able to specify a "fallback" id generator to use when `@GeneratedValue` has no `generator` member, and the entity does not explicitly declare any `@TableGenerator` or `@SequenceGenerator`. This fallback could even be per-`GenerationType`.

One way would be via a property setting, so that, for example, you could set:

    jakarta.persistence.generator.AUTO = myDefaultSequenceGenerator
    jakarta.persistence.generator.SEQUENCE = myDefaultSequenceGenerator
    jakarta.persistence.generator.TABLE = myDefaultTableGenerator

But actually the most natural solution would be to let you declare a `@TableGenerator` or `@SequenceGenerator` with no `name` at the package level. And then that generator would be used as the fallback for the whole package.

That's a great solution, but I believe @lukasj mentioned he has some problems with us using package-level annotations.

lukasj:
those "problems" are mostly relate to the build tool(s) being used:
* maven 3.6.0-3.6.2 was somehow ignoring package-info files as such causing issues running xjc/jaxb etc (I wasn't directly involved in debugging this but noticed problems in few projects)
* maven compiler version 3.10.0 changed the way it treats package-infos (partially described in [MCOMPILER-205](https://issues.apache.org/jira/browse/MCOMPILER-205)) and that has impact on multi-release builds (package-info files are not _in certain configurations by default_ recompiled under the other/lower release version)

OTOH I do not think we should be feeling blocked by issues in particular build tools - there are other tools having no problems with this and since Java language has package-info we should use it if it makes sense to us

gavinking:
@lukasj OK, then in that case I would consider this one for 3.2, since it ties in with work we already did.

Actually, a package-level annotation would also be a _super_ nice solution to #409, IMO.

gavinking:
```java
the most natural solution would be to let you declare a `@TableGenerator` or `@SequenceGenerator` with no `name` at the package level. And then that generator would be used as the fallback for the whole package.
```

If we allowed `@TableGenerator` and `@SequenceGenerator` the package level, the semantics would be:

- if `name` is specified, the annotation defines a single (global) generator as usual, but
- if `name` is defaulted, the annotation is a template which defines a generator of each entity which needs one (and its name is the name of the entity, consistently to what we defined earlier).

gavinking:
[...]

PR #514, opened 2023-08-29
package-level @SequenceGenerator and @TableGenerator

gavinking:
These two commits:

1. fix an oversight in #406, where the name defaulting for `@GeneratedValue` was defined only in Javadoc but not in the spec, and
2. adds package-level `@SequenceGenerator` and `@TableGenerator` annotations, as proposed in #511.


lukasj (approved):
LGTM

PR #517, opened 2023-08-30
deprecate CriteriaQuery.multiselect and add overloads of array(), tuple()

gavinking:
There are two commits here:

1. adds overloads of `array()`, `tuple()` accepting `List` to `CriteriaBuilder`, and
2. deprecates `multiselect()` because it's not typesafe.


lukasj (approved):
LGTM

PR #520, opened 2023-08-31
3.2 draft01 3.2.0 b01

lukasj:
we may want to merge this before anything else to keep history right and fix the version number (don't know why it got bumped) or do not merge this at all.. Opinions?

Also note that the PR on the spec repo has not been merged yet and I have no idea when it is going to happen (if at all)


lukasj:
let's keep the branch and do not merge this

PR #521, opened 2023-08-31
Project web site

lukasj:
this builds the web site for this project as can be seen at https://lukasj.github.io/persistence/. The website gets updated after each push to master, so every change in sources or spec gets presented almost instantly.  I only think there should be a bit more general information; any idea what to add?


lukasj:
Yes, I think the intro needs a bit more than just project scope. I'm also inclining to remove help/issues sections, since they are already linked from the menu.

gavinking (approved):

Issue #519, opened 2023-08-31
Introduce specialized expression types for easier JPA Criteria querying


beikov:
I don't know how feasible this is, but this idea just came to mind and I wanted to share it here.

What if we introduced special `Attribute`, `Expression` and `Path` types for certain known types to make it easier to construct predicates and/or functions. I propose the following additions:

* `NumberAttribute`, `NumberExpression` and `NumberPath` when the type extends `Number`
* `StringAttribute`, `StringExpression` and `StringPath` when the type extends `String`
* `TemporalAttribute`, `TemporalExpression` and `TemporalPath` when the type extends `Temporal`
* `ComparableAttribute`, `ComparableExpression` and `ComparablePath` when the type extends `Comparable`

Then we add some overloads to `Path`:

```java
NumberPath<Y> get(NumberAttribute<? super X, Y> attribute);
...
```

And all of a sudden we could write nicer queries that don't involve calling methods on the `CriteriaBuilder` e.g.:

```java
cq.where( root.get(MyType_.price).plus(1).gt(10) )
```

instead of

```java
cq.where( cb.gt( cb.sum( root.get(MyType_.price), 1), 10) )
```

Of course this is just a simple example, but I think that the major criticism against the JPA Criteria API is that it is hard to write and read, which I think is mostly due to the need of involving the `CriteriaBuilder` everywhere.

I know that there is a small "type explosion" here i.e. 3 classes for every type we want to add decent support for, but I think it's worth it and that the number of supported types will stay in the single digit realm :)

gavinking:
That's a good idea @beikov.

How about prototyping your proposal as a pull req?

AleksNo:
Hello,

i like the idea. :)


Cheers

gavinking:
I think the time might be ripe to circle back and take another look at this one. We should compare what @beikov has proposed with the similar modeling that has been done in Jakarta Data and align the models where reasonable.

gavinking:
Done, finally!

PR #522, opened 2023-09-01
[#519] Introduce specialized expression types for easier JPA Criteria querying

beikov:
Fixes https://github.com/jakartaee/persistence/issues/519


api/src/main/java/jakarta/persistence/criteria/BooleanPath.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Represents a simple or compound attribute path from a
+ * bound type or collection, and is a "primitive" expression
+ * of a boolean type.
+ *
+ * @since 3.2
+ */
+@SuppressWarnings("hiding")
+public interface BooleanPath extends Path<Boolean>, BooleanExpression<Boolean> {
```
lukasj:
```
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/persistence/persistence/api/src/main/java/jakarta/persistence/criteria/BooleanPath.java:[27,70] type jakarta.persistence.criteria.BooleanExpression does not take parameters
[INFO] 1 error
```

gavinking:
Sorry, @beikov, finally found time to look at this. A couple of comments which are really questions:

1. It seems to me that we should either clearly maintain the distinction between `Predicate` and `BooleanExpression`, in which case logical operators don't apply to `BooleanExpression`s (or, if they do, they return `BooleanExpression` instead of `Predicate`) or simply drop the distinction. [Today, JPQL does not allow the use of local operators with boolean expressions.]
2. If we bother having `ComparableExpression` at all, then I would have expected `NumberExpression` to extend it. But perhaps this was just an oversight?
3. I expected to find overloads of the `get()` operation which e.g. accept `NumberAttribute` and return `NumberExpression`. Did I just miss something?
4. I think we need to weigh this carefully. This really does expand the API surface by _a lot_. It might be worth it, but let's think it through.

lukasj (commented):
The direction is good, it only needs to be completed

- make sure the code compiles
- it needs more detaild review

beikov:
>It seems to me that we should either clearly maintain the distinction between `Predicate` and `BooleanExpression`, in which case logical operators don't apply to `BooleanExpression`s (or, if they do, they return `BooleanExpression` instead of `Predicate`) or simply drop the distinction. [Today, JPQL does not allow the use of logical operators with boolean expressions.]

`Predicate` extends `BooleanExpression`, but I think I'd prefer dropping the distinction, though I have no idea what that means in terms of spec changes. I'd prefer to return `Predicate` from the new methods and also consider making `BooleanPath` extend `Predicate`. Let me know what you think about this.

>If we bother having `ComparableExpression` at all, then I would have expected `NumberExpression` to extend it. But perhaps this was just an oversight?

Unfortunately, `Number` does not extend `Comparable`, so it's not possible to unify the two. The `Comparable` stuff is mainly about being a base type for the temporal, boolean and string types. I suppose that for the same reason there are two method styles in `CriteriaBuilder` e.g. `gt` and `greaterThan`.

>I expected to find overloads of the `get()` operation which e.g. accept `NumberAttribute` and return `NumberExpression`. Did I just miss something?

That's all in `Path` see `<Y extends Number> NumberPath<Y> get(NumberSingularAttribute<? super X, Y> attribute);`

>I think we need to weigh this carefully. This really does expand the API surface by a lot. It might be worth it, but let's think it through.

It does add a lot of API surface, which is why I asked if you agree this is a good idea for making JPA Criteria more attractive. I wanted to highlight that this approach is similar to what QueryDSL does, so we might be able to learn from their design: https://github.com/querydsl/querydsl/tree/master/querydsl-core/src/main/java/com/querydsl/core/types/dsl

gavinking:
@beikov 

> `Predicate` extends `BooleanExpression`, but I think I'd prefer dropping the distinction

I think that makes sense. It might actually be a mistake that `Predicate` is an `Expression<Boolean>`, but since it already is, my prejudice is that we shouldn't need both `Predicate` _and_ `BooleanExpression`.

> Unfortunately, `Number` does not extend `Comparable`, so it's not possible to unify the two. 

I see, OK.

> The `Comparable` stuff is mainly about being a base type for the temporal, boolean and string types.

I'm not convinced that's worth it. I don't think I agree that `Boolean` should even be treated as comparable (no matter what Java thinks). And having `ComparableExpression` just to abstract over `TemporalExpression` and `StringExpression` doesn't seem worthwhile to me. I would vote to drop it.

> That's all in `Path`

OK 👍 

> It does add a lot of API surface, which is why I asked if you agree this is a good idea for making JPA Criteria more attractive.

I'm not against this change; actually I'm in favor of it. 

I just like to make the "cons" explicit.

beikov:
>I think that makes sense. It might actually be a mistake that Predicate is an Expression<Boolean>, but since it already is (ever since JPA 2), my prejudice is that we shouldn't need both Predicate and BooleanExpression.

So you want to drop `BooleanExpression` and return `Predicate` in the new methods instead? I'm fine with that. Final words from @lukasj?

>I'm not convinced that's worth it. I don't think I agree that Boolean should even be treated as comparable (no matter what Java thinks). And having ComparableExpression just to abstract over TemporalExpression and StringExpression doesn't seem worthwhile to me. I would vote to drop it.

There are a few types that JPA supports which are also `Comparable` for which I would consider the `ComparableExpression` etc. to be useful, e.g. `UUID`, `java.util.Date`, `java.util.Calendar`, `java.lang.Enum`. So in the end, the static JPA metamodel would generate `ComparableSingularAttribute` for attributes of these types.

>Well, actually, looking at that again, I think you still could make NumberExpression extend ComparableExpression. Just remove references to the type Comparable from ComparableExpression.

I would prefer not to do that, but if @lukasj thinks the same way then I'd be open for it.

gavinking:
> So you want to drop `BooleanExpression` and return `Predicate` in the new methods instead?

Yeah, I _think_ so, assuming @lukasj agrees.

> There are a few types that JPA supports which are also `Comparable `for which I would consider the `ComparableExpression` etc. to be useful, e.g. `UUID`, `java.util.Date`, `java.util.Calendar`, `java.lang.Enum`. 

OK, so:

1. `UUID` doesn't have a meaningful order, and
2. `java.util.Date`, `java.util.Calendar` are now officially deprecated in JPA, _but_
3. I agree that there are _some_ enum types which do have a well-defined order, even though I argued that they're the exception [here](https://docs.jboss.org/hibernate/orm/6.3/introduction/html_single/Hibernate_Introduction.html#enums).

So, sure, point taken, it would be nice to be able to compare enums.

So, fine, `ComparableExpression` survives.

> I would prefer not to do that?

Why? Numbers are in principle comparable. (I don't think we need to consider complex numbers here.)

And all the known subtypes of `Number`, including `Double`, `Long`, `Byte`, `BigInteger`, and `BigDecimal` all implement `Comparable`.

The only reason `Number` itself doesn't implement `Comparable` is because it's not a generic type.

Actually, the solution might be even simpler. Just add `Comparable<X>` as an upper bound on the type parameter of `NumberExpression`. Does that work?

api/src/main/java/jakarta/persistence/criteria/NumberExpression.java
```diff
@@ -0,0 +1,407 @@
...
+
+/**
+ * Type for number query expressions.
+ *
+ * @param <X> the type of the expression
+ *
+ * @since 3.2
+ */
+@SuppressWarnings("hiding")
+public interface NumberExpression<X extends Number> extends Expression<X> {
```
gavinking, line 30: @beikov How about changing this to:

    public interface NumberExpression<X extends Number&Comparable<X>> extends ComparableExpression<X> {

beikov:
>Actually, the solution might be even simpler. Just add Comparable<X> as an upper bound on the type parameter of NumberExpression. Does that work?
>@beikov How about changing this to:
>`public interface NumberExpression<X extends Number&Comparable<X>> extends ComparableExpression<X> {`

That could work, yes, but it brings up the question why `CriteriaBuilder` has two method flavors for the relational stuff for `Number` vs. `Comparable`. I can only assume it has to do with the way Javac does type inference when you have `<T extends Number> Expression<T> operation(T... expressions)` i.e. if you pass arguments of the same type, `T` will have that type and otherwise it will default to `Number`. But maybe @lukasj can shed some light on this.

api/src/main/java/jakarta/persistence/criteria/NumberExpression.java
```diff
@@ -0,0 +1,407 @@
...
+
+    /**
+     * Create a predicate for testing whether this expression is
+     * greater than the first argument.
+     * @param y  expression
+     * @return greater-than predicate
+     * @see CriteriaBuilder#gt(Expression, Expression)
+     */
+    Predicate gt(Expression<? extends X> y);
```
gavinking, line 39: By the way, I think the purpose of `gt()` as opposed to `greaterThan()` is that it's supposed to let you do inter-numeric-type comparisons, that is, compare a `Double` to an `Integer`, for example.

So this should be `Predicate gt(Expression<? extends Number> y);` I guess.

gavinking:
> That could work, yes, but it brings up the question why `CriteriaBuilder` has two method flavors for the relational stuff for `Number` vs. `Comparable`. 

Yeah that's what I just commented on the code, before I saw your reply:

> it's supposed to let you do inter-numeric-type comparisons, that is, compare a `Double` to an `Integer`, for example.

lukasj:
wrt `Predicate` vs `BooleanExpression` - should this be designed from scratch, my preference would be to deprecate everything related to `Predicate` and promote usage of `BooleanExpression` instead because 1) `Predicate` itself sort of conflicts with `j.u.function.Predicate` and 2) `BooleanExpression` is more consistent with the rest of Criteria API. In any case, I agree that we don't need both of them

api/src/main/java/jakarta/persistence/criteria/ComparableExpression.java
```diff
@@ -0,0 +1,138 @@
...
+
+/**
+ * Type for comparable query expressions.
+ *
+ * @param <X> the type of the expression
+ *
+ * @since 3.2
+ */
+@SuppressWarnings("hiding")
+public interface ComparableExpression<X extends Comparable<? super X>> extends Expression<X> {
```
gavinking, line 27: Should add `Order asc()` and `Order desc()` here, right? And possibly overloads accepting `Nulls`.

api/src/main/java/jakarta/persistence/criteria/StringExpression.java
```diff
@@ -0,0 +1,427 @@
...
+
+import java.util.List;
+
+/**
+ * Type for string query expressions.
+ *
+ * @since 3.2
+ */
+@SuppressWarnings("hiding")
+public interface StringExpression extends ComparableExpression<String> {
```
gavinking, line 27: Possibly add `Order ascIgnoreCase()` and `Order descIgnoreCase()` here.

api/src/main/java/jakarta/persistence/criteria/Expression.java
```diff
@@ -138,4 +139,126 @@ public interface Expression<T> extends Selection<T> {
...
+     */
+    Order asc(Nulls nullPrecedence);
+
+    /**
+     * Create an ordering by the descending value of this expression.
+     * @param nullPrecedence  the precedence of null values
+     * @return descending ordering corresponding to this expression
+     * @since 3.2
+     */
+    Order desc(Nulls nullPrecedence);
```
gavinking, line 263: Oh I see, they are declared here. But I wonder if it's really correct to have them here instead of on `ComparableExpression`. Yes, I know `CriteriaBuilder` allows them for any `Expression`, but maybe that's just a mistake?

I mean, if we're going to have `ComparableExpression` at all, it should be the interface defining things related to comparisons, it seems to me.

beikov: Hard to say. I can tell you that e.g. `byte[]` or `char[]` are not `Comparable`, but on the database side, the type is most probably `varbinary` or `varchar` respectively, which both are orderable. Also, think about how converted types might not be `Comparable`, but the conversion target type is.

I understand the desire to make things "typesafe", but I honestly doubt that moving `asc`/`desc` to `ComparableExpression` would help. On the contrary, I think it might cause more trouble than good.

gavinking: I mean, it seems to me that sorting binary data lexicographically is never meaningful. As for `char[]`, well, if the user had intended this array of characters to be interpreted as a string with a lexicographic ordering, then _surely_ they would have used `String` for that. So I think it's totally legit to say that these types can't be ordered. 

(Honestly, I don't even know why we even allow `char[]` except for "we allow it because we can".)

>  I honestly doubt that moving `asc`/`desc` to `ComparableExpression` would help.

If that's the case then I guess it's pretty unclear what purpose `ComparableExpression` even serves. If it's not abstracting over types which have a well-defined order, then it's not doing any work.

gavinking:
>  I don't think I agree that `Boolean` should even be treated as comparable (no matter what Java thinks).

Actually this statement, which I made a long time ago, without thinking too hard, is completely wrong.

We absolutely want to be able to include a `boolean` in an `order by`. Fortunately `Boolean` is `Comparable` in Java, so there's no issue there.

api/src/main/java/jakarta/persistence/criteria/Expression.java
beikov: `ComparableExpression` is abstracting over the relational predicates i.e. greater-than, smaller-than, etc.

I understand the desire to also move the ordering aspect into that, though like I was trying to say, I believe there are real use cases that might not work out this way. I'm fine with moving the methods if you insist and @lukasj agrees. I was just trying to point out why the existing `asc`/`desc` methods in `CriteriaBuilder` might be the way they are.

jwgmeligmeyling:
> I think that makes sense. It might actually be a mistake that Predicate is an Expression<Boolean>, but since it already is (ever since JPA 2), my prejudice is that we shouldn't need both Predicate and BooleanExpression.

If this were to be changed, does that mean that a boolean expression can be used where a predicate is expected in the CNF?

I.e.:

* `case when entity.someBoolean then 1 end`
* `where entity.someBoolean` instead of `where entity.someBoolean is true` 
* `where someBooleanReturningFunction()`

Because otherwise it seems desirable to distinguish between `Expression<Boolean>`/`BooleanExpression` and `Predicate`.

api/src/main/java/jakarta/persistence/criteria/Expression.java
```diff
@@ -138,4 +139,126 @@ public interface Expression<T> extends Selection<T> {
...
+
+    /**
+     * Create an expression that returns null if this and the argument
+     * evaluate to null, and the value of the first non-null expression
+     * otherwise.
+     * @param y value
+     * @return coalesce expression
+     * @since 3.2
+     */
+    Expression<T> coalesce(T y);
```
jwgmeligmeyling, line 163: Shouldn't there be a `Expression<T> coalesce(T... y)` as well?

api/src/main/java/jakarta/persistence/criteria/Expression.java
```diff
@@ -138,4 +139,126 @@ public interface Expression<T> extends Selection<T> {
...
+
+    /**
+     * Create an expression that returns null if this and the argument
+     * evaluate to null, and the value of the first non-null expression
+     * otherwise.
+     * @param y expression
+     * @return coalesce expression
+     * @since 3.2
+     */
+    Expression<T> coalesce(Expression<? extends T> y);
```
jwgmeligmeyling, line 153: Don't you want covariant return types for these on `ComparableExpression` and `NumberExpression`?

I.e. `entity.overrideValue.coalesce(parent.defaultValue).avg()` instead of `cb.asNumber(entity.overrideValue.coalesce(parent.defaultValue)).avg()`.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -902,6 +906,69 @@ public interface CriteriaBuilder {
...
+    <T extends Number> NumberExpression<T> literal(T value);
+
+    /**
+     * Create a boolean expression for a literal.
+     * @param value  value represented by the expression
+     * @return expression literal
+     * @throws IllegalArgumentException if value is null
+     * @since 3.2
+     */
+    BooleanExpression literal(Boolean value);
```
jwgmeligmeyling, line 925: Don't you need an `asBoolean(Expression<Boolean>)` as well? (Also for String, Number, Temporal, LocalDate, LocalDateTime). Becomes more important if you decide not to add the covariant overrides, so one can still do: `cb.asNumber(entity.overrideValue.coalesce(parent.defaultValue)).avg()`.

Alternatively those can be implemented as `asBoolean()` on `Expression`.

api/src/main/java/jakarta/persistence/criteria/ComparableExpression.java
```diff
@@ -0,0 +1,138 @@
...
+
+    //aggregate functions:
+
+    /**
+     * Create an aggregate expression for finding the greatest of
+     * the values (strings, dates, etc).
+     * @return greatest expression
+     * @see CriteriaBuilder#greatest(Expression)
+     */
+    ComparableExpression<X> max();
```
jwgmeligmeyling, line 129: Add `greatest` and `least` as well (non-aggegrate functions)? Or aren't these considered standard enough?

api/src/main/java/jakarta/persistence/criteria/Expression.java
jwgmeligmeyling, line 1: Does `in(Expression<?>... expression)` currently allow a `Subquery<T>` as parameter?

While making the Criteria API more fluent, shouldn't `notIn` alternatives be added to the API?

api/src/main/java/jakarta/persistence/criteria/NumberExpression.java
```diff
@@ -0,0 +1,407 @@
...
+
+    /**
+     * Create a predicate for testing whether this expression is
+     * between the first and second arguments in value.
+     * @param x  expression
+     * @param y  expression
+     * @return between predicate
+     * @see CriteriaBuilder#between(Expression, Expression, Expression)
+     */
+    Predicate between(Expression<? extends X> x, Expression<? extends X> y);
```
jwgmeligmeyling, line 112: Is it really important the types are the same here? Why not allow comparison between floats and integers such as `5.0 between 4 and 6`? Or compare an integer against a long: `4 < Long.MAX_VALUE`?

This is why in Querydsl many of the methods on ComparableExpressions are typed like this:

```java
<A extends Number & Comparable<?>> BooleanExpression gt(A right)
```

api/src/main/java/jakarta/persistence/criteria/Expression.java
gavinking: Interesting. In fact the `Criteria` API has never had a clean way to do this AFAICT. (You can I guess do it in an ugly way with `In`.) Anyway, good catch.

Solved in #936

api/src/main/java/jakarta/persistence/criteria/ComparableExpression.java
gavinking: They are not defined by JPQL.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
gavinking: Meh. Should not be necessary.

api/src/main/java/jakarta/persistence/criteria/Expression.java
gavinking: Yes good point I have added them in: https://github.com/jakartaee/persistence/pull/936

api/src/main/java/jakarta/persistence/criteria/Expression.java
gavinking: That would give you the dreaded varargs warning. Also, it's asymmetric in an ugly way. So nope.

api/src/main/java/jakarta/persistence/criteria/NumberExpression.java
gavinking: This is why `lt()`, `ge()` and friends are defined for any `Number` by `CriteriaBuilder`. I already pointed that out above.

Anyway it's sorted here: https://github.com/jakartaee/persistence/pull/936

beikov:
Superseded by https://github.com/jakartaee/persistence/pull/936

PR #533, opened 2023-09-20
use monospace font instead of italics for spec literals

gavinking:
I have attempted to replace all the italicized literals with monospaced literals.

This was trickier than I hoped and I'm sure I've made some mistakes.


lukasj (approved):

lukasj:
let's deal with possible mistakes when we spot them

gavinking:
Thanks 🙏

PR #536, opened 2023-10-19
#508: Update sample code in javadoc to use @snippet

### LINKED ISSUE #508: Update sample code in javadoc to use @snippet
[JEP 413](https://openjdk.org/jeps/413) included in SE 18+ introduced `@snippet` tag to simplify inclusion of example source code in API documentation - see ie [javamagazine](https://blogs.oracle.com/javamagazine/post/java-javadoc-snippet)

Given that this tag is specific to the version of javadoc tool being used, it should have no impact on the source/binary level of the code - and my quick test (on OpenJDK Runtime Environment (build 21+35-2513)) has proven that.

So if we want to use the `@snippet` tag now for 3.2, then the build has to require SE 18+ (21 is preferred since that is going to be LTS) and use that in combination with `release=17`. The only drawback I can see here is that consumers who want to rebuild our sources with javadoc will be required to use SE 18+ as well.

@ivargrimstad / @edburns - just to doublecheck - the EE 11 platform is going to rebuild all included component specs by SE 21, correct?

lukasj:
given we are required to use SE 21 for build, we can move to snippets right away while keeping using the `--release` switch

fixes #508

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -23,7 +23,7 @@ jobs:
 
     strategy:
       matrix:
-        java_version: [ 17, 21 ]
+        java_version: [ 21 ]
 
     steps:
     - name: Checkout for build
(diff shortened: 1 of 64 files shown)


gavinking:
Oh, excellent!

gavinking (approved):

PR #538, opened 2023-10-25
minor clarification to what is returned by getResultList()

gavinking:
see #537


Bukama:
I have no review rights, but as I brought this up in the discussion I want to say that I like this clarification.

dmatej (approved):

lukasj (approved):

PR #542, opened 2023-11-02
improvements to Graph operations for handling AttributeNodes

gavinking:
@sebersole has requested the attached changes to the `Graph` interface

@lukasj WDYT?

There are two related changes here:

- make `addAttributeNode()` return the `AttributeNode`
- add `getOrAddAttributeNode()` and `getAttributeNode()`

### DIFF
--- a/api/src/main/java/jakarta/persistence/Graph.java
+++ b/api/src/main/java/jakarta/persistence/Graph.java
@@ -32,9 +32,10 @@ import java.util.List;
 public interface Graph<T> {
 
     /**
-     * Add an attribute nodes to the entity graph.
+     * Add an attribute node to the entity graph.
      *
      * @param attributeName  name of the attribute
+     * @return the added attribute node
      * @throws IllegalArgumentException if the attribute is not an
      *         attribute of this entity.
      * @throws IllegalStateException if the EntityGraph has been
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/Graph.java
```diff
@@ -32,28 +32,85 @@
...
+     * @param attributeName  name of the attribute
+     * @return the attribute node
+     * @throws IllegalArgumentException if the attribute is not an
+     *         attribute of this entity.
+     * @throws IllegalStateException if the EntityGraph has been
+     *         statically defined
+     *
+     * @since 3.2
+     */
+    <Y> AttributeNode<Y> getOrAddAttributeNode(String attributeName);
```
lukasj: The `Graph` has methods for adding and removing nodes, one method allowing iteration through existing nodes but no method for checking presence of the node. What if there are methods like `hasAttributeNode(String)` instead of `getOrAdd...`?

gavinking: OK, which then which would you prefer:

- `findAttributeNode()` which returns `null` if there is no such node, or
- `hasAttributeNode()` which return `boolean`?

lukasj: `boolean` variant is better (and safer)

gavinking: Servido!

lukasj: Gracias

lukasj: btw do we need these `getOrAdd...` since they can be sort of defaults of `if (!hasAttributeNode(x)) addAttributeNode(x)` ?

gavinking: @sebersole requested them.

I guess I see them as useful because `addAttributeNode()` would throw if there's an existing node (and you might not know that) .... except, _actually we don't say that explicitly_!

I wonder, @sebersole, if it would be enough to just give `addAttributeNode()` the semantics currently assigned to `getOrAddAttributeNode()`, instead of adding a separate operation.

sebersole: > I wonder, @sebersole, if it would be enough to just give addAttributeNode() the semantics currently assigned to getOrAddAttributeNode(), instead of adding a separate operation.

I think that's a good idea to combat method explosion.

gavinking: OK then, let's take that approach.

lukasj (approved):

gavinking:
Thanks for merging, @lukasj.

Issue #52, opened 2013-03-20
return type of createNativeQuery(String, Class)


lukasj:
Hi,

I've got a minor suggestion:

    EntityManager.createNativeQuery(String, Class)

could return a TypedQuery as the result class is specified:

    TypedQuery<T> createNativeQuery(java.lang.String sqlString, java.lang.Class<T> resultClass)

-Adrian

m-reza-rahman:
I don't think this is possible without breaking backwards compatibility and this seems very minor. I suggest just closing it to reduce clutter.

Reza Rahman
Jakarta EE Ambassador, Author, Blogger, Speaker

Please note views expressed here are my own as an individual community member and do not reflect the views of my employer.

lprimak:
I support this for being included in 4 as well

PR #543, opened 2023-11-03
slight adjustment to signature of EntityManager.createNativeQuery()

gavinking:
preparing the way for #52 (which we won't do until JPA 4)

Change suggested by @lukasj in response to problem raised by @sebersole.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -1071,11 +1071,14 @@ public interface EntityManager extends AutoCloseable {
     /**
      * Create an instance of {@link Query} for executing a native
      * SQL query.
+     *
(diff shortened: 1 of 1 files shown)


lukasj (approved):

beikov:
Since this was merged and will be part of Jakarta Persistence 3.2, this means https://github.com/hibernate/hibernate-orm/pull/7494#discussion_r1379785538 is not a problem anymore, correct @gavinking? I didn't check yet if introducing a type variable is a backwards compatible change, but I'd assume so.

gavinking:
@beikov Right, that's the idea.

lukasj:
@beikov wrt compatibility - see [JLS 13.4.13](https://docs.oracle.com/javase/specs/jls/se17/html/jls-13.html#jls-13.4.13)

PR #552, opened 2023-12-09
a clarification to field-level @Access

gavinking:
and move footnotes into main text

### DIFF
--- a/spec/src/main/asciidoc/ch02-entities.adoc
+++ b/spec/src/main/asciidoc/ch02-entities.adoc
@@ -290,7 +290,7 @@ explicitly overridden via use of the `Access` annotation defined in
 ==== Default Access Type
 
 By default, a single access type (`FIELD` or `PROPERTY`) is inferred for
-an entity hierarchy. The default access type of an entity hierarchy is
+an entity hierarchy. The _default access type_ of an entity hierarchy is
 determined by the placement of mapping annotations on the attributes of
 the entity classes and mapped superclasses of the entity hierarchy which
 do not explicitly specify an access type.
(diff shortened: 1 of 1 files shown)


sebersole:
What do you think about clarifying "It is an error if a default access type cannot be determined"?  I think the general approach for all providers is to look at `@Id` / `@EmbeddedId`, so maybe codify that?

Although, the question could then become why is this:

```java
@Entity
class Book {
    @Id
    Integer id;
}
```

not functionally the same as

```java
@Entity
@Access(FIELD)
class Book {
    @Id
    Integer id;
}
```

for the purpose of something like

```java
@Entity
class Book {
    @Id
    Integer id;
    ...

    @Access(PROPERTY)
    String getIsbn() ...
}
```

The only distinction being implicit v. explicit in that case.

gavinking:
> It is an error if a default access type cannot be determined

For my part, I'm completely happy to add that. @lukasj WDYT?

> Although, the question could then become...

So there's two things you could say, I suppose. Either:

- `@Access` is not considered a "mapping annotation" for the purposes of that section, and does not affect determination of the default access type, in which case your code would be accepted, or
- `@Access` _is_ considered a "mapping annotation", and that code is an error.

Either option is self-consistent, and either option is fine by me.

sebersole:
> > It is an error if a default access type cannot be determined
> 
> For my part, I'm completely happy to add that. @lukasj WDYT?

That is already in the original and in your PR.  I was just saying we might consider adding the bit about `@Id` to explicitly note that that does in fact constitute a determination.

sebersole:
> So there's two things you could say, I suppose. Either:

I also don't much care which way we interpret that.  But having it explicit and consistent is a win imo

lukasj (approved):

gavinking:
> > So there's two things you could say, I suppose. Either:
> 
> I also don't much care which way we interpret that. But having it explicit and consistent is a win imo

Let's migrate this discussion to issue #558.

Issue #570, opened 2024-01-25
parametrized 'in' predicate lists


gavinking:
A very old and extremely widely-used feature of Hibernate that I had frankly just _forgotten_ is not part of JPA is the ability to parameterize in arguments of the `in` predicate, for example:

```java
s.createQuery("from Entity e where e.id in (:ids)")
        .setParameterList("ids", List.if(1, 2, 3))
        .getResultList(); 
```

The named parameter `:ids` gets expanded to multiple JDBC parameters `?, ?, ?` when the "parameter list" is bound.

I think this is very much low-hanging fruit, and that we should add something similar to JPA. But I'm not sure that the way it's done in Hibernate is the most elegant option. How about something like this instead:

```java
s.createQuery("from Entity e where e.id in (:id,...)")
        .setParameter("id", List.if(1, 2, 3))
        .getResultList(); 
```

Here it's the syntax of the query that identifies what's going on, instead of the method call.

Thoughts?

beikov:
Am I missing something in your request or is this already covered by [collection_valued_input_parameter](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a5107)?

gavinking:
OMG that's crazy. So the BNF refers to an undefined rule named `collection_valued_input_parameter`, but:

- nowhere does the spec say what the syntax or semantics of that rule are (apparently you're just supposed to _guess_ that it's a positional or named parameter), and
- nowhere does the spec mention how one can specify an argument to a `collection_valued_input_parameter`, nor even say what it is, nor even provide an example.

Apparently, in Hibernate we interpret this as you may pass a `List` (but not an array) to `setParameter()`.

But that's actually completely undefined, as far as I can tell.

gavinking:
So we should _at least_ clarify the spec to make clear that this is allowed:

```java
s.createQuery("from Entity e where e.id in :ids")
        .setParameter("ids", List.of(1, 2, 3))
        .getResultList(); 
```

beikov:
The spec does not directly refer to the BNF rule, but AFAICS it does specify this: https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#input-parameters

gavinking:
OK, but again, that doesn't _actually_ specify anything.

In particular, it doesn't say how you bind multiple arguments to a "collection-valued input parameter", nor even say what a "collection-valued input parameter" actually is.

gavinking:
[...]

PR #571, opened 2024-01-26
nail down semantics of collection-valued query parameters

gavinking:
see #570

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -6160,6 +6160,34 @@ _?1_ ”).
 
 The use of positional parameters is not supported for criteria queries.
 
+==== Arguments to query parameters
+
+Arguments are assigned to query parameters by calling `Query.setParameter()`.
+The first parameter of `setParameter()` identifies the named or positional
+parameter of the query.
+
+An argument may be assigned to a single-valued parameter of a JPQL or
+native SQL query by passing the argument to the second parameter of
+`setParameter()`.
+
+[source,java]
+----
+query.setParameter("name", name)
+----
+
+A list of arguments may be assigned to a collection-valued parameter
+of a JPQL query by packaging the arguments in a non-null instance of
+`java.util.List` and passing the list as an argument to the second
+parameter of `setParameter()`. The list should contain at least one
+element. If the list is empty the behavior is undefined. Portable
+applications should not pass an empty list to a collection-valued
+parameter.
+
+[source,java]
+----
+query.setParameter("names", List.of(name1, name2, name3))
+----
+
 ==== Named Queries
 
 Named queries are static queries expressed in
--- a/spec/src/main/asciidoc/ch04-query-language.adoc
+++ b/spec/src/main/asciidoc/ch04-query-language.adoc
@@ -1250,7 +1250,7 @@ part of a conditional expression except in an
 `empty_collection_comparison_expression`, in a
 `collection_member_expression`, or as an argument to the SIZE operator.
 
-==== Input Parameters
+==== Input Parameters [[a5065]]
 
 An input parameter allows a value in the Java program to be safely
 interpolated into the text of the parameterized query.
@@ -1269,9 +1269,10 @@ arithmetic operations involving the input parameter will result in an
 unknown value. See <<a5676>>.
 ====
 
-Every input parameter must be single-valued, unless the parameter occurs
-as the right hand side of an `IN` expressions (see <<a5107>>). Such input
-parameters may be collection-valued.
+An input parameter might be _single-valued_ or _collection-valued_.
+An input parameter which occurs directly to the right of the `IN` keyword
+in an `IN` predicate, as defined in <<a5107>>, is collection-valued. Every
+other input parameter is single-valued
 
 The API for the binding concrete arguments to query parameters is described
 in <<a3125>>.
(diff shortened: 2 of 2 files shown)


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -6122,6 +6122,30 @@ _?1_ ”).
...
+native SQL query by passing the argument to the second parameter of
+`setParameter()`.
+
+[source,java]
+----
+query.setParameter("name", name)
+----
+
+A list of arguments may be assigned to a collection-valued parameter of
+a JPQL query by passing the arguments in an instance of `java.util.List`
```
beikov: Shouldn't it rather be Collection?

gavinking: I dunno, should it?

I would have thought `List` was more natural here. It's a parameter list, after all.

beikov: It's named `collection-valued parameter` after all. Also, in Hibernate we support collection, not just list. I don't know why it is natural to expect the in-predicate parameter **collection** to be ordered and hence a `List`. I would rather argue it's more natural to expect a `Set`, because duplicate values don't make sense, but either way, let's stick to `Collection`.

gavinking: >  I don't know why it is natural to expect the in-predicate parameter collection to be ordered and hence a List. 

Well because it looks (syntactically) like a tuple constructor.

And so if we ever wanted to generalize this to other things were tuple constructors make sense, we would want order to be significant. e.g. `insert into Book values :properties` or `where (a,b,c) = :tuple`, etc.

I know we don't support these things in JPA _now_, so it's not a strong argument, but we actually _do_ have tuple constructors and `insert ... values` in Hibernate, so it's not crazy to think that we might add them to JPQL someday.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -6122,6 +6122,30 @@ _?1_ ”).
...
+to the second parameter of `setParameter()`. The list may be empty.
+
+[source,java]
+----
+query.setParameter("names", List.of(name1, name2, name3))
+----
```
lukasj, line 6189: What about _null_? That can be both - a single value as well as collection value.

spec/src/main/asciidoc/ch04-query-language.adoc
```diff
@@ -1494,8 +1495,14 @@ If the value of a
 `state_valued_path_expression` or `in_item` in an IN or NOT IN expression
 is `NULL` or unknown, the value of the expression is unknown.
```
lukasj: This should be probably updated to:
_If the value of a state_valued_path_expression, in_item, subquery or collection_valued_input_parameter in an IN or NOT IN expression..._
to cover all cases. WDYT?

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: Does:

> by passing the arguments in a **non-null** instance of `java.util.List`

fix the problem?

spec/src/main/asciidoc/ch04-query-language.adoc
gavinking: Can a subquery evaluate to `null`?

Also, I think I would say that a `collection_valued_input_parameter` can by _empty_, but never _null_.

spec/src/main/asciidoc/ch03-entity-operations.adoc
lukasj: yes, that should guarantee that _null_ is always treated as a _single_value_ and the lowest boundary for _collection_value_ is _empty_ (never _null_)

gavinking: Done.

spec/src/main/asciidoc/ch04-query-language.adoc
lukasj: > Can a subquery evaluate to `null`?

IN expression definitely can (ie if set of values contains NULL) - at least at the SQL level; not sure about subquery

> 
> Also, I think I would say that a `collection_valued_input_parameter` can by _empty_, but never _null_.

Does sentence above - _There must be at least one element in the
comma separated list that defines the set of values for the `IN`
expression._ - allow empty?

gavinking: Honestly the more I look at that sentence of the spec, the less I understand it.

Consider:

```
gavin=# select 1 where 2 in (2, 3, null);
 ?column? 
----------
        1
(1 row)
```

```
mysql> select 1 where 2 in (2, 3, null);
+---+
| 1 |
+---+
| 1 |
+---+
1 row in set (0.00 sec)
```

It's not the case that a `null` item in the `in` list causes the whole `in` expression to evaluate to null (at least not on postgres or mysql). So I think that's simply a mistake in the legacy spec.

lukasj: > It's not the case that a null item in the in list causes the whole in expression to evaluate to null (at least not on postgres or mysql). So I think that's simply a mistake in the legacy spec.

try [this example](https://stackoverflow.com/questions/4539340/sql-select-with-in-subquery-returns-no-records-if-the-sub-query-contains-null/4539354) to notice probably not too intuitive behaviour depending on the ~~availability~~ presence of NULL in the list of values.

wrt empty list:
```
mysql> select 1 where 1 in ();
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near ')' at line 1
```

gavinking: @lukasj I just pushed a commit, WDYT?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -6122,6 +6122,31 @@ _?1_ ”).
...
+
+[source,java]
+----
+query.setParameter("name", name)
+----
+
+A list of arguments may be assigned to a collection-valued parameter
+of a JPQL query by packaging the arguments in a non-null instance of
+`java.util.List` and passing the list as an argument to the second
+parameter of `setParameter()`. The list may be empty.
```
```suggestion
parameter of `setParameter()`. The list should contain at least one element. 
If the list is empty the behavior is undefined. Portable applications should not
pass an empty list to a collection-valued parameter.
```

lukasj (approved):

Issue #573, opened 2024-02-06
add 'EntityTransaction.setTimeout()'


gavinking:
I think it would be useful to have a `setTimeout()` method on the `EntityTransaction` interface. We could characterize it as a hint to make clear that vendors are permitted to ignore it.

PR #574, opened 2024-02-06
add EntityTransaction.setTimeout

gavinking:
see #573


api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -558,8 +558,9 @@ Query setParameter(int position, Date value,
     CacheStoreMode getCacheStoreMode();
 
     /**
-     * Set the query timeout.
-     * @param timeout the timeout, or null to indicate no timeout
+     * Set the query timeout, in milliseconds. This is a hint.
+     * @param timeout the timeout, in milliseconds, or null to
+     *                indicate no timeout
```
lukasj, line 565: same change is needed in TypedQuery and StoredProcedureQuery

lukasj (commented):
Is usage of `query.setTimeout(500)` an alternative to `query.setHint("jakarta.persistence.query.timeout", 500)` or is there any difference? - it may be worth adding the answer to the javadoc

api/src/main/java/jakarta/persistence/Query.java
gavinking: Done.

gavinking:
> Is usage of `query.setTimeout(500)` an alternative to `query.setHint("jakarta.persistence.query.timeout", 500)` or is there any difference?

It's the same.

> it may be worth adding the answer to the javadoc

Done.

lukasj (approved):

Issue #547, opened 2023-11-17
XSD defines an entity version as unbounded (List)


sebersole:
The XSD for `<attributes/>` currently defines -

```
      <xsd:element name="version" type="orm:version"
                   minOccurs="0" maxOccurs="unbounded"/>
```

which means that multiple `<version/>` elements are allowed.  Clearly that is wrong.

sebersole:
This is true for both 3.1 and 3.2.  I did not look back any further.

PR #582, opened 2024-02-28
fix maxOccurs for version element in orm.xml

gavinking:
see #547

cc @sebersole

### DIFF
--- a/api/src/main/resources/jakarta/persistence/orm_3_2.xsd
+++ b/api/src/main/resources/jakarta/persistence/orm_3_2.xsd
@@ -387,7 +387,7 @@
       <xsd:element name="basic" type="orm:basic"
                    minOccurs="0" maxOccurs="unbounded"/>
       <xsd:element name="version" type="orm:version"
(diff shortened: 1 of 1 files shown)


api/src/main/resources/jakarta/persistence/orm_3_2.xsd
```diff
@@ -387,7 +387,7 @@
       <xsd:element name="basic" type="orm:basic"
                    minOccurs="0" maxOccurs="unbounded"/>
       <xsd:element name="version" type="orm:version"
-                   minOccurs="0" maxOccurs="unbounded"/>
+                   minOccurs="0"/>
```
lukasj, line 390: add this change also to ch12, around line 1349, please

lukasj: I found one more thing to update (unrelated to this), so I'll fix this in the spec doc with that change

lukasj (approved):

PR #585, opened 2024-03-04
move some additional sections from 4.8 to 4.9, and two later sections *into* 4.8

gavinking:
This just continues the previous reorg - there are no changes to text.


gavinking:
I have pushed a second commit which moves the discussion of the static metamodel into the chapter on the metamodel. (A bit of a non-brainer.) This involved some _very_ light rewriting of preambles, but definitely no semantic changes.

gavinking:
And now I pushed a third commit which fixes an incorrect example.

lukasj (approved):

gavinking:
Thanks @lukasj

Issue #452, opened 2023-08-11
JPQL optional identification variable


gavinking:
I propose that we bless queries like this one:

    select title from Book where publicationDate > :date order by title, publicationDate

This is something that SQL lets you do, of course. 

Hibernate has tolerated stuff like this for quite a long time, and now supports it robustly. I would be surprised if other providers don't also allow it. (But I don't know if EclipseLink supports it.)

Of course, we would probably want to impose some restrictions on this. It's most useful when there's only one entity in the query, and therefore no risk of field name collision, so the "minimal" functionality would be to allow it only in that case.

This isn't a super-high priority, but it would be a good way to improve portability of queries between providers.

gavinking:
As a result of activity in Jakarta Data, this is now a much higher priority than I previously rated it.

@lukasj let's chat about this one soon.

gavinking:
Summary of current proposal:

1. When there is a `FROM` element and no `JOIN`s, the identification variable is optional and defaults to `this`.
2. The `SELECT` clause becomes optional, and defaults to `SELECT this`.
3. If the first element of a path expression is not an identification variable, it is interpreted as if it began with `this.`.

Thus, the query:

```
from Book where title like :title
```

Is equivalent to:

```
select this from Book as this where this.title like :title
```

This significantly streamlines the syntax of simple queries, without doing violence to the semantics of JPQL.

PR #588, opened 2024-03-11
make identification variables (and the SELECT) clause optional in JPQL

gavinking:
This is a minimal change to support the subset in Jakarta Data i.e. it is only allowed when there is just one thing in the `FROM` clause.

see #452


gavinking:
Interesting. I came up with an alternative approach that in certain ways is even more "minimal".

We could say that whenever the identification variable is missing (in a range declaration, or in a path expression) then it is implicitly assigned the variable `this`. 

So you could write:

```sql
select this from Entity where name like :name
```

And that would be equivalent to:

```sql
select this from Entity this where this.name like :name
```

I actually like this quite a lot. WDYT @lukasj?

gavinking:
> We could say that whenever the identification variable is missing (in a range declaration, or in a path expression) then it is implicitly assigned the variable this.

I pushed a second commit which shows what that would look like.

lukasj (approved):

lukasj:
it looks better with `this`, thanks!

gavinking:
Awesome! 🎉🎉🎉

PR #592, opened 2024-03-15
#589: avoid collisions in metamodel's variables

### LINKED ISSUE #589: Define behaviour for collisions in variable names in metamodel
right now in 3.2, there we have:

> For every persistent attribute y declared by class X, the metamodel class must contain a field declaration as follows:
> `public static final String Y = "y";`
> where the field name Y is obtained by transforming each lowercase character in the attribute name y to uppercase, and inserting > an underscore if the character following the transformed character is uppercase.

and later:

> For every persistent non-collection-valued attribute y declared by class X, where the type of y is Y, the metamodel class must contain a declaration as follows:
> public static volatile SingularAttribute<X, Y> y;

so if one has an entity with an attribute name in an uppercase - there might be valid usecases for this (an example I've noticed can be a variable called _IQ_),  then this leads to collisions. We should either change the newly added requirement for the fieldname to ie `public static final String Y_ = "y";` (add an underscore, or sth else, to the variable name) or say, that we do not handle collisions, or say that the latter is in lowercase (right now it is silent, so my expectation is that the variable name is taken as it is)

lukasj:
fixes #589

this should also handle the case where the name of a named thing is with a dot (ie `Employee.findAll`)

### DIFF
--- a/spec/src/main/asciidoc/ch05-metamodel-api.adoc
+++ b/spec/src/main/asciidoc/ch05-metamodel-api.adoc
@@ -71,7 +71,7 @@ metamodel class must contain a field declaration as follows:
 +
 [source,java]
 ----
-public static final String Y_ = "y";
+public static final String Y = "y";
 ----
 +
 where the field name `Y` is obtained by transforming each lowercase
@@ -165,13 +165,19 @@ where `n` is the name `"n"` with every character which is not a legal Java ident
 replaced with an underscore.
 
 Import statements must be included for the
-needed `jakarta.persistence.metamodel` types as appropriate (e.g.,
-`jakarta.persistence.metamodel.SingularAttribute`,
-`jakarta.persistence.metamodel.CollectionAttribute`,
-`jakarta.persistence.metamodel.SetAttribute`,
-`jakarta.persistence.metamodel.ListAttribute`,
-`jakarta.persistence.metamodel.MapAttribute`) and all classes `X`, `Y`,
-`Z`, `R`, and `K`.
+needed `jakarta.persistence` and `jakarta.persistence.metamodel` types as appropriate (e.g.,
+`jakarta.persistence.EntityGraph`, `jakarta.persistence.TypedQueryReference`,
+`jakarta.persistence.metamodel.SingularAttribute`, `jakarta.persistence.metamodel.CollectionAttribute`,
+`jakarta.persistence.metamodel.SetAttribute`, `jakarta.persistence.metamodel.ListAttribute`,
+`jakarta.persistence.metamodel.MapAttribute`, `jakarta.persistence.metamodel.EntityType`,
+`jakarta.persistence.metamodel.EmbeddableType`, `jakarta.persistence.metamodel.MappedSuperclassType`)
+and all classes `X`, `Y`, `Z`, `R`, and `K`.
+
+[NOTE]
+====
+Implementations of this specification are not required to resolve naming collisions
+when generating canonical metamodel classes.
+====
 
 [NOTE]
 ====
@@ -195,7 +201,6 @@ import jakarta.persistence.Entity;
 import jakarta.persistence.Id;
 import jakarta.persistence.ManyToOne;
 import jakarta.persistence.OneToMany;
-
 
 @Entity
 public class Order {
@@ -223,6 +228,7 @@ The corresponding canonical metamodel class, `Order_`, is as follows:
 package com.example;
 
 import java.math.BigDecimal;
+import jakarta.persistence.metamodel.EntityType;
 import jakarta.persistence.metamodel.SingularAttribute;
 import jakarta.persistence.metamodel.SetAttribute;
 import jakarta.persistence.metamodel.StaticMetamodel;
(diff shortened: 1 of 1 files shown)


gavinking:
@lukasj I'm not sure I like the idea of doing it for _every_ field. I would have done it only for the fields where a collision occurs.

lukasj:
While that is certainly feasible, I like the naming consistency this (unexpectedly) brings in wrt usage of the trailing underscore - it is the same for the name of the canonical class (`Cls_`) and the name of the field (`Cls_.ID_`); with static imports this also allows one to easily recognise generated constants from user defined ones.

gavinking:
mmmrmrrmrmrmrrrmmmm I can live with it I guess.

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -149,6 +150,8 @@ must contain a declaration as follows:
 public static volatile TypedQueryReference<R> _n_;
 ----
 +
+where `n` is the name `"n"` with not a legal Java identifier characters
+replaced with an underscore.
```
gavinking, line 153: @lukasj not sure what this sentence is trying to say.

gavinking: I think you mean "with every character which is not"

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -159,6 +162,8 @@ a declaration as follows:
 public static volatile EntityGraph<X> _n;
 ----
 +
+where `n` is the name `"n"` with not a legal Java identifier characters
+replaced with an underscore.
```
gavinking, line 166: ditto

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -149,6 +150,8 @@ must contain a declaration as follows:
 public static volatile TypedQueryReference<R> _n_;
 ----
 +
+where `n` is the name `"n"` with not a legal Java identifier characters
```
gavinking:
```suggestion
where `n` is the name `"n"` with every character which is not a legal Java identifier character
```

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -159,6 +162,8 @@ a declaration as follows:
 public static volatile EntityGraph<X> _n;
 ----
 +
+where `n` is the name `"n"` with not a legal Java identifier characters
```
gavinking:
```suggestion
where `n` is the name `"n"` with every character which is not a legal Java identifier character
```

spec/src/main/asciidoc/ch05-metamodel-api.adoc
lukasj: yes, that was the intent, thanks for the better language

gavinking (approved):

gavinking:
Hold on a sec.

@lukasj, if we're adding a trailing `_`, then in principle we no longer need the upper-capitalization. We could let people write:

```java
@OneToMany(mappedBy=Book_.publisher_)
Set<Book> books;
```

instead of:

```java
@OneToMany(mappedBy=Book_.PUBLISHER_)
Set<Book> books;
```

WDYT?

lukasj:
well, it's `public static final` constant for which the java (naming) convention is to use upper-case

gavinking:
> well, it's `public static final` constant for which the java (naming) convention is to use upper-case

I know, but I'm pretty sure we don't need to go to court to break that convention if it's the right thing to do.

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -72,13 +72,14 @@ metamodel class must contain a field declaration as follows:
 +
 [source,java]
 ----
-public static final String Y = "y";
+public static final String Y_ = "y";
```
lukasj: remove underscore, add disclaimer about non-portable behaviour in case of collisions

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -158,15 +161,23 @@ a declaration as follows:
...
-`jakarta.persistence.metamodel.ListAttribute`,
-`jakarta.persistence.metamodel.MapAttribute`) and all classes `X`, `Y`,
-`Z`, `R`, and `K`.
+needed `jakarta.persistence` and `jakarta.persistence.metamodel` types as appropriate (e.g.,
+`jakarta.persistence.EntityGraph`, `jakarta.persistence.TypedQueryReference`,
+`jakarta.persistence.metamodel.SingularAttribute`, `jakarta.persistence.metamodel.CollectionAttribute`,
+`jakarta.persistence.metamodel.SetAttribute`, `jakarta.persistence.metamodel.ListAttribute`,
+`jakarta.persistence.metamodel.MapAttribute`, `jakarta.persistence.metamodel.EntityType`,
+`jakarta.persistence.metamodel.EmbeddableType`, `jakarta.persistence.metamodel.MappedSuperclassType`)
+and all classes `X`, `Y`, `Z`, `R`, and `K`.
```
gavinking, line 169: I wonder why we even need this paragraph. I think I would just delete it.

spec/src/main/asciidoc/ch05-metamodel-api.adoc
```diff
@@ -158,15 +161,23 @@ a declaration as follows:
...
+
+[NOTE]
+====
+Implementations of this specification are not required to resolve naming collisions
+when generating canonical metamodel classes.
```
gavinking:
```suggestion
resulting from the rules above when generating canonical metamodel classes.
```

gavinking (approved):
Couple of suggestions you can accept or ignore as you please.

spec/src/main/asciidoc/ch05-metamodel-api.adoc
lukasj: I've made this shorter

Issue #594, opened 2024-03-16
'id()' function


gavinking:
For a very long time, the query language in Hibernate has allowed you to refer to the identifier of an entity via the syntax `id(foo)`, and the version using `version(foo)`. It was never clear that this was especially useful for anything, so I've never proposed it here.

However, it turns out that `id()` is _really_ useful for defining generic repository interfaces in Jakarta Data, and we have really been going in circles and inventing quite nasty workarounds due to its absence in JPQL. And I think that `version()` will be useful for similar reasons.

I propose that we add the `id()` and `version()` functions to JPQL.

PR #596, opened 2024-03-16
introduce id() and version() functions

gavinking:
- this is the most minimal form of this which supports what we want to do in Jakarta Data
- for now, only allow them in comparison expressions with input parameters
- also make the section on Entity Type Expressions more readable

@see #594


gavinking:
@lukasj The reason I'm proposing this _now_, even though it seems like something that's otherwise highly non-urgent at best, is that:

- We're about to write down something a bit inelegant in the Jakarta Data spec, and, while it's bearable, I would prefer not to have it there frozen in stone until Jakarta is exterminated in the AI apocalypse.
- @njr-11 is doing an implementation of Jakarta Data on EclipseLink, and I believe that this would make his life at least _slightly_ easier.

It's not critical, but it would be nice.

Issue #374, opened 2022-07-04
stateless entity manager


gavinking:
A feature that Hibernate has had for a very long time, but which has never made it into the JPA spec is `StatelessSession`.

<https://docs.jboss.org/hibernate/orm/6.1/javadocs/org/hibernate/StatelessSession.html>

The idea of a stateless session is to allow the program to directly perform persistence operations without going via a persistence context, much like what you would get with handcoded SQL:

- For some users, this is a slightly more intuitive way to work, but 
- the primary motivation is that it relieves the program of the need to explicitly `clear()` the persistence context when performing operations which affect a large number of entity instances.

Now, the flipside is that the semantics differ in significant ways to the semantics of a regular `EntityManager`:

- Naturally, there's no first-level cache, but also
- the second-level cache is bypassed, since for the most common usecase we don't want to go filing the cache up with data we're probably not going to use again soon.
- There's no transactional write-behind, nor automatic dirty checking, since these behaviors depend on having a persistence context.
- On the downside, there's no way to update a collection (except for an unowned `@OneToMany`), since it's hard to come up with an efficient way to implement that without a persistence context for the common usecase.
- Lazy loading is an [explicit operation](https://docs.jboss.org/hibernate/orm/6.1/javadocs/org/hibernate/StatelessSession.html#fetch(java.lang.Object)).
- Finally, and most significantly: with no dirty checking, updating an entity is also always an [explicit operation](https://docs.jboss.org/hibernate/orm/6.1/javadocs/org/hibernate/StatelessSession.html#update(java.lang.Object)).

To emphasize these semantic differences, the persistence operations are given distinct names:

- `insert()` (not `persist()`)
- `delete()` (not `remove()`)
- `update()`
- `get()` (not `find()`)

(Though `refresh()` is still called `refresh()`.)

There are two additional (minor) limitations, though they're not fundamental to the idea, and could in principle be removed: there's no cascading of persistence operations, and there's no persistence callback events. These features don't seem terribly useful for stateless sessions but nor would they be difficult to implement.

I think that it's well-worth discussing whether a stateless entity manager API should be introduced in the JPA spec.

rbygrave:
[...]

PR #601, opened 2024-03-21
Improve Javadoc about OptimisticLockException

quaff:


api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -246,7 +248,9 @@ public interface EntityManager extends AutoCloseable {
      * association marked {@link CascadeType#REMOVE cascade=REMOVE}. If the
      * given entity instance is already removed, it is ignored. If the
      * given entity is new, it is itself ignored, but the operation still
-     * cascades.
+     * cascades. It will cause later {@link #flush()} throws
+     * {@link OptimisticLockException} if entity is updated or deleted by
+     * another transaction.
```
gavinking: Note sure what this is trying to say.

gavinking:
So I believe&mdash;but I'm not sure, without digging more deeply, that we actually permit the provider to throw `OptimisticLockException` synchronously from `persist()` or `remove()`. Usually that happens from `flush()` but I think the provider has freedom here.

quaff:
> So I believe—but I'm not sure, without digging more deeply—that we actually permit the provider to throw `OptimisticLockException` synchronously from `persist()` or `remove()`. Usually that happens from `flush()` but I think the provider has freedom here.

Unlike [stateless entity manager](https://github.com/jakartaee/persistence/issues/374), for stateful entity manager, there is no database interaction until flush, `remove()` doesn't know the entity is stale.

gavinking:
> for stateful entity manager, there is no database interaction until flush, remove() doesn't know the entity is stale.

That is one possible implementation, sure. But what the _spec actually says_ is stuff like:

> The entity X will be entered into the database at or before transaction commit or as a result of the flush operation.

>  A removed entity X will be removed from the database at or before transaction commit or as a result of the flush operation.

> The persistence provider runtime is permitted to perform synchronization to the database at other times as well when a transaction is active and the persistence context is joined to the transaction. 

And of course, it's definitely the case that synchronization can happen when `Query.getResultList()` is called.

So optimistic lock failures can in principle happen at lots of other moments, not only in the ones you've listed here.

api/src/main/java/jakarta/persistence/EntityManager.java
quaff: I changed it to 
```java
     * <p>Note that {@link OptimisticLockException} may not be thrown by this
     * method but defer to database synchronization operation.
```

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -236,6 +236,8 @@ public interface EntityManager extends AutoCloseable {
      * @throws TransactionRequiredException if there is no transaction
      *         when invoked on a container-managed entity manager of
      *         that is of type {@link PersistenceContextType#TRANSACTION}
+     * @throws OptimisticLockException when optimistic locking conflict
+     *         occurs
```
gavinking:
```suggestion
     * @throws OptimisticLockException if an optimistic lock failure
     *         occurs
```

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -247,13 +249,17 @@ public interface EntityManager extends AutoCloseable {
...
+     * method but defer to database synchronization operation.
      * @param entity  a managed, new, or removed entity instance
      * @throws IllegalArgumentException if the instance is not an entity
      *         or is a detached entity
      * @throws TransactionRequiredException if invoked on a
      *         container-managed entity manager of type
      *         {@link PersistenceContextType#TRANSACTION} and there is
      *         no transaction
+     * @throws OptimisticLockException when optimistic locking conflict
+     *         occurs
```
gavinking:
```suggestion
     * @throws OptimisticLockException if an optimistic locking conflict
     *         occurs
```

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -564,6 +570,8 @@ <T> T find(EntityGraph<T> entityGraph, Object primaryKey,
      * @throws TransactionRequiredException if there is
      *        no transaction or if the entity manager has not been
      *        joined to the current transaction
+     * @throws OptimisticLockException when optimistic locking conflict
+     *         occurs
```
gavinking:
```suggestion
     * @throws OptimisticLockException if an optimistic locking conflict
     *         occurs
```

gavinking:
Thanks for this @quaff. I have used this work as the basis for #772, which is somewhat more complete.

PR #608, opened 2024-03-28
update spec license to 1.1

lukasj:


api/src/main/javadoc/doc-files/speclicense.html
```diff
@@ -14,14 +14,13 @@
 
 -->
 
-<title>Eclipse Foundation Specification License - v1.0</title>
+<title>Eclipse Foundation Specification License - v1.1</title>
 </head>
 <body>
 <h1>Eclipse Foundation Specification License - v1.0</h1>
```
gavinking: Why does it still say 1.0 here?

lukasj: because I'm bad in copy&paste programming :-D

gavinking (approved):

PR #609, opened 2024-03-31
consistently use monospace font in ch4

gavinking:
this has been bugging me for a long time


gavinking:
@lukasj there should be no changes to words here, only the addition of backticks around words that deserve them.

lukasj (approved):

PR #624, opened 2024-05-12
errata: @EntityResult.lockMode should default to NONE

gavinking:
Since `OPTIMISTIC` has a slightly-weird definition in JPA.

@lukasj I messed up, is it too late to fix this?


lukasj:
> @lukasj I messed up, is it too late to fix this?

yes, the ballot is already ongoing...

lukasj (approved):

gavinking:
Thanks 🙏

PR #631, opened 2024-05-21
3.2/3.2.0 release branch

lukasj:
bumps the version number after the release


gavinking:
Why 3.3 of spec and 3.2.1 of API?

lukasj:
that's what the release job does - minor increments only - but whatever having comparatively higher version number would work too. Note that this is purely technical thing and we can tweak the version the way we want through a different commit(s) later

gavinking (approved):

PR #638, opened 2024-06-15
allow @JoinColumns(value) to be empty and defaulted

gavinking:
this allows foreignKey to be specified independently for multi-column joins

also improve the javadoc for @JoinColumn and @JoinColumns

(I will open an issue for this.)


gavinking:
Thanks, @lukasj !

Issue #600, opened 2024-03-19
Standardize the initial value of integer version


quaff:
Test rely on version value assertion is failed when switch from Hibernate to EclipseLink, because Hibernate starts from `0` but EclipseLink starts from `1`, It would be nice if the spec enforce it. 
I'd like it starts from `1` because it's human readable then can be display in UI directly.

gavinking:
The most obvious way to do this is to let you write:

```java
@Version
int version = 1;
```

and I have a feeling Hibernate already lets you do that.

quaff:
> ```java
> @Version
> int version = 1;
> ```

Both of Hibernate and EclipseLink works fine, perhaps we should repurpose the issue to make this feature official, any non-negative integer initial value will be the seed of version, null initial value is considered as undefined and vendor is free to handle it.

gavinking:
> Both of Hibernate and EclipseLink works fine, perhaps we should repurpose the issue to make this feature official

That sounds good to me.

> any non-negative integer initial value will be the seed of version, negative or null initial value is considered as undefined and vendor is free to handle it.

Yes.

quaff:
EclipseLink treats `int version = 0` as `Integer version = null`, so the word `non-negative` should be revised to `positive`.

quaff:
Superseded by https://github.com/jakartaee/persistence/pull/649.

quaff:
@gavinking timestamp version applied here too, think about
```java
@Version
LocalDateTime version = LocalDateTime.of(...);
```
Which one should be the initial value, `LocalDateTime.of(...)` or `LocalDateTime.now()` ?

gavinking:
> timestamp version applies here too

Yes, sure, I know.

PR #649, opened 2024-08-29
Clarify the seed value of version

quaff:
Supersede https://github.com/jakartaee/persistence/issues/600

### DIFF
--- a/api/src/main/java/jakarta/persistence/Version.java
+++ b/api/src/main/java/jakarta/persistence/Version.java
@@ -41,12 +41,24 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * {@code long}, {@link Long}, {@code java.sql.Timestamp},
  * {@link java.time.Instant}, {@link java.time.LocalDateTime}.
  *
- * <p>This field declares a version number:
+ * <p>Any <strong>positive number</strong> or <strong>non-null timestamp</strong>
+ * initial value will be the seed of version, otherwise vendors are free to
+ * seed the version.
+ *
+ * <p>This field declares a version number, the version seed is vendor-specific:
  *
  * {@snippet :
  * @Version
  * @Column(name = "REVISION")
  * protected int version;
+ * }
+ *
+ * <p>This field declares a version number, the version seed is <code>1</code> across vendors:
+ *
+ * {@snippet :
+ * @Version
+ * @Column(name = "REVISION")
+ * protected int version = 1;
  * }
  *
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/persistence/Version.java
```diff
@@ -41,14 +41,26 @@
  * {@code long}, {@link Long}, {@code java.sql.Timestamp},
  * {@link java.time.Instant}, {@link java.time.LocalDateTime}.
  *
- * <p>This field declares a version number:
+ * <p>Any <strong>positive number</strong> or <strong>non-null timestamp</strong>
+ * initial value will be the seed of version, otherwise vendors are free to
+ * seed the version.
```
gavinking, line 46: I don't see how this actually addresses the real actually-known use case here which is to enable a portable way to choose between an initial version of `0` and an initial version of `1`.

With this language:

    @Version int version = 1;

lets me portably get an initial value of `1`. OK, great.

But:

    @Version int version = 0;

Just does something non-portable and vendor-specific.

So there's no portable way to request an initial value of `0`.

Is that correct?

quaff: > So there's no portable way to request an initial value of 0.

Yes, there is no way to differ `int version = 0;` from `int version;`, both of them should be treated as non-predefined, if negative numbers are treated that way, why not zero?

gavinking: > if negative numbers are treated that way, why not zero?

Well, because counting from `0` is perfectly legit, and often used in computing.

gavinking: Indeed, _it's what Hibernate does by default_, so it can't be terrible.

quaff: > > if negative numbers are treated that way, why not zero?
> 
> Well, because counting from `0` is perfectly legit, and often used in computing.

I'm not against that if EclipseLink team would like to support counting from `0`.

gavinking: Exactly, so it's something that needs to be discussed properly with @lukasj and we need to think through the full impact. i.e. we need an issue to track this. We can't just make this sort of change as a PR with no discussion.

quaff: > we need an issue to track this. 

Could you reopen https://github.com/jakartaee/persistence/issues/600?

PR #654, opened 2024-09-20
three minor improvements

gavinking:
No semantic changes here.

### DIFF
--- a/spec/src/main/asciidoc/ch02-entities.adoc
+++ b/spec/src/main/asciidoc/ch02-entities.adoc
@@ -2308,29 +2308,31 @@ public class ShoppingCart extends Cart {
 
 The mapping of class hierarchies is specified through metadata.
 
-There are three basic strategies that are
-used when mapping a class or class hierarchy to a relational database:
+There are three basic strategies which may be used when mapping an
+entity class hierarchy to tables in a relational database.
 
-* a single table per class hierarchy
-* a joined subclass strategy, in which fields
-that are specific to a subclass are mapped to a separate table than the
-fields that are common to the parent class, and a join is performed to
-instantiate the subclass.
-* a table per concrete entity class
+* With a _single table per class hierarchy_, every field declared by
+  any class in the hierarchy maps to a column of the same table.
+  A subclass shares the table of its superclasses.
+* With a _table per concrete entity class_, every field declared or
+  inherited by a given subclass is mapped to the same table, but each
+  concrete class in the hierarchy has its own separate table.
+* For the _joined subclass_ strategy, any field which is declared by
+  a subclass is mapped to a separate table from the fields which are
+  declared by its superclass. Table joins may be used to retrieve all
+  fields declared and inherited by the subclass.
 
-An implementation is required to support the
-single table per class hierarchy inheritance mapping strategy and the
-joined subclass strategy.
+An implementation is required to support the single table per class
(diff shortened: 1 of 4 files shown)


spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2308,29 +2308,31 @@ public class ShoppingCart extends Cart {
...
+* With a _table per concrete entity class_, every field declared or
+  inherited by a given subclass is mapped to the same table, but each
+  concrete class in the hierarchy has its own separate table.
+* For the _joined subclass_ strategy, any field which is declared by
+  a subclass is mapped to a separate table from the fields which are
+  declared by its superclass. Table joins may be used to retrieve all
+  fields declared and inherited by the subclass.
+
+An implementation is required to support the single table per class
+hierarchy inheritance mapping strategy and the joined subclass strategy.
```
scottmarlow, line 2326: The old removed wording `single table per class hierarchy inheritance mapping strategy` is shown as required but would it be easier on the readers eyes to use the new wording?  

Perhaps:
> An implementation is required to support the `single table per class
hierarchy` and the `joined subclass strategy`.

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2308,29 +2308,31 @@ public class ShoppingCart extends Cart {
...
 
 [NOTE]
 ====
-Support for the table per concrete class
-inheritance mapping strategy is optional in this release. Applications
-that use this mapping strategy will not be portable.
+Support for the table per concrete class inheritance mapping strategy
+is optional in this release. Applications that use this mapping strategy
+are not portable between all providers.
```
scottmarlow, line 2332: Previously was `table per concrete class` required and now is optional?  Or was it optional before?

Note that the TCK for the Jakarta EE Platform specification cannot have any optional tests.  From the perspective of the Persistence component TCK, I think we should only add tests for required features.

gavinking: It has always been optional. This edit just changes:

> Support for the table per concrete class inheritance mapping strategy is optional in this release. Applications that use this mapping strategy will not be portable.

To:

> Support for the table per concrete class inheritance mapping strategy is optional in this release. Applications that use this mapping strategy are not portable between all providers.

There's no semantic change.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: @scottmarlow The italicized text indicates the names of the three straggles, namely:

1. "single table per class hierarchy"
2. "table per concrete entity class"
3. "joined subclass"

You propose:

> An implementation is required to support the single table per class hierarchy ...

But that doesn't make sense. The single table per class hierarchy what? That's not a well-formed noun phrase.

gavinking:
@scottmarlow are you happy with this? I would like to merge it. Thanks.

spec/src/main/asciidoc/ch02-entities.adoc
scottmarlow: That is clearer now,  I was ignoring the `italicized`

gavinking: OK, good.

scottmarlow:
Thanks Gavin!

gavinking:
Thank you @scottmarlow.

Issue #614, opened 2024-04-20
jakarta.annotation.Nonnull should imply optional=false


gavinking:
We should allow:

```java
@Nonnull
String name;
```

as an alternative to:

```java
@Basic(optional = false)
String name;
```

and similarly for `@ManyToOne` and `@OneToOne`.

gavinking:
Done.

PR #656, opened 2024-09-20
jakarta.annotation.Nonnull implies optional=false

gavinking:
see issue #614


njr-11 (approved):

gavinking:
Thanks, @njr-11. This was proposed back in April last year, and I've not heard a single object to it. Indeed, it's pretty unobjectionable. So merging.

Issue #622, opened 2024-04-29
Introduce `Query::getResultCount`


quaff:
Hibernate introduce `SelectionQuery.getResultCount()` since 6.5.0, It's useful for paged query, `Query` object can be reused for both data and count query.
```java
Query query = entityManager.createQuery(queryString);
query.setFirstResult(10);
query.setMaxResults(10);
result.setTotal(query.getResultCount()); // must ignore `firstResult` and `maxResults`
result.setData(query.getResultList());
```

gavinking:
Done!

PR #660, opened 2024-09-20
add Query.getResultCount()

gavinking:
see issue #622

### DIFF
--- a/api/src/main/java/jakarta/persistence/Query.java
+++ b/api/src/main/java/jakarta/persistence/Query.java
@@ -11,6 +11,7 @@
  */
 
 // Contributors:
+//     Gavin King      - 4.0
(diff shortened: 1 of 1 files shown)


gavinking:
Thanks @njr-11, appreciated.

Since this method has a trivial but correct implementation (just execute the query and call `size()` on the resulting list), it doesn't make JPA harder to implement. On the other hand this is a very convenient thing to have.

So I'm just going to merge this.

njr-11:
> this method has a trivial but correct implementation (just execute the query and call `size()` on the resulting list)

Good point - this isn't going to be useful for pagination if an implementation does that, leading to incorrect sizes for large datasets and worse performance.  I opened #777 to address that.

Issue #473, opened 2023-08-19
allow @NamedQuery and @NamedNativeQuery more control over query execution


gavinking:
Currently, a `@NamedQuery` or ` @NamedNativeQuery` may specify query hints using the verbose and stringly-typed annotation `@QueryHint`.

I propose making certain "hints" available in a typesafe way, for example:

```java
public @interface NamedQuery {
    ...
    FlushModeType flushMode() default AUTO;
    CacheStoreMode cacheStoreMode() default USE;
    CacheRetrieveMode cacheRetrieveMode() default USE;
    int timeout() default -1;
}
```

In the case of `@NamedNativeQuery`, it might even be useful to add a way to specify the tables or entities affected by a native query, to allow auto-flush to work correctly:

```java
public @interface NamedNativeQuery {
    ...
    FlushModeType flushMode() default AUTO;
    CacheStoreMode cacheStoreMode() default USE;
    CacheRetrieveMode cacheRetrieveMode() default USE;
    int timeout() default -1;

   // control over auto-flush

    String[] affectedTables() default {};
    Class<?>[] affectedEntities() default {};
}
```

gavinking:
@njr-11 spotted that this proposed change has an unacceptable impact on backward compatibility.

It looks like the developing consensus is that #780 offers an alternative that is viable, and more oriented toward the future (i.e. is better integrated with Jakarta Data), whereas this proposal was in improvement to stuff that's more legacy-oriented.

So I'm going to close this issue in favor of #780.

PR #662, opened 2024-09-20
add cache modes and timeout to @NamedQuery, @NamedNativeQuery

gavinking:
see issue #473

### DIFF
--- a/api/src/main/java/jakarta/persistence/NamedNativeQuery.java
+++ b/api/src/main/java/jakarta/persistence/NamedNativeQuery.java
@@ -11,6 +11,7 @@
  */
 
 // Contributors:
+//     Gavin King       - 4.0
 //     Gavin King       - 3.2
 //     Petros Splinakis - 2.2
 //     Linda DeMichiel  - 2.1
@@ -96,12 +97,6 @@ public @interface NamedNativeQuery {
     String query();
 
     /**
-     * Query properties and hints.
-     * (May include vendor-specific query hints.)
-     */
-    QueryHint[] hints() default {};
-
-    /**
      * The class of each query result. If a {@link #resultSetMapping
      * result set mapping} is specified, the specified result class
      * must agree with the type inferred from the result set mapping.
@@ -145,4 +140,34 @@ public @interface NamedNativeQuery {
      * @since 3.2
      */
     ColumnResult[] columns() default {};
+
+    /**
+     * (Optional) The {@linkplain CacheStoreMode cache store mode}
+     * to use in query execution.
+     * @since 4.0
+     * @see Query#setCacheStoreMode
+     */
+    CacheStoreMode cacheStoreMode() default CacheStoreMode.USE;
+
+    /**
+     * (Optional) The {@linkplain CacheRetrieveMode cache retrieve mode}
+     * to use in query execution.
+     * @since 4.0
+     * @see Query#setCacheRetrieveMode
+     */
+    CacheRetrieveMode cacheRetrieveMode() default CacheRetrieveMode.USE;
+
+    /**
+     * (Optional) A query timeout in milliseconds. By default,
+     * there is no timeout.
+     * @since 4.0
+     * @see Query#setTimeout
+     */
+    int timeout() default -1;
+
+    /**
+     * Query properties and hints.
+     * (May include vendor-specific query hints.)
+     */
+    QueryHint[] hints() default {};
 }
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/persistence/NamedQuery.java
```diff
@@ -86,9 +87,34 @@
...
+     */
+    CacheStoreMode cacheStoreMode() default CacheStoreMode.USE;
+
+    /**
+     * (Optional) The {@linkplain CacheRetrieveMode cache retrieve mode}
+     * to use in query execution.
+     * @since 4.0
+     * @see Query#setCacheRetrieveMode 
+     */
+    CacheRetrieveMode cacheRetrieveMode() default CacheRetrieveMode.USE;
```
njr-11: I just though of a problem.  The default values  here will override any corresponding value that was set on the EntityManager, potentially causing a breaking change in behavior.  To do this without a breaking change, we would need a different default representing UNSPECIFIED.

gavinking: Ah, good point. Yes, we need to think this through.

gavinking: > we would need a different default representing `UNSPECIFIED`

Damn, I can't really think of a better option than this one. And I don't like it much, because setting it onto an `EntityManager` would have to be an error.

Maybe we should just put this issue on ice.

gavinking: So this was a problem we simply didn't run into when we added these members to Hiberate's `@NamedQuery` annotation, because in Hibernate the annotation historically *always* had control over the cache modes and there was never any delegation-to-EM with that annotation.

gavinking:
I have moved this one back to draft mode, because it results in a backward-incompatible change to behavior, as spotted by Nathan.

gavinking:
I have proposed a solution, by adding `CacheStoreMode.DEFAULT` and `CacheRetrieveMode.DEFAULT`. I'm not sure I love it, but I wanted to see what you guys think.

api/src/main/java/jakarta/persistence/CacheStoreMode.java
```diff
@@ -41,22 +41,49 @@
...
+     * context.
+     *
+     * @see EntityManager#getCacheStoreMode()
      */
     USE,
 
     /**
-     * Don't insert into cache. 
+     * Specifies that entity data may not be inserted into the
+     * second-level cache when read from the database, nor when
```
njr-11:
```suggestion
     * Specifies that entity data may neither be inserted into the
     * second-level cache when read from the database, nor when
```

api/src/main/java/jakarta/persistence/CacheStoreMode.java
```diff
@@ -41,22 +41,49 @@
...
+     * written to the database.
      */
     BYPASS,
 
     /**
-     * Insert/update entity data held in the cache when read from
-     * the database and when written to the database. Force refresh
-     * of cache for items read from database.
+     * Specifies that entity data must be inserted or updates in
```
njr-11:
```suggestion
     * Specifies that entity data must be inserted or updated in
```

api/src/main/java/jakarta/persistence/NamedNativeQuery.java
```diff
@@ -145,4 +140,34 @@
...
+     */
+    CacheStoreMode cacheStoreMode() default CacheStoreMode.USE;
+
+    /**
+     * (Optional) The {@linkplain CacheRetrieveMode cache retrieve mode}
+     * to use in query execution.
+     * @since 4.0
+     * @see Query#setCacheRetrieveMode
+     */
+    CacheRetrieveMode cacheRetrieveMode() default CacheRetrieveMode.USE;
```
njr-11: These were missed:
```suggestion
    CacheStoreMode cacheStoreMode() default CacheStoreMode.DEFAULT;

    /**
     * (Optional) The {@linkplain CacheRetrieveMode cache retrieve mode}
     * to use in query execution.
     * @since 4.0
     * @see Query#setCacheRetrieveMode
     */
    CacheRetrieveMode cacheRetrieveMode() default CacheRetrieveMode.DEFAULT;
```

gavinking:
I've moved the Javadoc changes to #778, since I'm still not convinced but this is a net improvement.

gavinking:
Closing, since #780 is a more promising direction than #473.

Issue #62, opened 2013-06-22
Load entities in read-only mode


lukasj:
Relational databases typically benefit from the knowledge, whether a transation will potentially modify and information (so locks are needed), or only read-only queries are executed (so no locks are needed). For similar reason, EclipseLink (and hopefully other JPA implementations, too) know query hints for "ready-only".

Unfortunately when using such vendor-specific hints, this will induce the problem that a portable application must know all these hints for all JPA implementations (or there will be no Performance gain for the unknown ones). This is not smart from the view of an ISV.

Hence I want to propose that the next maintenance release of the JPA specification defines a unique name and value to enable the read-only query mode independently of the actual JPA implementation.

Proposal: A compliant implementation which has a read-only query mode MUST enable this read-only query mode when the "javax.persistence.readonly" with a value of "true" is provided.

mkarg:
@mkarg I was the original reporter.

gavinking:
> For clarity's sake, EclipseLink's "read-only" hint regards how it manages the first-level cache during a query; it's not about a making the transaction read-only.

Hibernate has a very similar feature, with a very similar semantic.

So this is definitely something we should discuss standardizing at some stage. (However, I would advocate doing it in a more typesafe way, not via a strongly-typed hint.)

gavinking:
> (However, I would advocate doing it in a more typesafe way, not via a stringly-typed hint.)

For example, if #454 goes through, it could be a `FindOption`.

gavinking:
Actually, if #454 _does_ go through, I think we should even consider including this in 3.2.

gavinking:
> Actually, if #454 _does_ go through, I think we should even consider including this in 3.2.

Well, it did not make it into 3.2, but H7 introduces a `ReadOnlyMode` enum, which is a `FindOption`, which I think we should promote to the spec. We would also add `setReadOnly()` to `Query`.

gavinking:
I have proposed `ManagedEntityMode` and related APIs.

There are two open issues here:

1. How "deep" is the effect of the `ManagedEntityMode`? Are eagerly-fetched associated entities also read-only?
2. When happens when I request an entity a second time in a different mode? For example, does a second call to `find()` "override" the read-only mode of an already-loaded entity? (I think it should.)

gavinking:
A third question is:

- when an entity is in read-only mode, do cascades still get processed at flush time? (I suppose they should not be.)

gavinking:
[...]

PR #668, opened 2024-09-20
introduce ManagedEntityMode

gavinking:
finally addresses issue #62


beikov:
The `readOnly` hint on a transaction should IMO affect the default managed entity mode of a session: https://github.com/jakartaee/transactions/issues/220

gavinking:
> The `readOnly` hint on a transaction should IMO affect the default managed entity mode of a session:

Indeed, that would be ideal.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -276,6 +276,48 @@ After an entity has been removed, its state
...
+
+If a managed entity loaded in read-only mode would be returned by a method
+of `EntityManager` which does not request `READ_ONLY` mode, or among the
+results of a query which does not request `READ_ONLY` mode, a
+`PersistenceException` is thrown.
+
+If an entity loaded in read-only mode is passed to the `refresh()` method
+of `EntityManager`, its mode is automatically reset to `READ_WRITE`. If
+an entity loaded in read-only mode is passed to the `remove()` or `lock()`
+method of `EntityManager`, a `PersistenceException` is thrown.
```
gavinking: I'm not at all sure that these are the right things to say here. Thoughts?

beikov: Looks good to me, though if we add `@ReadOnly`, we might have to adapt this passage a bit.

gavinking: @beikov I changed all that, because I realized that that definition is just incredibly fragile.

That's completely different, and much more complex, in subsequent commits.

Issue #230, opened 2019-10-15
Could we please add this hibernate annotation to the spec: @OptimisticLock(excluded = true)


rainerschamm:
Adding this or something similar would make optimistic locking a lot more useful.

Another suggestion would be to make: LockModeType.NONE
on and em.find(id, ..., LockModeType.NONE)
actually skip the version increment.

rainerschamm:
Oh yea, sorry. I completely forgot about this. To be honest I only remember requesting this on the mailing list at the time.
Please merge or delete.

gavinking:
@anija-anil @njr-11 while updating the spec for `@ExcludeFromVersioning`, I realized that there's an alternative way to do this that's worth considering. 

Instead of:

```java
@ExcludeFromVersioning
LocalDateTime lastAccessed;
```

We could hang this info off the `@Basic` annotation, so:

```java
@Basic(versioned=false)
LocalDateTime lastAccessed;
```

where `versioned` would default to `true`.

What do you guys think?

I had tunnel-visioned on this being a separate annotation because that's what it is in Hibernate, and that's what's proposed in this issue description, but it doesn't need to be, and an element of `@Basic` is arguably simpler in some sense.

gavinking:
> an element of @Basic is arguably simpler in some sense.

Ahhhh wait, no, scratch that, if it were an element then we would have to also add it to (at least) `@ManyToOne` and `@ElementCollection`, and probably also to `@OneToMany`, `@ManyToMany`, and `@OneToOne`.

So the separate annotation is simpler.

gavinking:
Also, if we stick with the separate annotation (probably better, I guess) then maybe it should be called `ExcludedFromVersioning` (an adjectival phrase) instead of `ExcludeFromVersioning` (a verb phrase).

njr-11:
I like the name `@ExcludedFromVersioning`.

I tried to think of some other appropriate annotation names just to see if other alternatives might make sense, and it is difficult to find a name that concisely indicates what this does without adding in the possibility of misinterpretation.  One idea that I thought of is `@NonVersionIncrementing`, but I wouldn't say that is any better than `@ExcludedFromVersioning`.

[...]

PR #669, opened 2024-09-20
introduce @ExcludeFromVersioning

gavinking:
finally addresses issue #230

### DIFF
--- a/api/src/main/java/jakarta/persistence/ExcludeFromVersioning.java
+++ b/api/src/main/java/jakarta/persistence/ExcludeFromVersioning.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2008, 2024 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+//     Gavin King      - 4.0
+
+package jakarta.persistence;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.FIELD;
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Specifies that mutation of the annotated field or property does
+ * not result in incrementation of the {@linkplain Version version
+ * field or property} of the entity.
+ * <p>
+ * This annotation indicates that the application tolerates lost
+ * updates to the annotated field or property when concurrent
+ * optimistic transactions modify the same entity instance, and
+ * that no {@link OptimisticLockException} should occur in this
+ * scenario.
+ *
+ * @see Version
+ *
+ * @since 4.0
+ */
+@Target({METHOD, FIELD})
+@Retention(RUNTIME)
+public @interface ExcludeFromVersioning {}
--- a/api/src/main/java/jakarta/persistence/Version.java
+++ b/api/src/main/java/jakarta/persistence/Version.java
@@ -67,6 +67,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  *
  * @see LockModeType
  * @see PersistenceUnitUtil#getVersion(Object)
+ * @see ExcludeFromVersioning
  *
  * @since 1.0
  */


api/src/main/java/jakarta/persistence/ExcludeFromVersioning.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) 2008, 2024 Oracle and/or its affiliates. All rights reserved.
```
anija-anil: Since this PR is being merged this year, could we update the copyright year to reflect 2025?

gavinking:
```suggestion
 * Copyright (c) 2008, 2025 Oracle and/or its affiliates. All rights reserved.
```

gavinking: Done.

gavinking:
> LGTM

Thanks @anija-anil, so now I've also added this to the list of annotations in the specification.

gavinking:
OK, I have renamed it to `@ExcludedFromVersioning`.

I believe this one is ready to merge now.

api/src/main/java/jakarta/persistence/ExcludedFromVersioning.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) 2008, 2025 Oracle and/or its affiliates. All rights reserved.
```
pzygielo: This looks like new file. Is it correct to use 2008 then?

I'm sorry for asking - are you Oracle-tied?

gavinking:
```suggestion
 * Copyright (c) 2025, 2025 Oracle and/or its affiliates. All rights reserved.
```

api/src/main/java/jakarta/persistence/ExcludedFromVersioning.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) 2025, 2025 Oracle and/or its affiliates. All rights reserved.
```
anija-anil: The two years mean the year the file was created and the last updated year. If it's a new file, then it's just 2025.
Copyright (c) 2025 ...

gavinking:
```suggestion
 * Copyright (c) 2025 Oracle and/or its affiliates. All rights reserved.
```

api/src/main/java/jakarta/persistence/ExcludedFromVersioning.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) 2025 Oracle and/or its affiliates. All rights reserved.
```
pzygielo, line 2: Still don't know - how is Oracle involved here?

gavinking: Hi, @pzygielo, FYI I'm an *extremely* busy guy with many other much more important things on my plate right now than worrying about copyright headers. This is the copyright notice that accompanies all the JPA source code, and I simply do not have time to waste on investigating what would be involved in changing it. If you're prepared to do that investigation, and propose a pull request changing all the headers, you're very welcome to do so. That's the beauty of this being an open source project.

pzygielo: @waynebeaton, @ivargrimstad - unless you are also *extremely* busy with other much more important things - can you please confirm that it's no longer necessary to waste the time on copyright headers?

> a pull request changing all the headers

Is that correct that all files would need to be updated?

This https://www.eclipse.org/projects/handbook/#ip-copyright-headers:

> Contributors should be consistent when expressing copyright on a particular file.

suggests to me it's per-file, so in this case the new file should follow guidelines IMO.

pzygielo: > This is the copyright notice that accompanies all the JPA source code,

Should you use something like
https://github.com/jakartaee/persistence/blob/6155bf465fe03c63b330c40038dd7f03f1d40f19/api/src/main/java/jakarta/persistence/AttributeOverrides.java#L2

I'd be silent.

gavinking:
Look, I really don't want to lose my Wednesday morning debating this. However:

1. There's an actual `copyright-template` file in this project, and every source file except that *one* single file you just found follows it. I don't know what the history of that one file is, or why it's different from all the others.
2. On every open source project I've ever worked on, the source code all had uniform copyright headers, since managing divergent headers is an obvious nightmare, and there's no chance of keeping the headers correct over time. 
3. I assume that the foundation has written policies on the format of the headers, so, for now, I'm going to continue to follow the format of the `copyright-template` file, which was created during the time period I was not working on JPA, and so I don't know the considerations which went into its creation.

I now request that you take this discussion to a more appropriate forum. Thanks in advance.

pzygielo:
> Look, I really don't want to lose my Wednesday morning debating this. However:

Forgive me. I'm only trying to help by commenting what I see and find as mistake. And I don't understand how Oracle is involved, other way than just being referenced in other files.

>     1. There's an actual `copyright-template` file in this project, 

This is not a file for people to follow, but for the copyright maven plugin to use.

>     3. I assume that the foundation has written policies on the format of the headers, 

Yes! As a handbook section you must be familiar with: https://www.eclipse.org/projects/handbook/#ip-copyright-headers

> I now request that you take this discussion to a more appropriate forum. 

What would it be, if I have concerns regarding this specific contribution?

But I have no more comments.

gavinking:
This has been sitting open for 13 months now, and the issue itself is six years old. 

So a huge thanks to @anija-anil for helping out with the review.

Issue #650, opened 2024-08-30
add PersistenceProvider.getClassTransformer()


gavinking:
Currently the Jakarta EE container can only get hold of the `ClassTransformer` for a persistence provider *after* the persistence unit has been started. The provider has to call `PersistenceUnitInfo.addTransformer()`. This is a pretty weird design and results in chicken/egg dependencies for the container.

But we can solve this problem by decoupling the lifecycle here, and adding a method `PersistenceProvider.createClassTransformer(PersistenceUnitInfo)`, allowing enhancement and bootstrap of the persistence unit to occur in two phases.

UPDATE: just to expand on the problem here, because other folks have started to run into this issue, the problem is that:

-  the container needs to enhance entities as it is loading classes from the EAR, 
- but it can't get hold of the enhancer (the `ClassTransformer`) until it calls `PersistenceProvider.createContainerEntityManagerFactory()` to create the persistence unit.

This is a really broken design IMO, since creating the `EntityManagerFactory` should be something that can happen well after the classes are loaded by the container.

gavinking:
A second thing I would like to do here is to clearly place the responsibility for scanning for entity classes back to the container. I think it's a mistake to place this responsibility on the persistence provider.

This can be easily solved by adding `getAllManagedClassNames()` to `PersistenceUnitInfo`.

gavinking:
See the proposal here: https://github.com/jakartaee/persistence/pull/671/files

scottmarlow:
https://github.com/wildfly/wildfly/pull/18863 proves that WildFly will be able to implement PersistenceProvider.getClassTransformer() with Hibernate ORM in that https://github.com/wildfly/wildfly/pull/18863 passes all of the WildFly unit testing.

scottmarlow:
> A second thing I would like to do here is to clearly place the responsibility for scanning for entity classes back on the container. I think it's a mistake to place this responsibility on the persistence provider.
> 
> This can be easily solved by adding `getAllManagedClassNames()` to `PersistenceUnitInfo`.

https://github.com/wildfly/wildfly/blob/main/jpa/hibernate6/src/main/java/org/jboss/as/jpa/hibernate/HibernateArchiveScanner.java might be an example of the container doing the scanning for entity classes for Hibernate ORM integration.  

[...]

PR #671, opened 2024-09-20
add PUI.getAllManagedClassNames and PP.getClassTransformer

gavinking:
Straw man for discussion.

see issue #650

### DIFF
--- a/api/src/main/java/jakarta/persistence/spi/PersistenceProvider.java
+++ b/api/src/main/java/jakarta/persistence/spi/PersistenceProvider.java
@@ -11,6 +11,7 @@
  */
 
 // Contributors:
+//     Gavin King      - 4.0
 //     Gavin King      - 3.2
 //     Linda DeMichiel - 2.1
 //     Linda DeMichiel - 2.0
@@ -138,5 +139,37 @@ public interface PersistenceProvider {
      * @since 2.0
      */
     ProviderUtil getProviderUtil();
+
+    /**
+     * Obtain a transformer supplied by the provider that is called
+     * for every new class definition or class redefinition that gets
+     * loaded by the loader returned by the
+     * {@link PersistenceUnitInfo#getClassLoader} method. The
+     * transformer has no effect on the result returned by the
+     * {@link PersistenceUnitInfo#getNewTempClassLoader} method.
+     * Classes are only transformed once within the same classloading
+     * scope, regardless of how many persistence units they may be
+     * a part of.
+     * <p>The given instance of {@link PersistenceUnitInfo} may
+     * return {@code null} when any one the accessor methods
+     * {@link PersistenceUnitInfo#getClassLoader()},
+     * {@link PersistenceUnitInfo#getJtaDataSource()}, or
+     * {@link PersistenceUnitInfo#getNonJtaDataSource()} is called
+     * by an implementation of this method.
+     * <p>If the container calls this method before invoking
+     * {@link #createContainerEntityManagerFactory} to create the
+     * {@link EntityManagerFactory}, then the transformer returned
+     * by this method is used instead of any transformer registered
+     * via {@link PersistenceUnitInfo#addTransformer}. The container
+     * is not required to call this method.
+     * @return  provider-supplied transformer that the
+     * container invokes at class-(re)definition time
+     * @param info  metadata for use by the persistence provider
+     * @param map  a Map of integration-level properties for use
+     * by the persistence provider, which will not usually contain
+     * a {@code ValidatorFactory} or {@code BeanManager}.
+     * @since 4.0
+     */
+    ClassTransformer getClassTransformer(PersistenceUnitInfo info, Map<?, ?> map);
 }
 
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/persistence/spi/PersistenceProvider.java
```diff
@@ -138,5 +139,37 @@ public interface PersistenceProvider {
...
     ProviderUtil getProviderUtil();
+
+    /**
+     * Obtain a transformer supplied by the provider that is called
+     * for every new class definition or class redefinition that gets
+     * loaded by the loader returned by the
+     * {@link PersistenceUnitInfo#getClassLoader} method. The
+     * transformer has no effect on the result returned by the
+     * {@link PersistenceUnitInfo#getNewTempClassLoader} method.
+     * Classes are only transformed once within the same classloading
```
scottmarlow, line 150: I think the wording is fine but wanted to point out that the transformer may need to deal with concurrency issues when multiple EE threads try to load the same application class at the same time.  I think we are saying that each class is never transformed more than once which is true even with multiple threads loading the same class at the same time (e.g. as each application class should only be defined exactly once).

api/src/main/java/jakarta/persistence/spi/PersistenceProvider.java
```diff
@@ -138,5 +139,37 @@ public interface PersistenceProvider {
...
+     * {@link PersistenceUnitInfo#getClassLoader()},
+     * {@link PersistenceUnitInfo#getJtaDataSource()}, or
+     * {@link PersistenceUnitInfo#getNonJtaDataSource()} is called
+     * by an implementation of this method.
+     * <p>If the container calls this method before invoking
+     * {@link #createContainerEntityManagerFactory} to create the
+     * {@link EntityManagerFactory}, then the transformer returned
+     * by this method is used instead of any transformer registered
+     * via {@link PersistenceUnitInfo#addTransformer}. The container
+     * is not required to call this method.
```
scottmarlow, line 164: :+1:

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
```diff
@@ -19,6 +20,7 @@
 
 import javax.sql.DataSource;
 import java.util.List;
+import java.util.Map;
```
scottmarlow: I don't think we are using java.util.Map with the below changes made to api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java, am I missing something?

gavinking: Thanks, fixed.

api/src/main/java/jakarta/persistence/spi/PersistenceProvider.java
gavinking: > the transformer may need to deal with concurrency issues when multiple EE threads try to load the same application class at the same time

I really think that's a problem for the container to deal with. There's no requirement that the EE container use multiple threads for class loading, and so if it chooses to do that, it's going to have to deal with the consequences.

> I think we are saying that each class is never transformed more than once which is true even with multiple threads loading the same class at the same time 

I have no special insight into what the following sentence means, since I just copy/pasted what was already there:

>  Classes are only transformed once within the same classloading scope, regardless of how many persistence units they may be a part of.

I understand it to be saying that if a single entity class is a member of multiple persistence units that happen to share a single class loader, then the entity class is only transformed once.

What I don't understand is who is responsible for ensuring that? Container or provider? That Javadoc doesn't seem to very clearly answer that, but this is really not my area.

scottmarlow: > I really think that's a problem for the container to deal with. There's no requirement that the EE container use multiple threads for class loading, and so if it chooses to do that, it's going to have to deal with the consequences.

Sure the container does take responsibility but the persistence provider gets the more difficult part of also dealing with concurrency as well which might be surprising.  

> I understand it to be saying that if a single entity class is a member of multiple persistence units that happen to share a single class loader, then the entity class is only transformed once.

+1 for keeping what we have, I just wanted to talk a bit more to see if we see the underlying requirements the same.

api/src/main/java/jakarta/persistence/spi/PersistenceProvider.java
```diff
@@ -138,5 +139,37 @@ public interface PersistenceProvider {
...
+     * for every new class definition or class redefinition that gets
+     * loaded by the loader returned by the
+     * {@link PersistenceUnitInfo#getClassLoader} method. The
+     * transformer has no effect on the result returned by the
+     * {@link PersistenceUnitInfo#getNewTempClassLoader} method.
+     * Classes are only transformed once within the same classloading
+     * scope, regardless of how many persistence units they may be
+     * a part of.
+     * <p>The given instance of {@link PersistenceUnitInfo} may
+     * return {@code null} when any one the accessor methods
```
njr-11:
```suggestion
     * return {@code null} when any of the accessor methods
```

gavinking:
@scottmarlow Can I go ahead and merge this ... we won't know for sure if it's right until you guys try implementing it.

scottmarlow (approved):

gavinking:
Thanks folks!

Issue #437, opened 2023-08-09
Allow passing in an entity graph on EntityManager.refresh 


trajano:
At present I do the following to reload an entity such that all the fields in the entity graph provided are loaded

```
    entityManager.detach(entity);
    final var entityType = entityManager.getMetamodel().entity(entity.getClass());
    final var entityGraph = entityManager.getEntityGraph(entityType.getName());
    return (T)
        entityManager.find(
            entity.getClass(),
            entity.getId(),
            Map.of(
                "jakarta.persistence.loadgraph", // or "jakarta.persistence.fetchgraph"
                entityGraph,
                "jakarta.persistence.cache_retrieve_mode",
                CacheRetrieveMode.BYPASS));
```

It would be cleaner I think if I can simply do

```
    final var entityType = entityManager.getMetamodel().entity(entity.getClass());
    final var entityGraph = entityManager.getEntityGraph(entityType.getName());
    entityManager.refresh(entity, 
            Map.of(
                "jakarta.persistence.loadgraph", // or "jakarta.persistence.fetchgraph"
                entityGraph));
   // and no cache retrieve mode
```

To reload the entity with the graph that was provided.

gavinking:
This would be useful, and will be easier to in light of the direction we've decided to take #383.

gavinking:
See #672

gavinking:
I am moving this one off the road map for 4.0. We have a lot on our plate and:

1. this one imposes an implementation burden and,
2. risk, since we do not have a prototype implementation anywhere yet.

PR #672, opened 2024-09-20
add overload of refresh() accepting an EntityGraph

gavinking:
see issue #437

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -886,6 +886,26 @@ public interface EntityManager extends AutoCloseable {
                  RefreshOption... options);
 
     /**
+     * Refresh the state of the given managed entity instance from the
+     * database, along with all associated entities reachable by
+     * following the given {@link EntityGraph}, overwriting changes
+     * made to the entity, if any. This operation does not cascade to
+     * associations marked {@link CascadeType#REFRESH cascade=REFRESH}
+     * unless they are included in the given entity graph.
+     * @param entity  a managed entity instance
+     * @throws IllegalArgumentException if the instance is not an entity
+     *         or if the entity is not managed
+     * @throws TransactionRequiredException if there is no
+     *         transaction when invoked on a container-managed
+     *         entity manager of type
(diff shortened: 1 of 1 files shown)


sebersole:
Or just make EntityGraph implement RefreshOption, which allows avoidning the overload.

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -885,6 +885,26 @@ void refresh(Object entity, LockModeType lockMode,
...
+     *         or if the entity is not managed
+     * @throws TransactionRequiredException if there is no
+     *         transaction when invoked on a container-managed
+     *         entity manager of type
+     *         {@link PersistenceContextType#TRANSACTION}
+     * @throws EntityNotFoundException if the entity no longer exists in
+     *         the database
+     * @since 4.0
+     */
+    <T> void refresh(T entity, EntityGraph<T> entityGraph);
```
gavinking: So we either need to add `RefreshOption...` as an additional parameter here, or do as @sebersole suggests, and make `EntityGraph` a `RefreshOption`.

gavinking: Actually I wonder if this should be:

```java
<T> void refresh(T entity, EntityGraph<? super T> entityGraph, RefreshOption... options);
```

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -885,6 +885,29 @@ void refresh(Object entity, LockModeType lockMode,
...
+     *         or if the entity is not managed
+     * @throws TransactionRequiredException if there is no
+     *         transaction when invoked on a container-managed
+     *         entity manager of type
+     *         {@link PersistenceContextType#TRANSACTION}
+     * @throws EntityNotFoundException if the entity no longer exists in
+     *         the database
+     * @since 4.0
+     */
+    <T> void refresh(T entity, EntityGraph<T> entityGraph,
```
gavinking: This should probably be:

```suggestion
    <T> void refresh(T entity, EntityGraph<? super T> entityGraph,
```

PR #678, opened 2024-10-24
Polish ConstraintMode

quaff:
Replace tab with spaces for consistency.

Although the codebase is mixing usage of tab and spaces, but it's not good to do that in single source file, it's better to unify across the entire codebase, I like tab personally.


gavinking (approved):

gavinking:
> the codebase is mixing usage of tab and spaces

I actually was not aware of this.

quaff:
> > the codebase is mixing usage of tab and spaces
> 
> I actually was not aware of this.

Enforcing code style make the diff more concise.

Issue #684, opened 2024-11-11
@MapKey(name=...) inconsistent with @OrderBy(....)


gavinking:
`@MapKey` and `@OrderBy` are extremely conceptually-similar in that they each reference a field of the associated entity. They are both different to `@OrderColumn` and `@MapKeyColumn` in that these latter annotations provide _column_ names.

But `@MapKey` has a `name` member like `@XxxColumn` annotations instead of a `value` member like `@OrderBy`.

This is inconsistent and unnecessarily verbose.

I think we should add `value` and deprecate `name`.

gavinking:
In a way it should really be `@MapKey(mappedBy= .... )`.

PR #685, opened 2024-11-11
deprecate @MapKey(name=..) and replace with @MapKey(..)

gavinking:
see #684

### DIFF
--- a/api/src/main/java/jakarta/persistence/MapKey.java
+++ b/api/src/main/java/jakarta/persistence/MapKey.java
@@ -63,7 +63,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * public class Department {
  *     ...
  *     @OneToMany(mappedBy = "department")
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/MapKey.java
```diff
@@ -95,5 +95,19 @@
      * mapped as {@link IdClass}, an instance of the primary key
      * class is used as the key.
      */
+    String value() default "";
```
pzygielo, line 98: Shouldn't the doc say _If the {@code value} element is not specified_ instead?

gavinking: Yes, thanks, good catch. I will fix that.

gavinking: See https://github.com/jakartaee/persistence/pull/748

Issue #687, opened 2024-11-23
entity listener with multiple @PrePersist (for example) methods


gavinking:
I can't find where we specify whether this is allowed:

```java
class EntityListener {
   @PrePersist void onPersistBook(Book book) {}
   @PrePersist void onPersistAuthor(Author book) {}
}
```
I think it _should_ be allowed.

gavinking:
Oh, wait, I did find it, buried in the section on XML:

> An entity listener class can define multiple
callback methods. However, at most one method of an entity listener
class can be designated as a pre-persist method, post-persist method,
pre-remove method, post-remove method, pre-update method, post-update
method, and/or post-load method, regardless of whether the XML
descriptor is used to define entity listeners or whether some
combination of annotations and XML descriptor elements is used.

But I think we should change this.

gavinking:
This improvement poses only a modest burden on implementers. So consider it done.

PR #688, opened 2024-11-23
specify that an entity listener may have multiple callbacks of same type

gavinking:
Also write proper javadoc, and clean up the spec.

see #687

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityListeners.java
+++ b/api/src/main/java/jakarta/persistence/EntityListeners.java
@@ -27,10 +27,60 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * entity or mapped superclass. This annotation may be applied 
  * to an entity class or mapped superclass.
  *
+ * <pre>
+ * &#64;Entity
+ * &#64;EntityListeners(BookObserver.class)
+ * class Book { ... }
+ * </pre>
+ *
  * <p>The specified entity listener classes may have callback
  * methods annotated {@link PrePersist}, {@link PreUpdate},
  * {@link PreRemove}, {@link PostPersist}, {@link PostUpdate},
- * and/or {@link PostRemove}.
+ * {@link PostRemove}, and/or {@link PostLoad}. A callback
+ * method defined by an entity listener class must have the
+ * signature {@code void method(E entity)} where {@code E} is
+ * a supertype of an entity class or mapped superclass to which
+ * the entity listener applies.
+ *
+ * <pre>
+ * class BookObserver {
+ *
+ *     &#64;PostPersist
+ *     void newBook(Book book) {
+ *         ...
+ *     }
+ *
+ * }
+ * </pre>
+ *
+ * <p>An entity listener class may have multiple callback
+ * methods for a given type of lifecycle event, but at most
+ * one callback method for a given type of event and given
+ * entity type.
+ *
+ * <pre>
+ * class Observer {
+ *
+ *     // called only for Books
+ *     &#64;PostPersist
+ *     void newBook(Book book) {
+ *         ...
+ *     }
+ *
+ *     // called only for Authors
+ *     &#64;PostPersist
+ *     void newAuthor(Author author) {
+ *         ...
+ *     }
+ *
+ *     // called for any entity type to which the listener applies
+ *     &#64;PostLoad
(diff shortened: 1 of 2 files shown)


api/src/main/java/jakarta/persistence/EntityListeners.java
```diff
@@ -27,17 +27,69 @@
...
+ *
  * <p>The specified entity listener classes may have callback
  * methods annotated {@link PrePersist}, {@link PreUpdate},
  * {@link PreRemove}, {@link PostPersist}, {@link PostUpdate},
- * and/or {@link PostRemove}.
+ * {@link PostRemove}, and/or {@link PostLoad}. A callback
+ * method defined by an entity listener class must have the
+ * signature {@code void method(E entity)} where {@code E} is
+ * a supertype of an entity class or mapped superclass to which
+ * the entity listener applies.
```
njr-11: Although this is technically accurate because Java considers a class to be a supertype of itself, I wonder if readers will understand or be confused here because the possibility of being the entity class, which is the most likely usage pattern, is not directly stated.  If that is a concern, here is one possible rewording,

```suggestion
 * signature {@code void method(E entity)} where {@code E} is
 * an entity class, mapped superclass, or supertype of the entity
 * class or mapped supertype to which the entity listener applies.
```

api/src/main/java/jakarta/persistence/EntityListeners.java
```diff
@@ -27,17 +27,69 @@
...
+ *         ...
+ *     }
+ *
+ * }
+ * </pre>
+ *
+ * <p>An entity listener class may have multiple callback
+ * methods for a given type of lifecycle event, but at most
+ * one callback method for a given type of event and given
+ * entity type.
```
njr-11: The entity type isn't always given because the given type can be a superclass instead.  If I understand the intent here, it should not be possible in the following example to have `@PostPersist void postPersist(Object o)` alongside the other `@PostPersist` methods because it would result in multiple PostPersist events for the same entity.

```suggestion
 * one callback method for a given type of event and entity
 * type.
```

njr-11 (commented):
This seems like a nice improvement and it shouldn't cause any breaking changes in behavior because it is adding support for a pattern that was previously disallowed.

One consequence I can think of is that it will no longer be possible to validate a callback listener on its own.  For example,

```
 class MyEntityListener {
     @PostPersist
     void postPersistTitled(Titled t) { ... }

     @PostPersist
     void postPersistReservable(Reservable r) { ... }

}
```
the above won't be known to be invalid until it is known that the entity
`Book implements Titled, Reservable`
exists.  I don't know if this matters but wanted to point it out in case it does.

gavinking:
@njr-11 

> One consequence I can think of is that it will no longer be possible to validate a callback listener on its own.

I guess it depends how you interpret this restriction:

>  an entity listener class may have multiple callback methods for a given  type of lifecycle event, but at most one callback method for a given type of event and given entity type.

Does "entity type" here mean that any entity class can only be assignable to one entity listener. Or does it just mean that each listener method must have a distinct parameter type.

I take the point of your other comments that I've used "entity type" in an ambiguous way here.

api/src/main/java/jakarta/persistence/EntityListeners.java
gavinking: I think this is what I really mean:

```suggestion
 * parameter type.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1370,6 +1372,14 @@ The following rules apply to lifecycle callback methods:
   contextfootnote:[Note that this caution applies also to the actions of
   objects that might be injected into an entity listener].
 
+An entity or entity listener may have multiple lifecycle callback methods:
+
+- an entity class may have at most one callback method for a given type
+  of lifecycle event, but
+- an entity listener class may have multiple callback methods for a given
+  type of lifecycle event, but at most one callback method for a given
+  type of event and given entity type.
```
gavinking: Same here:

```suggestion
  type of event and given parameter type.
```

gavinking:
WDYT now, @njr-11 ?

gavinking:
Alright! Let's do this.

PR #690, opened 2024-12-17
additional overload of CriteriaBuilder.between()

henriquediascampos:
Recently, I noticed the absence of a `Predicate` specification that tests whether a given literal value `v` is *between* two dynamic expressions `x` and `y`. 

In a real-world scenario, I encountered this need while checking if a specific date fell between two values stored in the database. At the time, I worked around this by combining "greater than" and "less than" conditions to achieve the desired result. However, wouldn't it be much cleaner and less verbose if we could express this condition directly in a clearer way?

This addition introduces a new `between` method that improves query readability and conciseness. I believe this enhancement would make the API more intuitive for developers who encounter similar use cases.

*(This is my first contribution; please let me know if I missed any steps or if further adjustments are needed. I appreciate your feedback!)*


api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -527,6 +527,15 @@ public interface CriteriaBuilder {
      */
     <Y extends Comparable<? super Y>> Predicate between(Expression<? extends Y> v, Y x, Y y);
 	
+    /**
+     * Create a predicate for testing whether the first argument is 
+     * between the second and third arguments in value.
+     * @param v  value 
+     * @param x  expression
+     * @param y  expression
+     * @return between predicate
```
gavinking, line 537:
```suggestion
     * @return between predicate
     * @since 4.0
```

gavinking (approved):
I have no objection to adding this.

gavinking:
@henriquediascampos You need to update the file header as so:

https://github.com/jakartaee/persistence/commit/f3f4a4c79bc29aa0b74571fa32501203da9df2f6

gavinking:
Also probably best to have an issue.

henriquediascampos:
@gavinking If you could help me with a question.
Should I still make the changes you requested?
I got a bit confused because I noticed you already pushed a commit (f3f4a4c) fixing what I had missed.
This is my first contribution, so I’m still a bit confused.

gavinking:
> I got a bit confused because I noticed you already pushed a commit

I can't push to your branch though.

PR #691, opened 2024-12-20
programmatic result set mappings

gavinking:
I've been promising this for *years* now. Here is a concrete proposal, finally.

See #9.


gavinking:
I've added integration with `Tuple`, which has not received a lot of love lately. This lets you write:

```java
 var isbn = column("isbn", String.class);
 var title = column("title", String.class);
 var pages = column("pages", int.class);

 entityManager.createNativeQuery("select isbn, title, pages from Book",
                                 tuple(isbn, title, pages))
     .getResultList()
     .forEach(tuple -> {
         String bookIsbn = tuple.get(isbn);
         String bookTitle = tuple.get(title);
         int bookPages = tuple.get(pages);
         ...
     });
```
Naturally this is completely type safe.

gavinking:
superseded by #803

Issue #704, opened 2025-03-17
Query.setEntityGraph()


gavinking:
The hints `jakarta.persistence.fetchgraph` and `jakarta.persistence.loadgraph` are defined as applying not only to `find()` but also to queries, but there's currently no type-safe way to specify an `EntityGraph` for a query (you have to use `setHint()`).

We should consider adding a `setEntityGraph()` method to `Query`, with the argument interpreted as a load graph. (Fetch graphs don't seem to make sense here.)

nxnen8282:
Our system attempted to address this issue:

Add type-safe setEntityGraph() method to Query and TypedQuery interfaces to allow specifying an EntityGraph for query execution without using raw setHint(). The method is interpreted as a load graph since fetch graphs don't make sense in query context.

Changes:
- Add setEntityGraph() method to Query interface
- Add covariant setEntityGraph() method to TypedQuery interface
- Add proper JavaDoc documentation following existing style

You can examine the suggested modifications here: https://github.com/nxnen8282/jakartaee-persistence-704/commit/279359016cd534c45358e926567969629ab0f9a4.


> [!CAUTION]
> Advisory: This solution was generated using automated tools, and we strongly recommend reviewing the code thoroughly before implementation. The solution may be partial - please consider it as inspiration for your own implementation.

---

http://missionfixinternet.com

gavinking:
Also add `entityGraph()` member to `@NamedQuery` to allow specification of a named entity graph.

gavinking:
An alternative idea is to add an overload of `createQuery()` which accepts an `EntityGraph` instead of a result class. This is consistent with the solution we adopted for `EntityManager.find()`.

gavinking:
Done in #705.

PR #705, opened 2025-03-17
add TypedQuery.setEntityGraph and @NamedQuery.entityGraph

gavinking:
see #704


api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -1054,6 +1054,24 @@ void refresh(Object entity,
...
+     * @param qlString a Jakarta Persistence query string
+     * @param resultGraph entity graph interpreted as a load graph
+     * @return the new query instance
+     * @throws IllegalArgumentException if the query string is
+     *         found to be invalid or if the query result is
+     *         found to not be assignable to the root type of
+     *         the entity graph
+     * @since 4.0
+     */
+    <T> TypedQuery<T> createQuery(String qlString, EntityGraph<T> resultGraph);
```
gavinking, line 1074: Perhaps this should be:

```suggestion
    <T> TypedQuery<? extends T> createQuery(String qlString, EntityGraph<T> resultGraph);
```

since we accept `EntityGraph`s for a supertype of the query result type.

gavinking: Actually I suppose there's no advantage to that, and only downside.

gavinking:
I wonder if we should also add:

```java
<T> TypedQuery<T> createQuery(CriteriaQuery<T> criteriaQuery, EntityGraph<? super T> graph);
```

gavinking:
> I wonder if we should also add:
> 
> ```java
> <T> TypedQuery<T> createQuery(CriteriaQuery<T> criteriaQuery, EntityGraph<? super T> graph);
> ```

It's not *necessary*, since you can write:

```java
em.createQuery(criteria).setEntityGraph(graph).getResultList();
```

But that's slightly more verbose than:

```java
em.createQuery(criteria, graph).getResultList();
```

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -160,6 +161,29 @@ default Stream<X> getResultStream() {
...
+
+    /**
+     * Retrieve the {@link EntityGraph} to be applied to the entity
+     * returned by the query, or {@code null} if no entity graph was
+     * specified via {@link #setEntityGraph(EntityGraph)} or
+     * {@link NamedQuery#entityGraph()}.
+     * @return the entity graph or {@code null}
+     * @since 4.0
+     */
+    EntityGraph<X> getEntityGraph();
```
gavinking: Should be:

```suggestion
    EntityGraph<? super X> getEntityGraph();
```

gavinking:
I think this is fine, and since nobody has spoken up with any objections, I'm going to merge it.

PR #720, opened 2025-04-24
deuglify the copypasta

gavinking:


gavinking:
@lukasj I have backed out the changes to the text and just retained the background color changes.

Issue #730, opened 2025-06-06
Subscription at run time to life cycle events


njr-11:
Background:
In Jakarta Data 1.1, @gavinking introduced an API to Jakarta Data to subscribe to pre/post insert/update/delete [life cycle events](https://github.com/jakartaee/data/tree/main/api/src/main/java/jakarta/data/event) and is also working on introducing new operations for merge/persist/refresh and so forth for stateful entities where the life cycle is managed by the JPA Persisetnce Provider that is used by the Jakarta Data Provider.

As a consequence of the latter, the Jakarta Data Provider must delegate tracking of the entity life cycle (when the insert/update/delete actually happen) to the Jakarta Persistence Provider.

Jakarta Persistence currently defines a couple of ways for applications to ask for life cycle notifications: by annotating entities or specifying in the orm.xml file.  But neither of these are something that a Jakarta Data provider can use.  One idea would be to standardize a new EntityManager property that can supplied to https://jakarta.ee/specifications/persistence/3.2/apidocs/jakarta.persistence/jakarta/persistence/entitymanagerfactory#createEntityManager(java.util.Map) to supply a listener for life cycle events across all entities that are managed by the EntityManager.

gavinking:
Seems to me that the most straightforward approach here would be to allow dynamic listener registration, for example:

```java
EntityManagerFactory emf = .... ;
emf.getEntityListeners(Book.class)
        .addListener(PostPersist.class, 
                book -> raisePostInsertEvent(book));
```

where:

1. the argument to `getEntityListeners()` is any entity class or any supertype of an entity class, including `Object.class`,
2. the first argument to `addListener()` is a callback annotation type, and
3. the second argument to `addListener()` is a function which accepts the entity instance.

gavinking:
See proposal in #731.

nlisker:
Not to hijack this issue, but I have a related scenario that I'd like covered. In my production code, the `@Entity` is required to register its `@EntityListener`s. However, when testing, I might want to add listeners to entities that are irrelevant to production code either to test the functionality or to mock functionality of a remote service. One way to do it is via an orm.xml, but the tendency is to have this setup in Java.

I was thinking about an interface that allows listening on the entity by implementing it in a listening class as a way to attach an "external" listener:
```java
class MyListener implements EntityListener<MyEntity> {

    @PostPersist
    ...
}
```

[...]

PR #731, opened 2025-06-15
add ability to register lifecycle callback event listeners via EMF

gavinking:
this is needed by Jakarta Data implementors

see #730

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityListenerRegistration.java
+++ b/api/src/main/java/jakarta/persistence/EntityListenerRegistration.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright (c) 2008, 2023 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0,
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+//     Gavin King - 4.0
+
+package jakarta.persistence;
+
+import java.util.function.Consumer;
+
+/**
+ * Represents the registration of an entity lifecycle event
+ * listener with the {@link EntityManagerFactory}. This object
+ * may be used to {@linkplain #cancel} the registration.
+ *
+ * @see EntityManagerFactory#addListener(Class, Class, Consumer)
+ *
+ * @since 4.0
+ */
+public interface EntityListenerRegistration {
+    /**
+     * Remove the listener.
+     */
+    void cancel();
+}
--- a/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
+++ b/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
@@ -17,6 +17,7 @@
 
 package jakarta.persistence;
 
+import java.lang.annotation.Annotation;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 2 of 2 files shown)


api/src/main/java/jakarta/persistence/EntityListenerRegistration.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (c) 2008, 2023 Oracle and/or its affiliates. All rights reserved.
```
njr-11:
```suggestion
 * Copyright (c) 2025 Oracle and/or its affiliates. All rights reserved.
```

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -371,6 +372,36 @@ public interface EntityManagerFactory extends AutoCloseable {
...
+     * and given entity type, returning an {@link EntityListenerRegistration}
+     * which may be used to {@linkplain EntityListenerRegistration#cancel cancel}
+     * the registration.
+     * <p>
+     * The given entity type may be an entity class or any supertype of
+     * an entity class, including {@code java.lang.Object}. The lifecycle
+     * event type is identified by the class of a callback annotation type,
+     * for example, {@link PostPersist PostPersist.class}. When a lifecycle
+     * event of the given type occurs for any entity assignable to the given
+     * entity type, the listener is invoked with the entity instance.
```
njr-11, line 387: The user will want to know if the listener might be invoked under the same transaction a the entity lifecycle operation, such that transactional operations performed by a listener would be rolled back/committed along with the lifecycle operation versus whether they happen independently of it.
It should be stated whether or not the listener is allowed to modify the supplied entity instance (and if so, what the consequence of it are), and whether or not it can be relied upon that the supplied entity instance reflects automatically generated or autoincremented values.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -371,6 +372,36 @@ public interface EntityManagerFactory extends AutoCloseable {
...
+     *
+     * @param entityType An entity class, or any supertype of an entity
+     *                   class
+     * @param callbackType The kind of entity callback, represented by
+     *                     the class of the callback annotation type
```
njr-11: Recommend listing the minimum set of callback annotations (the spec-defined ones) that must be supported by a Jakarta Persistence provider.

njr-11 (commented):
This will work great for Jakarta Data.  I added a few comments for additional clarification that would be helpful to users.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: Yeah, all this is already defined in the spec, but I agree it would be nice to have much more of that info in the javadoc. I actually already started work on that here: https://github.com/jakartaee/persistence/pull/642/files

njr-11: Yes, that's even better to define that information within the specific annotations rather than here because application code that uses those annotations directly will need to know the same things.

sebersole:
This looks good to me.  And useful probably for lots of integrations.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -371,6 +372,41 @@ public interface EntityManagerFactory extends AutoCloseable {
...
+     *
+     * @param entityType An entity class, or any supertype of an entity
+     *                   class
+     * @param callbackType The kind of entity callback, represented by
+     *                     the class of the callback annotation type,
+     *                     usually one of the following classes:
+     *                     {@link PrePersist}, {@link PostPersist},
+     *                     {@link PreUpdate}, {@link PostUpdate},
+     *                     {@link PreRemove}, {@link PreRemove},
+     *                     and {@link PostLoad}.
```
gavinking, line 397: @njr-11 I've added the list of callback annotation types you asked for.

But not that we're going to have to remember to update this list once the new callback types go in.

gavinking:
I've also added some text to the spec mentioning this stuff.

I think this is pretty much good to go.

gavinking:
I even remembered to update the header.

PR #737, opened 2025-07-22
columnDefinition does not specify a database identifier and should not be quoted

### LINKED ISSUE #736: spec claims that columnDefinitions should be quoted
@sebersole just pointed out to me that section 2.15 ("Naming of Database Objects") seems to say that the content of a `columnDefinition` should be quoted when `<delimited-identifiers/>` is enabled.

That just seems completely wrong, and this just looks to me like a clear mistake in the spec.

The examples of `columnDefinition` in the spec are things like:

```java
@Lob
@Column(name="DESC", columnDefinition="CLOB NOT NULL", table="EMP_DETAIL")
```

Nothing in `CLOB NOT NULL` should be quoted.

I think we should fix this.

gavinking:
fixes #736

### DIFF
--- a/spec/src/main/asciidoc/ch02-entities.adoc
+++ b/spec/src/main/asciidoc/ch02-entities.adoc
@@ -2400,7 +2400,7 @@ such.
 
 For example, assuming the use of an English
 locale, the following must be passed to the database as undelimited
-identifers so that they will be treated as equivalent for all databases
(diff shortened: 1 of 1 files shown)


gavinking:
Thanks Nathan!

I rebased and force-pushed to resolve the conflict.

PR #695, opened 2025-01-09
add getNamedEntityGraphs() to EntityType + introduce GraphFactory

gavinking:
see #693

PR #765, opened 2025-11-21
add EntityManager.getEntityGraph(Class, String)

### LINKED ISSUE #764: EntityManager.getEntityGraph(Class,String)
Currently there are two "good" ways to get a typed reference to a named entity graph:

1. `em.getEntityManagerFactory().getNamedEntityGraphs(Book.class).get("BookWithAuthors")`
2. `Book_._BookWithAuthors`

@marko-bekhta pointed out that neither of these is perfect for every use case:

- option (1) is sorta hard to discover, and quite verbose
- option (2) depends on the annotation processor

A very trivial enhancement would be to add a method with this signature to `EntityManager`:

```java
<T> EntityGraph<T> getEntityGraph(Class<T> rootType, String name);
```

which is in principle just a convenient way to do (1).

This definitely isn't a critical problem to solve, but given that it's so trivial, it might be worth doing anyway.

gavinking:
fixes #764

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2025 Oracle and/or its affiliates. All rights reserved.
  *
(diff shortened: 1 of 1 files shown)


gavinking:
Actually this pull request is a subset of what is proposed in #695. Probably better to just do #695.

PR #768, opened 2025-11-21
use asciidoc definition list for the exception summary

gavinking:
instead of weird bold titles


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -3341,37 +3341,35 @@ immediately by calls to `getOutputParameterValue`.
...
+PersistenceException ::
 
 The `PersistenceException` is thrown by the
 persistence provider when a problem occurs. It may be thrown to report
 that the invoked operation could not complete because of an unexpected
 error (e.g., failure of the persistence provider to open a database
-connection).
-
-All other exceptions defined by this
+connection). All other exceptions defined by this
```
njr-11: If you want to preserve the two paragraphs and have both indented, I think this would work,

```suggestion
connection).
+
All other exceptions defined by this
```

njr-11 (approved):

PR #770, opened 2025-11-22
add PessimisticLockScope.FETCHED and @NamedQuery.lockScope

### LINKED ISSUE #711: PessimisticLockScope.FETCHED
Currently `PessimisticLockScope` has two values. It's not entirely well-specified what either of them mean for a query, but I think it's clear what's intended for `find()` or `lock()` or `refresh()`.

1. for `NORMAL`, only the root entity is locked&mdash;associated entities and collections are not locked, but
2. for `EXTENDED`, state held in `@CollectionTable`s and `@JoinTable`s belonging to the root entity is also locked, but the associated entities are not locked.

It seems to me that `NORMAL` is a perfectly good default, and it's what you probably want 95% of the time.

Contrariwise, `EXTENDED` seems to have very strange semantics, and would seem to require locking state in lazy collection you're not even fetching. I'm struggling to imagine ever having a use for this.

On the other hand a thing which does seem like it could be useful is a new option, call it `FETCHED` or whatever, which specifies that all collections and entities fetched by `find()`, `refresh()`, or query are locked with the given lock level.

If this is a good and useful idea, it's easy enough to add. On the other hand, if it's *not*  worth adding `FETCHED`, then I would frankly consider deprecating `PessimisticLockScope` altogether. We have never seen any usage of `EXTENDED` in the Hibernate community. (In fact, I believe that our implementation is currently completely broken and not a single person except me has even noticed.)

gavinking:
fixes #711

also rewrite the Javadoc of PessimisticLockScope to be much clearer

### DIFF
--- a/spec/src/main/asciidoc/appA-revision-history.adoc
+++ b/spec/src/main/asciidoc/appA-revision-history.adoc
@@ -43,6 +43,8 @@ Added `Graph.getGraphedType()` and `AttributeNode.getAttribute()`
 
 Overloaded `addNamedQuery()` to take `TypedQuery` and return `TypedQueryReference`
 
+Added `FETCHED` to `PessimisticLockScope`, `lockScope` to `@NamedQuery`, and `setLockScope()` to `Query`
+
 `jakarta.annotation.Nonnull` now implies `optional=false`
 
 Allow `value` member of `@JoinColumns` to be defaulted


gavinking:
@sebersole Please read this carefully. WDYT?

api/src/main/java/jakarta/persistence/PessimisticLockScope.java
```diff
@@ -11,53 +11,95 @@
...
+     * </ul>
+     *
+     * <p>The pessimistic lock does not extend to data which was not
+     * fetched as part of the operation which obtained the pessimistic
+     * lock.
+     *
+     * @since 4.0
      */
-    EXTENDED
+    CASCADED
```
sebersole: It sounds like you intend this to correlate to Hibernate's FETCH scope?  I think that is a better term.  "Cascaded" makes me think there is a new `CascadeType#LOCK` that we base the scope on.

gavinking: > It sounds like you intend this to correlate to Hibernate's `FETCH` scope?

Yes.

>  I think that is a better term.

I'm think I'm fine with either.

gavinking: In Hibernate you called the options `INCLUDE_COLLECTIONS` and `INCLUDE_FETCHES`. But here we have `EXTENDED` and ???

sebersole: The current/legacy docs for EXTENDED were quite unclear and we missed locking owned join-tables in our original implementation. As our javadocs say, our impl is not entirely spec-compliant. We (Hibernate) could consider adding a fully spec compliant EXTENDED value/implementation if that is your point, I'm fine with that. But let's talk about that offline.

sebersole: As for what to call it here, I'd go with EXTENDED and FETCHED though I'd be more included to maybe phrase it along the lines of "we lock all rows which are fetched at the SQL/JDBC level".

gavinking: > though I'd be more included to maybe phrase it along the lines of "we lock all rows which are fetched at the SQL/JDBC level".

Why exactly?

My thinking is that, semantically, we only need to require implementations to lock rows which are actually made visible to the application. So it's fine to have this definition fairly high level.

api/src/main/java/jakarta/persistence/PessimisticLockScope.java
```diff
@@ -11,53 +11,95 @@
...
- * or used with the {@link NamedQuery} annotation.
+ * Controls how a {@linkplain LockModeType pessimistic lock} applied to
+ * an entity affects associated collections and relationships.
+ * <p>
+ * A {@code PessimisticLockScope} may be passed as a {@link FindOption}
+ * {@link RefreshOption}, or {@link LockOption}. Alternatively, it may
+ * be specified via the property {@code jakarta.persistence.lock.scope}.
+ * This property may be passed as an argument to methods of
+ * {@link EntityManager}, {@link Query}, and {@link TypedQuery} which
+ * accept a {@link LockModeType} or it may be used in the {@link NamedQuery}
```
njr-11: This statement seems partially wrong.
I see that EntityManager has methods that accept LockModeType along with properties that can include this.
But the only methods I see on Query and TypedQuery that accept LockModeType are,
```
setLockMode(LockModeType lockMode)
```
which does not have a properties argument that could take a `jakarta.persistence.lock.scope`

gavinking: It's called `setHint()`, and it's documented as:

> Set a query **property** or hint.

Yeah, don't look at me.

gavinking: Oh I suppose you want this to say:

```suggestion
 * This property may be passed as an argument to methods of
 * {@link EntityManager} which accept a {@link LockModeType} or to 
 * {@link Query#setHint}, or it may be used in the {@link NamedQuery}
```

njr-11: I should have clarified in my review comment that I wasn't clear on whether it needed to say something like that vs whether you had intended to provide some additional methods that had not been included.  This answers the question.  The new proposed wording is much better.

gavinking: OK, sorted.

gavinking:
Alright so I guess this is ready to go. I don't really see a need to let it stew further.

If there are implementability concerns, we can always declare it to be an optional feature. (There is plenty of precedent for that in JPA.)

Also the blast radius is small, meaning that this is quite easy to back out.

gavinking:
Done, thanks @njr-11  and @sebersole.

PR #772, opened 2025-11-23
improve javadoc surrounding optimistic locking

gavinking:
and standardize on terminology "optimistic locking conflict" instead of "optimistic lock failure"

inspired by #601 by @quaff

### DIFF
--- a/api/src/main/java/jakarta/persistence/Entity.java
+++ b/api/src/main/java/jakarta/persistence/Entity.java
@@ -65,7 +65,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * annotated {@link Id} or {@link EmbeddedId} holding the primary key
  * of the entity. An entity class may optionally have a field or
  * property annotated {@link Version}, which holds a value used to
- * detect optimistic lock failure.
+ * detect optimistic locking conflicts.
  *
  * <p>Fields or properties of an entity class are persistent by default.
  * The {@link Transient} annotation or the Java {@code transient} keyword
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -200,7 +200,7 @@ public interface EntityManager extends AutoCloseable {
      * This operation cascades to every entity related by an association
      * marked {@link CascadeType#PERSIST cascade=PERSIST}. If the given
      * entity instance is already managed, that is, if it already belongs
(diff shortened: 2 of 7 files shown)


gavinking:
> Should this be added to `executeUpdate` as well?  It seems like the need to flush and resulting possibility of optimistic locking conflict would also apply there.

Ummm. There is some very specific language in the spec about update queries. I will check but I believe the spec says they don't do optimistic lock checking. I don't recall what it says about flushing.

njr-11:
> There is some very specific language in the spec about update queries. I will check but I believe the spec says they don't do optimistic lock checking. I don't recall what it says about flushing.

I know the spec says it's the responsibility of the application, not the JPA provider, to perform version checking and updates for these queries.  But that isn't what I was thinking of here.  I was thinking of a scenario where, prior to `executeUpdate`, the application has updated a managed entity and those updates haven't yet been flushed to the database.  After that, the application attempts `executeUpdate`.  In order for `executeUpdate` to behave as expected with the data the appliacation previously modified within the persistence context, it seems like that data would need to first be flushed to the database - and that flush (not the `executeUpdate` JPQL itself) might encounter a optimistic locking conflict.

gavinking:
> Bulk update maps directly to a database update operation, bypassing optimistic locking checks. 
> Portable applications must manually update the value of the version column, if desired, and/or manually validate the value of the version column.
>
> The persistence context is not synchronized with the result of the bulk update or delete.
>
> [NOTE]
> 
> Caution should be used when executing bulk
> update or delete operations because they may result in inconsistencies
> between the database and the entities in the active persistence context.
> In general, bulk update and delete operations should only be performed
> within a transaction in a new persistence context or before fetching or
> accessing entities whose state might be affected by such operations._

That is pretty ambiguous, frankly.

I think it _intends_ to say that changes resulting from the SQL UPDATE don't affect in-memory state. But you could just-about read it as saying that no flush is required before execution.

gavinking:
I'm searching, and I can't find any unambiguous statements about this.

I think the right action to take would be to open an issue requesting a clarification.

njr-11:
> I think the right action to take would be to open an issue requesting a clarification.

Yeah, that would be nice to clarify for users and definitely doesn't need to hold up this issue. In the meantime, users can always invoke `flush` on their own prior to `executeUpdate` to be safe.  I'll get this approved.

gavinking:
https://github.com/jakartaee/persistence/issues/775

PR #777, opened 2025-11-24
Ensure getResultCount is appropriate for pagination

njr-11:
Query.getResultCount() is intended for pagination of large amounts of data, so it needs to be clear that an implementation shouldn't fetch all of the data and invoke `.size()` on the resulting `List`, which would cap the results reported at `Integer.MAX_VALUE` even though `getResultCount()` returns `long`.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Query.java
+++ b/api/src/main/java/jakarta/persistence/Query.java
@@ -68,15 +68,16 @@ public interface Query {
 
     /**
      * <p>Determine the maximum number of results that could in
-     * principle be returned by the query with no
+     * principle be returned by the query if no
      * {@linkplain #getFirstResult() offset} or
-     * {@linkplain #getMaxResults() limit} applied to the query.</p>
+     * {@linkplain #getMaxResults() limit} were applied.</p>
      *
      * <p>The {@code getResultCount} method should not cause query
      * results to be fetched from the database.</p>
      *
      * @return the maximum number of results that could in principle
-     *         be returned by the query without any offset and limit.
+     *         be returned by the query if no offset or limit were
+     *         applied
      * @throws IllegalStateException if called for a Jakarta
      *         Persistence query language UPDATE or DELETE statement
      * @throws QueryTimeoutException if the query execution exceeds


gavinking:
So, what I meant by my comment on the other issue was I thought it was a *good* thing that there was a trivial, correct, but "bad" implementation of `getResultCount()`, since it meant that this feature is not a barrier to an implementation being able to pass the TCK.

I've never considered that performance considerations, or the ability to handle large quantities of data, were things that the spec itself should have any opinion on. Implementations should have complete freedom when it comes to this sort of "quality" concern.

gavinking:
On the other hand, you do have a very good point that 

> the size of the result list that would be returned

makes it sound as if `Integer.MAX_VALUE` is an upper bound on the result.

We should definitely find a better way to say that.

How about: 

> the maximum number of results that could in principle be returned by the query

or something like that?

njr-11:
> How about:
> 
> > the maximum number of results that could in principle be returned by the query

That's a nice way to word it.  I updated the PR to use that and removed the language about Integer.MAX_VALUE.

Regarding the other point, I would argue that the ability to handle large quantities of data is a matter of capability, not only performance.  Specifications should address capability and the integration with other specifications that is needed to achieve it.  But I'll get rid of the detail that talks specifically about pagination and soften the other statement by switching it to "should".  That will encourage behavior that works well with pagination without completely ruling out implementations that want to retrieve every matching record from the database.

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -67,11 +67,16 @@ public interface Query {
...
 
     /**
-     * Determine the size of the query result list that would be
-     * returned by calling {@link #getResultList()} with no
+     * <p>Determine the maximum number of results that could in
+     * principle be returned by the query with no
      * {@linkplain #getFirstResult() offset} or
-     * {@linkplain #getMaxResults() limit} applied to the query.
-     * @return the size of the list that would be returned
+     * {@linkplain #getMaxResults() limit} applied to the query.</p>
```
gavinking: How about:

> Determine the maximum number of results that could in principle be returned by the query if no offset or limit were applied.

njr-11: That sounds good.  It is updated now.

Issue #780, opened 2025-11-26
annotations for static queries usable in Jakarta Data


gavinking:
Since:

1. #473 has run aground, and
2. it would be nice to have a similar capability to the one proposed in #473 for use with Jakarta Data,

we could consider adding some kind of replacement/alternative to `@NamedQuery`, with the following differences:

- The annotation could be placed at the method level
- The `name` would be optional
- There would be a full range of options like `CacheXxxxMode`s, timeouts, etc.

That is, Jakarta Persistence would define its own "query annotation" in the sense of the Jakarta Data spec.

This is straightforward to add; the main barrier is that I have no idea what a reasonable name for this annotation would be. We can't call it `@JPQL` or `@JpqlQuery` or whatever because the query language isn't called that anymore. We can't call it `@NamedQuery` for multiple reasons. And we can't call it just `@Query`, also for multiple reasons. Nor does `@EntityQuery` work, since such queries are not required to return entities. I suppose `@PersistenceQuery` is available, but that's awful. The best option I've thought of so for is `@QueryWithOptions` or `@QueryWithHints` but ugghh.

Can anyone think of a reasonable name?

Now, an alternative solution, which works for Jakarta Data, but doesn't really help solve the need in #473 is that we could introduce a `@QueryOptions` or `@QueryHints` annotation that's meant to be used together with Jakarta Data's `@Query`. So you would write:

```java
@Query("from Book")
@QueryOptions(cacheStoreMode = BYPASS)
List<Book> allBooks()
```

The problem is that this annotation is completely useless in Jakarta Persistence. It's only useful in Jakarta Data, which makes it an uncomfortable thing to have in this spec.

Note that this issue is very similar to #735.

beikov:
How about `@Select` and `@NativeSelect`? As well as `@Modification` and `@NativeModification`? Maybe also something for stored procedures analogous to `@NamedStoredProcedure`?

I feel like the annotations are going to feel a little misplaced anyway, unless we come up with some way to specify that annotations shall be interpreted by an annotation processor or runtime.

gavinking:
That is a great idea @beikov! Thanks

gavinking:
So, here's an idea based on Christian's comments. 

Let me know what y'all think of this:

```java
/**
 * Specifies a query expressed in the Jakarta Persistence Query Language.
 * This annotation may be applied to any abstract method of any class or
 * interface belonging to the persistence unit, or to a method of a
[...]

Issue #118, opened 2015-09-29
Allow simplified form for @NamedEntityGraph


lukasj:
Spin off from #98:

as of 2.1 definition of @NamedEntityGraph can be too verbose, see ie [https://github.com/javaee-samples/javaee7-samples/blob/master/jpa/entitygraph/src/main/java/org/javaee7/jpa/entitygraph/Movie.java](https://github.com/javaee-samples/javaee7-samples/blob/master/jpa/entitygraph/src/main/java/org/javaee7/jpa/entitygraph/Movie.java)

It would be good to allow simplified form, ie:

```
@NamedEntityGraph(
    attributes = {"id", "name", "description", "subgraph.property", "anotherSubgraph.property"}
)
```

gavinking:
`@NamedEntityGraph` as it exists today is IMO not usable at all.

I like the concept of the `@NamedEntityGraphMember` annotation, which is quite similar to Hibernate's `@FetchProfileOverride` described here:

<https://docs.jboss.org/hibernate/orm/6.3/introduction/html_single/Hibernate_Introduction.html#fetch-profiles>

A different option is an embedded minilanguage, something @sebersole has experimented with, which could look like:

    title, author(firstName, lastName), publicationDate, publisher(name)

and could easily go into an annotation:

```java
@NamedEntityGraph(graph="title, author(firstName, lastName), publicationDate, publisher(name)")
```

This is an issue we should explore further when we have time.

AleksNo:
Hi.

True, @NamedEntityGraph is almost unusable. It is ugly and it becomes very messy very quickly. Especially if you want more than one @NamedEntityGraph on an Entity. I use dynamic EntityGraphs all the time because of it.



Cheers

sebersole:
The "graph language" actually grew initially out of the desire to make building dynamic graphs more concise fwiw.  In Hibernate you can say

```
GraphParser.parse( "title, author(firstName, lastName), publicationDate, publisher(name)" )
```

It would be nice to have that option for named graphs as well.

gavinking:
Please take a look at the strawman proposal in #670, and let me know what you think.

gavinking:
Done, after ten years.

PR #786, opened 2025-11-27
add new annotations for declaring static queries

gavinking:
see #780

This is a rough draft for feedback.

The main value in adding this stuff is in making Jakarta Data query methods much more expressive when the repository implementation is backed by Jakarta Persistence.

### DIFF
--- a/api/src/main/java/jakarta/persistence/query/NativeWriteQuery.java
+++ b/api/src/main/java/jakarta/persistence/query/NativeWriteQuery.java
@@ -61,7 +61,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 public @interface NativeWriteQuery {
 
     /**
-     * The query string in the Jakarta Persistence Query Language.
+     * The query string in the native SQL dialect of the database.
      */
     String query();
 


gavinking:
WDYT @njr-11 @otaviojava @beikov ?

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
```diff
@@ -0,0 +1,125 @@
...
+ * }
+ *
+ * <p> When this annotation is applied to a method of a class or interface
+ * belonging to the persistence unit, a reference to a query declared using
+ * this annotation may be obtained from the static metamodel class of the
+ * annotated class or interface. In addition, the query is treated as a
+ * named query, where the query name is the name of the annotated method.
+ * {@snippet :
+ * List<Book> books =
+ *         em.createQuery(Library_._findBooksByTitle_)
```
beikov, line 64: What's the deal with the underscore before and after the method name?

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * The query string in the native SQL dialect of the database.
+     */
+    String query();
+
+    /**
+     * A list of tables holding data which affects the result of
+     * the query. If neither {@code tables} nor {@code entities}
+     * is correctly specified, the query might return results
+     * which are stale with respect to modifications made within
```
beikov: Shouldn't that align with the default behavior of executing a native query? Also, wouldn't we need a way to specify this through the `Query` interface then as well, through a hint maybe?

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
```diff
@@ -0,0 +1,125 @@
...
+     * the current persistence context.
+     */
+    String[] tables() default {};
+
+    /**
+     * A list of entity types with state affecting the result of
+     * the query. If neither {@code tables} nor {@code entities}
+     * is correctly specified, the query might return results
+     * which are stale with respect to modifications made within
```
beikov: Same as above, I think we might need a way to configure this directly on `Query` as well, though it seems a bit odd to configure this for non-native queries. Maybe it's time for a `NativeQuery` interface in JPA?

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
gavinking: https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#canonical-metamodel

api/src/main/java/jakarta/persistence/query/ReadQuery.java
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Specifies a query expressed in the Jakarta Persistence Query Language,
+ * usually a {@code SELECT} statement, which is executed by calling
+ * {@link jakarta.persistence.Query#getResultList() getResultList()} or
+ * {@link jakarta.persistence.Query#getSingleResult() getSingleResult()}.
+ *
+ * <p> This annotation may be applied to any abstract method of any class
+ * or interface belonging to the persistence unit, or to a method of a
+ * Jakarta Data repository. The return type of the method must agree with
```
beikov, line 38: It might be necessary to clarify the return types a bit, because obviously `List` doesn't necessarily agrees with the type returned by the query. Also, what about `Stream`? Maybe we should list a couple of supported types and specify that vendors are free to support more?

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
gavinking: > Shouldn't that align with the default behavior of executing a native query?

WDYM?

> wouldn't we need a way to specify this through the `Query` interface then as well

_Need_, no. Would it be nice? Yeah, sure, probably.

Unfortunately we don't have any sort of `NativeQuery` interface, and I guess I would prefer not to pollute the API used to execute JPQL.

> through a hint maybe?

Yeah, that's a good idea.

api/src/main/java/jakarta/persistence/query/ReadQuery.java
```diff
@@ -0,0 +1,116 @@
...
+    String query();
+
+    /**
+     * The lock mode type to use in query execution.
+     * If a {@code lockMode} other than {@link LockModeType#NONE}
+     * is specified, the query must be executed in a transaction
+     * and the persistence context joined to the transaction.
+     * @see jakarta.persistence.Query#setLockMode
+     */
+    LockModeType lockMode() default LockModeType.NONE;
```
beikov, line 99: Probably also need a way to specify a `PessimisticLockScope`? How about also adding entity graphs? Maybe also `firstResult` and `maxResults`?

api/src/main/java/jakarta/persistence/query/ReadQuery.java
gavinking: We could spell it out a bit more, but I think it's implied that "agrees" here would mean:

1. agrees with the type returned by `getResultList()` or `getSingleResult()` when executed via JPA, and
2. agrees with the Jakarta Data spec when executed via Jakarta Data.

But sure, it would be good to say that explicitly.

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
beikov: > > Shouldn't that align with the default behavior of executing a native query?
> 
> WDYM?

I mean [this](https://jakarta.ee/specifications/persistence/3.1/jakarta-persistence-spec-3.1#a4374):

>When queries are executed within a transaction, if FlushModeType.AUTO is set on the Query, TypedQuery, or StoredProcedureQuery object, or if the flush mode setting for the persistence context is AUTO (the default) and a flush mode setting has not been specified for the query object, the persistence provider is responsible for ensuring that all updates to the state of all entities in the persistence context which could potentially affect the result of the query are visible to the processing of the query. The persistence provider implementation may achieve this by flushing those entities to the database or by some other means. If FlushModeType.COMMIT is set, the effect of updates made to entities in the persistence context upon queries is unspecified.

AFAIU, this would imply that we flush everything if no tables are specified.

api/src/main/java/jakarta/persistence/query/ReadQuery.java
gavinking: > Probably also need a way to specify a `PessimisticLockScope`?

Correct, but that belongs in #711, not here.

> How about also adding entity graphs?

Desirable, but I'm waiting to see what happens with #118 before doubling down on the current approach.

> Maybe also `firstResult` and `maxResults`?

This one I had not considered. But not sure I think it's useful.

beikov: The usefulness of anything other than the query string AFAICT depends on Jakarta Data support or the addition of the details to `TypedQueryReference`.

api/src/main/java/jakarta/persistence/query/NativeReadQuery.java
gavinking: Wait, in `AUTO` we default to "always flush" for native queries? For some reason I thought we had the opposite default.

gavinking: Shit you're right. Definitely getting senile.

gavinking: Ooooh I see. Here's the reason (gulp).

```java
 * <li>In JPA-compliant mode, {@link FlushModeType#AUTO} specifies that the
 *     session should be flushed before execution of a native query when the
 *     affected tables are not known.
 * <li>Otherwise, when Hibernate is not operating in JPA-compliant mode,
 *     {@code AUTO} specifies that the session is <em>not</em> flushed before
 *     execution of a native query, unless the affected tables are known and
 *     Hibernate determines that a flush is required.
```

gavinking: I have removed all this functionality. Maybe it's useful, but it doesn't need to be addressed here.

api/src/main/java/jakarta/persistence/query/ReadQuery.java
gavinking: Done.

api/src/main/java/module-info.java
```diff
@@ -14,6 +14,7 @@
 
     requires java.logging;
     requires transitive java.sql;
+    requires jakarta.persistence;
```
beikov: That seems wrong 😅 

```suggestion
```

api/src/main/java/jakarta/persistence/query/ReadQuery.java
beikov: Maybe also allow `Stream` explicitly?

api/src/main/java/module-info.java
gavinking: No idea what happened there. IJ.

api/src/main/java/jakarta/persistence/query/ReadQuery.java
gavinking: Meh. I don't know any use convincing case for `getResultStream()`. I don't know why it was even added.

beikov: Since `getResultStream()` does scrolling in Hibernate ORM, any use case that you'd use `ScrollableResult` with forward only scrolling for is applicable here. Most DBs use a cursor and only fetch and execute one-by-one, which can help if you look for something that you can't express and SQL. Either way, it's there, so I think we should support it.

gavinking: Scratch what I said. Jakarta Data allows `Stream`, so we may as well put it on the list.

api/src/main/java/jakarta/persistence/query/NativeWriteQuery.java
```diff
@@ -0,0 +1,81 @@
...
+ * @since 4.0
+ */
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface NativeWriteQuery {
+
+    /**
+     * The query string in the Jakarta Persistence Query Language.
+     */
+    String query();
```
otaviojava: why not value?

Thus, we can type this way:

```java
@NativeWriteQuery("query")
```

The annotation already is about query,  thus, in the default case, it looks redundant doing this:

```java
@NativeWriteQuery(query = "query")
```

gavinking, line 66: @otaviojava If you have a query written in JPQL, and no other information to supply, you can just use `@Query` from Jakarta Data. 

These annotations are most useful in more advanced cases, when you're typically adding other info. And in those cases, "value" is a pretty bad name.

otaviojava (approved):

gavinking:
So, @njr-11 if you split out the options, you would have something like this for the options:

```java
@Target(METHOD)
@Retention(RUNTIME)
public @interface ReadQueryOptions {

    /**
     * The {@linkplain CacheStoreMode cache store mode} to use.
     * @see jakarta.persistence.Query#setCacheStoreMode
     */
    CacheStoreMode cacheStoreMode() default CacheStoreMode.USE;

    /**
     * The {@linkplain CacheRetrieveMode cache retrieve mode}
     * to use.
     * @see jakarta.persistence.Query#setCacheRetrieveMode
     */
    CacheRetrieveMode cacheRetrieveMode() default CacheRetrieveMode.USE;

    /**
     * A query timeout in milliseconds.
     * By default, there is no timeout.
     * @see jakarta.persistence.Query#setTimeout
     */
    int timeout() default -1;

    /**
     * Query properties and hints.
     * May include vendor-specific query hints.
     * @see jakarta.persistence.Query#setHint(String, Object)
     */
    QueryHint[] hints() default {};
}
```
```java
@Target(METHOD)
@Retention(RUNTIME)
public @interface WriteQueryOptions {

    /**
     * A query timeout in milliseconds.
     * By default, there is no timeout.
     * @see jakarta.persistence.Query#setTimeout
     */
    int timeout() default -1;

    /**
     * Query properties and hints.
     * May include vendor-specific query hints.
     * @see jakarta.persistence.Query#setHint(String, Object)
     */
    QueryHint[] hints() default {};
}
```

But of course these annotations would be pretty useless on their own, you can't even use them with `@NamedQuery` or `@NamedNativeQuery`. So you also need:

```java
@Target(METHOD)
@Retention(RUNTIME)
public @interface ReadQuery {

    /**
     * The query string in the Jakarta Persistence Query Language.
     */
    String query();

    /**
     * The lock mode type to use in query execution.
     * If a {@code lockMode} other than {@link LockModeType#NONE}
     * is specified, the query must be executed in a transaction
     * and the persistence context joined to the transaction.
     * @see jakarta.persistence.Query#setLockMode
     */
    LockModeType lockMode() default LockModeType.NONE;
}
```
```java
@Target(METHOD)
@Retention(RUNTIME)
public @interface WriteQuery {

    /**
     * The query string in the Jakarta Persistence Query Language.
     */
    String query();
}
```
```java
@Target(METHOD)
@Retention(RUNTIME)
public @interface NativeReadQuery {

    /**
     * The query string in the native SQL dialect of the database.
     */
    String query();

    /**
     * The name of a {@link SqlResultSetMapping}, as defined in metadata.
     * The named result set mapping is used to interpret the result set
     * of the native SQL query.
     *
     * <p>Alternatively, the elements {@link #entities}, {@link #classes},
     * and {@link #columns} may be used to specify a result set mapping.
     * These elements may not be used in conjunction with
     * {@code resultSetMapping}.
     */
    String resultSetMapping() default "";

    /**
     * Specifies the result set mapping to entities.
     * May not be used in combination with {@link #resultSetMapping}.
     */
    EntityResult[] entities() default {};

    /**
     * Specifies the result set mapping to constructors.
     * May not be used in combination with {@link #resultSetMapping}.
     */
    ConstructorResult[] classes() default {};

    /**
     * Specifies the result set mapping to scalar values.
     * May not be used in combination with {@link #resultSetMapping}.
     */
    ColumnResult[] columns() default {};
}
```
```java
@Target(METHOD)
@Retention(RUNTIME)
public @interface NativeWriteQuery {

    /**
     * The query string in the native SQL dialect of the database.
     */
    String query();
}
```
And now we've got a whole zoo.

You could in principle drop the distinction:

1.  between `@ReadQuery`/`@WriteQuery` and `@NativeReadQuery`/`@NativeWriteQuery`,  and
2.  between `@ReadQueryOptions` and `@WriteQueryOptions`.

but then we would be right back where we started with the problem of what to call these annotations. `@Query` is right out. So is `@JPQL` or variations thereof. I'm left with options like:

-  `@JakartaQuery` + `@NativeQuery`, or
-  `@StaticQuery` + `@StaticNativeQuery`.

and I'm not a fan of either of those.

gavinking:
Actually I suppose we could also:

1. somehow repurpose `@SqlResultSetMapping`, making the `name` optional, and 
2. find some other way to specify a `LockMode` (a new annotation?).

and then you could indeed get away with two annotations for queries, dropping the read/write distinction, for example:

```java
@StaticQuery("from Book where title like ?1")
@ReadQueryOptions(cacheStoreMode=BYPASS)
@LockMode(OPTIMISTIC)
List<Book> booksByTitle(String title);
```
```java
@StaticNativeQuery("select * from books where title like ?1")
@ReadQueryOptions(cacheStoreMode=BYPASS, timeout=1000)
@SqlResultSetMapping(entities=@EntityResult(entityClass=Book.class, fields=....))
List<Book> booksByTitle(String title);
```

But, damn. This is all starting to get a bit out of hand...

api/src/main/java/jakarta/persistence/query/NativeWriteQuery.java
```diff
@@ -0,0 +1,81 @@
...
+ * must treat this annotation as a Jakarta Data query annotation.
+ *
+ * @since 4.0
+ */
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface NativeWriteQuery {
+
+    /**
+     * The query string in the Jakarta Persistence Query Language.
```
gavinking:
```suggestion
     * The query string in the native SQL dialect of the database.
```

gavinking:
@njr-11 see #788.

njr-11:
> @njr-11 see #788.

I added a few minor comments to 788, but it looks good. I think we should go with that approach.

gavinking:
> I think we should go with that approach.

Then I'm closing this one.

PR #788, opened 2025-11-29
alternative annotations for declaring static queries

gavinking:
Alternative to #786 for #780. For @njr-11.

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -2851,7 +2851,7 @@ canonical static metamodel, as specified in <<canonical-metamodel>>.
 public EntityManager em;
     // ...
 
-    customers = em.createNamedQuery(Customer_._findAllCustomersWithName_)
+    customers = em.createQuery(Customer_._findAllCustomersWithName_)
         .setParameter("custName", "Smith")
         .getResultList();
 ----


api/src/main/java/jakarta/persistence/query/NativeQueryMapping.java
```diff
@@ -0,0 +1,47 @@
...
+ * or the Eclipse Distribution License v. 1.0 which is available at
+ * http://www.eclipse.org/org/documents/edl-v10.php.
+ *
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+//     Petros Splinakis - 2.2
+//     Linda DeMichiel - 2.1
+//     Linda DeMichiel - 2.0
```
njr-11: Update to 2025 and list Gavin as contributor instead of names that were probably copied from another file
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
 *
 * This program and the accompanying materials are made available under the
 * terms of the Eclipse Public License v. 2.0 which is available at
 * http://www.eclipse.org/legal/epl-2.0,
 * or the Eclipse Distribution License v. 1.0 which is available at
 * http://www.eclipse.org/org/documents/edl-v10.php.
 *
 * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
 */

// Contributors:
//     Gavin King - 4.0
```

api/src/main/java/jakarta/persistence/query/ReadQueryOptions.java
```diff
@@ -0,0 +1,70 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface ReadQueryOptions {
```
njr-11, line 54:
```suggestion
/**
 * <p>Configures options that apply to a read operation. This method
 * may annotate a method that is annotated {@link StaticQuery},
 * {@link StaticNativeQuery}, or {@code jakarta.data.repository.Query}
 * that performs a read operation. Alternatively, it may annotate a
 * Jakarta Data repository method that is annotated
 * {@code jakarta.data.repository.Find}.</p>
 */
public @interface ReadQueryOptions {
```

api/src/main/java/jakarta/persistence/query/WriteQueryOptions.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+@Target(METHOD)
+@Retention(RUNTIME)
+public @interface WriteQueryOptions {
```
njr-11, line 52:
```suggestion
/**
 * <p>Configures options that apply to a write operation. This method
 * may annotate a method that is annotated {@link StaticQuery},
 * {@link StaticNativeQuery}, or {@code jakarta.data.repository.Query}
 * that has a {@code value} that supplies an {@code UPDATE} or
 * {@code DELETE} statement. Alternatively, it may annotate a
 * Jakarta Data repository method that is annotated
 * {@code jakarta.data.repository.Delete} and uses the
 * parameter-based automatic query pattern.</p>
 */
public @interface WriteQueryOptions {
```

api/src/main/java/jakarta/persistence/query/StaticQuery.java
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * <p> This annotation may be applied to any abstract method of any class
+ * or interface belonging to the persistence unit, or to a method of a
+ * Jakarta Data repository. The return type of the method must agree with
+ * the type returned by the query.
+ * {@snippet :
+ * interface Library {
+ *     @StaticQuery("from Book where title like :title")
+ *      @ReadQueryOptions(cacheStoreMode = CacheStoreMode.BYPASS)
+ *     List<Book> findBooksByTitle(String title);
```
njr-11, line 46:
```suggestion
 *     @StaticQuery("from Book where title like :title")
 *     @ReadQueryOptions(cacheStoreMode = CacheStoreMode.BYPASS)
 *     List<Book> findBooksByTitle(String title);
```

api/src/main/java/jakarta/persistence/query/StaticQuery.java
```diff
@@ -0,0 +1,86 @@
...
+ * Specifies a query expressed in the Jakarta Persistence Query Language,
+ * usually a {@code SELECT} statement, which is executed by calling
+ * {@link jakarta.persistence.Query#getResultList getResultList()},
+ * {@link jakarta.persistence.Query#getResultList getResultStream()}, or
+ * {@link jakarta.persistence.Query#getSingleResult getSingleResult()}.
+ *
+ * <p> This annotation may be applied to any abstract method of any class
+ * or interface belonging to the persistence unit, or to a method of a
+ * Jakarta Data repository. The return type of the method must agree with
+ * the type returned by the query.
```
njr-11: We could consider omitting the use of `StaticQuery` with Jakarta Data because the behavior is already covered by `jakarta.data.repository.Query`.  If we omit it,

```suggestion
 * or interface belonging to the persistence unit. The return
 * type of the method must agree with the type returned by
 * the query.
```

api/src/main/java/jakarta/persistence/query/StaticQuery.java
```diff
@@ -0,0 +1,86 @@
...
+ * }
+ * {@snippet :
+ * Book book =
+ *         em.createQuery(Library_._getBookWithIsbn_)
+ *           .setParameter("isbn", isbn)
+ *           .getSingleResult();
+ * }
+ *
+ * <p> An implementation of Jakarta Data backed by Jakarta Persistence
+ * must treat this annotation as a Jakarta Data query annotation.
```
njr-11, line 93: Another place if we decide to omit,
```suggestion
```

api/src/main/java/jakarta/persistence/query/StaticQuery.java
```diff
@@ -0,0 +1,86 @@
...
+ * <p> A method return type <em>agrees</em> with the type returned by the
+ * query if either:
+ * <ul>
+ * <li>it is exactly {@code R}, {@link java.util.List List&lt;R&gt;}, or
+ *     {@link java.util.stream.Stream Stream&lt;R&gt;} where the query
+ *     return type is assignable to {@code R}, or
+ * <li>the method is a Jakarta Data repository method, and its return type
+ *     is a legal query method return type for the given query, as specified
+ *     by Jakarta Data.
+ * </ul>
```
njr-11, line 62: If we omit the use of StaticQuery with Jakarta Data in favor of Data's existing '@jakarta.data.repository.Query',
```suggestion
 * query if it is exactly {@code R}, {@link java.util.List List&lt;R&gt;}, or
 * {@link java.util.stream.Stream Stream&lt;R&gt;} where the query
 * return type is assignable to {@code R}.
```

api/src/main/java/jakarta/persistence/query/StaticQuery.java
gavinking: We could in principle omit `@StaticQuery` with Jakarta Data. But we need `@StaticNativeQuery`, so that would no be very symmetric.

gavinking:
I have implemented by latest [proposal](https://github.com/jakartaee/persistence/issues/780#issuecomment-3630953268) for `StaticQueryReference`. Please take a look.

gavinking:
After a bunch of work, I've moved this pull request out of draft mode. I'm ready to move forward with this.

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * A reference to a query declared using a {@link StaticQuery}
+ * or {@link StaticNativeQuery} annotation.
+ *
+ * @param <R> The result type of the query
+ *
+ * @since 4.0
+ */
+public final class StaticQueryReference<R>
```
beikov: Why not make it a record?

gavinking: Because `TypedQueryReference` has getter-style accessors.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
```diff
@@ -41,6 +83,49 @@ public interface TypedQueryReference<R> {
...
+     * array of parameter names} to obtain an assignment
+     * of arguments to parameters.
+     * </ul>
+     *
+     * @see Query#setParameter(int, Object)
+     * @see Query#setParameter(String, Object)
+     *
+     * @since 4.0
+     */
+    Object[] getArguments();
```
beikov: IMO, it would be better to have a `Map<Object, Object>` or `Map<Integer, Object>` along with `Map<String, Object>` to represent the parameter values.
Also note that something very valuable is lost here, which is the static type that the user specifies for a parameter. This type can be crucial for understanding how to bind a `null`, so we should strive to capture the static type of a parameter value if possible.
Maybe with something like 

```java
record QueryParameterValue(
  Object nameOrPosition, 
  Object value,
  java.lang.reflect.Type type) {

  QueryParameterValue {
    if (!(nameOrPosition instanceof Integer) || !(nameOrPosition instanceof String)) {
      throw new IllegalArgumentException( "Invalid name or position given: " + nameOrPosition );
    }
  }
}
```

and then just have a `List<QueryParameterValue> getParameterValues()`

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
beikov: It could have both, those getters and the record component accessors, but I understand that it might seem inconsistent. It seems though, as if `StaticQueryReference` is more like a "helper" for implementors to avoid generating code for user classes that uses vendor specific APIs, rather than being an actual API, so maybe making it a record is fine after all, since it's never going to be used by users directly?
Note that records have performance advantages over classes, which might be non-negligible since this would be created for every query execution.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
gavinking: > IMO, it would be better to have a `Map<Object, Object>` or `Map<Integer, Object>` along with `Map<String, Object>` to represent the parameter values.

I considered that, but ultimately decided that this approach was simpler and easier to deal with.

> This type can be crucial for understanding how to bind a null, so we should strive to capture the static type of a parameter value if possible.

I guess that could in principle be added. But note that JPA's `Query` and `TypedQuery` API's don't have a good way to specify the type of a `null` value, so I guess it can't be *that* crucial most of the time, or we would have heard about it.

And implementations could subclass `StaticQueryReference` (another advantage of it not being a `record`) to hang additional info like this off it if they want to go beyond the spec.

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
gavinking: > It could have both

It could, but that's ugly.

> more like a "helper" for implementors to avoid generating code for user classes that uses vendor specific APIs, rather than being an actual API

That's correct. It could be removed without breaking anything, but I think it's nicer if the metamodel can in principle be generated without dependencies on vendor classes. That's not a *requirement*, but I would like to make it *possible*.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
beikov: >I guess that could in principle be added. But note that JPA's Query and TypedQuery API's don't have a good way to specify the type of a null value, so I guess it can't be that crucial most of the time, or we would have heard about it.

I heard plenty of this, especially when it's about native queries. People have to resort to using `TypedParameterValue` or native Hibernate ORM APIs to get this to work. Note that e.g. Spring Data JPA now wraps values as `TypedParameterValue` automatically to handle this problem, after many people have requested this.

>And implementations could subclass StaticQueryReference (another advantage of it not being a record) to hang additional info like this off it if they want to go beyond the spec.

You marked `StaticQueryReference` as `final` now, so it couldn't be subclassed, but ultimately, implementations can go beyond by having a sub-interface to `TypedQueryReference` and then a record subtype that offers those extensions.

gavinking: I don't think I meant to make it `final`. IntelliJ did that for me when I converted it from a `record` :)

gavinking: I'm not against packaging an array of classes in there for the parameter types. I just thought it was weird when JPA APIs don't give you any way to do anything useful with them. 

It's almost like that's the more serious problem here and we should be adding a new method to `Query`.

beikov: I think it's worth adding `setParameter` variants then that also take a `java.lang.reflect.Type` and/or `Class` then to have that be modeled in JPA directly and get rid of the `TypedParameterValue` overall.

gavinking: I will put something together.

api/src/main/java/jakarta/persistence/SqlResultSetMapping.java
```diff
@@ -20,38 +20,70 @@
...
+ * }
+ * {@snippet :
+ * Query query = entityManager.createNativeQuery(
+ *         """
+ *           SELECT o.id AS order_id,
+ *                  o.total AS order_total,
+ *                  o.item_id AS order_item,
+ *                  i.desc_name AS item_name
+ *           FROM orders o, order_items i
+ *           WHERE order_quantity > 25 AND order_item = i.id
```
njr-11: to match with `o.total AS order_total` above
```suggestion
 *           WHERE order_total > 25 AND order_item = i.id
```

api/src/main/java/jakarta/persistence/SqlResultSetMapping.java
```diff
@@ -20,38 +20,70 @@
...
+ *         """,
+ *         Order_.MAPPING_ORDER_RESULTS
+ * );
+ * }
+ *
+ * <p>In this example, a mapping is specified by annotating a method
+ * which declares a {@link jakarta.persistence.query.StaticNativeQuery},
+ * and the mapping does not need to explicitly specify its name:
+ * {@snippet :
+ * @StaticNativeQuery("SELECT * FROM orders WHERE order_quantity > :?1")
```
njr-11: Another case where rename to `order_total` was missed.
```suggestion
 * @StaticNativeQuery("SELECT * FROM orders WHERE order_total > ?")
```

api/src/main/java/jakarta/persistence/query/StaticNativeQuery.java
```diff
@@ -0,0 +1,104 @@
...
+ * or interface belonging to the persistence unit, or to a method of a
+ * Jakarta Data repository. The return type of the method must agree with
+ * the type returned by the query.
+ * {@snippet :
+ * interface Library {
+ *     @StaticNativeQuery("select * from books where title like ?1")
+ *     @ReadQueryOptions(cacheStoreMode = CacheStoreMode.BYPASS)
+ *     List<Book> findBooksByTitle(String title);
+ *
+ *     @StaticNativeQuery("select * from books where isbn = ?1")
```
njr-11: SQL likely intended for native query rather than JPQL
```suggestion
 *     @StaticNativeQuery("select * from books where title like ?")
 *     @ReadQueryOptions(cacheStoreMode = CacheStoreMode.BYPASS)
 *     List<Book> findBooksByTitle(String title);
 *
 *     @StaticNativeQuery("select * from books where isbn = ?")
```

api/src/main/java/jakarta/persistence/query/StaticNativeQuery.java
```diff
@@ -0,0 +1,104 @@
...
+ * string {@code "."}, and the name of the annotated member, for example,
+ * {@code "Library.getBookWithIsbn"}.
+ *
+ * <p> If a {@link jakarta.persistence.SqlResultSetMapping} exists with
+ * the same name as the query, it is used to map the result set of the
+ * query. If the {@code @SqlResultSetMapping} annotation occurs on the
+ * same method as {@code @StaticNativeQuery} annotation, name defaulting
+ * rules conspire to ensure that no name needs to be explicitly specified.
+ * {@snippet :
+ * @StaticNativeQuery("select * from books where isbn = ?1")
```
njr-11:
```suggestion
 * @StaticNativeQuery("select * from books where isbn = ?")
```

api/src/main/java/jakarta/persistence/SqlResultSetMapping.java
njr-11: missed rename
```suggestion
 * @StaticNativeQuery("SELECT * FROM orders WHERE order_total > ?")
```

api/src/main/java/jakarta/persistence/SqlResultSetMapping.java
njr-11: to match with `o.total AS order_total`
```suggestion
 *           WHERE order_total > 25 AND order_item = i.id
```

njr-11 (commented):
Added a few minor comments

gavinking:
> Added a few minor comments

Thanks!

gavinking:
Any more feedback on this? @sebersole?

sebersole:
I really like the split for read- and write-options compared to the previous one.

+1 from me!

api/src/main/java/jakarta/persistence/query/StaticQuery.java
```diff
@@ -0,0 +1,96 @@
...
+ * {@code "Library.findBooksByTitle"}.
+ *
+ * <p> An implementation of Jakarta Data backed by Jakarta Persistence
+ * must treat this annotation as a Jakarta Data query annotation.
+ *
+ * <p>The parameter types of the annotated method must be basic types.
+ * If this annotation occurs on a generic method, on a method with a
+ * parameter type which is not a basic type, or on a method with a
+ * return type or parameter type involving a type variable, the behavior
+ * is undefined.
```
gavinking: I have added this copy pasta. Does it LGTY, @njr-11 ?

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
```diff
@@ -0,0 +1,126 @@
...
+            Map<String, Object> hints,
+            FindOption... options) {
+        this.resultType = resultType;
+        this.name = name;
+        this.parameterTypes = parameterTypes;
+        this.parameterNames = parameterNames;
+        this.arguments = arguments;
+        this.options = List.of(options);
+        this.hints = hints;
+    }
```
gavinking, line 67: I slightly reworked the design of `StaticQueryReference` after the discussion with @beikov.

- now includes parameter types
- using `List` instead of arrays so we can have immutability

api/src/main/java/jakarta/persistence/query/StaticQuery.java
njr-11: Yes, that looks good.  This will prevent the use of special parameters that might otherwise require modifying the query.  We definitely do not want to modify native queries.  For non-native queries, if the user wants PageRequest or Order parameters, they can just use Jakarta Data's Query annotation instead.

api/src/main/java/jakarta/persistence/query/ReadQueryOptions.java
```diff
@@ -27,25 +27,43 @@
...
+     * cache store mode of the persistence context.
      * @see jakarta.persistence.Query#setCacheStoreMode
      */
     CacheStoreMode cacheStoreMode() default CacheStoreMode.USE;
 
     /**
      * The {@linkplain CacheRetrieveMode cache retrieve mode}
-     * to use.
+     * to use. The presence of this annotation overrides the
+     * default cache store mode of the persistence context.
```
njr-11:
```suggestion
     * default cache retrieve mode of the persistence context.
```

api/src/main/java/jakarta/persistence/query/StaticNativeQuery.java
```diff
@@ -22,24 +22,30 @@
...
- * usually a SQL {@code SELECT} statement, which is executed by calling
- * {@link jakarta.persistence.Query#getResultList getResultList()},
- * {@link jakarta.persistence.Query#getResultStream getResultStream()}, or
- * {@link jakarta.persistence.Query#getSingleResult getSingleResult()}.
+ * Specifies a query expressed in the native SQL dialect of the database.
+ * This annotation may be applied to:
+ * <ul>
+ * <li>any method of any class or interface belonging to the persistence
+ *     unit, or
+ * <li>a query method or default method of a Jakarta Data repository with
```
njr-11: Mentioning default method here will cause confusion to users, who will be confused about why they would want to write a default method implementation and also add annotation to the method that defines the behavior.  At best, this will seem redundant (if the two match), and at worst result in uncertain behavior (if they don't match).  Leaving this out of the Javadoc will not stop Data providers who know what they ware doing from being able to implement the behavior of the annotation by turning the method into a default method that still has the annotation.

```suggestion
 * <li>a query method of a Jakarta Data repository with
```

api/src/main/java/jakarta/persistence/query/StaticQuery.java
```diff
@@ -22,24 +22,32 @@
...
- * usually a {@code SELECT} statement, which is executed by calling
- * {@link jakarta.persistence.Query#getResultList getResultList()},
- * {@link jakarta.persistence.Query#getResultList getResultStream()}, or
- * {@link jakarta.persistence.Query#getSingleResult getSingleResult()}.
+ * Specifies a query expressed in the Jakarta Persistence Query Language.
+ * This annotation may be applied to:
+ * <ul>
+ * <li>any method of any class or interface belonging to the persistence
+ *     unit, or
+ * <li>a query method or default method of a Jakarta Data repository with
```
njr-11:
```suggestion
 * <li>a query method of a Jakarta Data repository with
```

api/src/main/java/jakarta/persistence/query/StaticNativeQuery.java
gavinking: I'll take it out for now, but notice that this isn't really imposing any sort of requirement on Data. It's the Persistence metamodel which is going to get the TypedQueryRefence method generated into it, and the Data spec doesn't actually really need to know anything at all about that.

gavinking:
Alright, so this has been through quite a few design iterations now, and lot of people have looked at it.

I had not originally considered this as belonging in M1, but I think it's turned out really nicely and is in more than a good enough state to go in. (Of course that doesn't mean we can't still make changes.)

So I think I'm just about ready to pull through trigger on this one.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
```diff
@@ -41,6 +84,73 @@ public interface TypedQueryReference<R> {
...
+     */
+    List<Class<?>> getParameterTypes();
+
+    /**
+     * The names assigned to the supplied
+     * {@linkplain #getArguments arguments} to query
+     * parameters, or {@code null} if no arguments were
+     * supplied. If the query has named parameters, these
+     * are interpreted as the parameter names. Otherwise,
+     * if the query has positional parameters, they are
```
beikov: It feels a bit arbitrary to me to exclude positional parameters here. Why not have a the `QueryParameterValue` record [I proposed before](https://github.com/jakartaee/persistence/pull/788#discussion_r2603213009) and expose just a single list?
Also, it seems unfortunate to return `null` if there are no parameters. An empty list would be better IMO.

gavinking: > It feels a bit arbitrary to me to exclude positional parameters here.

I don't understand. I'm not excluding anything. I'm passing all the useful information that's available to the processor.

> Why not have a the `QueryParameterValue` and expose just a single list?

Because that just seems unnecessarily complicated for what we're trying to achieve here. This already does everything we need without introducing additional complications. 

And it makes things easier for the annotation processor. With this approach, the processor doesn't even need to parse the query string to determine what kind of the parameters the query has. (Though of course our implementation will indeed do that.)

> Also, it seems unfortunate to return null if there are no parameters. An empty list would be better IMO.

`null` means something different to an empty parameter list. If the `TypedQueryReference` represents a `@NamedQuery` annotation on a type, then you get `null`. If it represents a `@StaticQuery` annotation on a method, then you get an empty list.

gavinking: >  If the TypedQueryReference represents a @NamedQuery annotation on a type, then you get null. If it represents a @StaticQuery annotation on a method, then you get an empty list.

I have clarified this point in the javadoc.

beikov: > > It feels a bit arbitrary to me to exclude positional parameters here.
> 
> I don't understand. I'm not excluding anything. I'm passing all the useful information that's available to the processor.

Ok, so maybe I misread the paragraph then. So the idea is that you're always binding the method parameter names and the persistence provider can choose to interpret that as query parameter name (if named parameters are used) or ignore it (if positional parameters are used).

> > Why not have a the `QueryParameterValue` and expose just a single list?
> 
> Because that just seems unnecessarily complicated for what we're trying to achieve here. This already does everything we need without introducing additional complications.
> 
> And it makes things easier for the annotation processor. With this approach, the processor doesn't even need to parse the query string to determine what kind of the parameters the query has. (Though of course our implementation will indeed do that.)

With what I understood now, the annotation parser wouldn't need to parse the query and we could just use this simple version, binding the method parameter name:

```java
record QueryParameterValue(
  String parameterName, 
  Object value,
  java.lang.reflect.Type type) {}
```

Anyway, since we can store the method parameter names and types as static finals and have to instantiate only a single list also with your approach, I'm fine with either approach. I guess my confusion simply was with where the values for `getParameterNames()` would come from.

> > Also, it seems unfortunate to return null if there are no parameters. An empty list would be better IMO.
> 
> `null` means something different to an empty parameter list. If the `TypedQueryReference` represents a `@NamedQuery` annotation on a type, then you get `null`. If it represents a `@StaticQuery` annotation on a method, then you get an empty list.

Thanks for clarifying. I forgot that `TypedQueryReference` also has other uses.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
```diff
@@ -41,6 +84,73 @@ public interface TypedQueryReference<R> {
...
+     * {@linkplain #getArguments arguments} to query
+     * parameters, or {@code null} if no arguments were
+     * supplied.
+     * <p>
+     * Any mutation of the returned list results in an
+     * {@link UnsupportedOperationException}.
+     *
+     * @since 4.0
+     */
+    List<Class<?>> getParameterTypes();
```
beikov, line 134: Why not capture the full type? An implementation can choose to use raw types if they want, but I'd rather not throw away that information if possible. Think about e.g. `List<String>` vs `List<Integer>` which have different SQL types.

```suggestion
    List<java.lang.reflect.Type> getParameterTypes();
```

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @Override
+    public String getName() {
+        return "";
```
beikov: Is returning an empty string done on purpose? Also, there is no way to access the `name` field as of now.

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
```diff
@@ -0,0 +1,126 @@
...
+    private final List<FindOption> options;
+    private final Map<String, Object> hints;
+
+    public StaticQueryReference(
+            Class<R> resultType, String name,
+            List<Class<?>> parameterTypes,
+            List<String> parameterNames,
+            List<Object> arguments,
+            Map<String, Object> hints,
+            FindOption... options) {
```
beikov, line 57: Maybe I'm missing something, but it seems easier to me to just make it a list up here.

```suggestion
            List<FindOption> options) {
```

gavinking: I'm thinking about what's nicer to read (and generate) in the generated code.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
gavinking: That's going to be *waaaaay* harder for the annotation processor.

beikov: Hmm, AFAICT it would simply be this: `TheClass.class.getMethod("methodName", arg1RawType.class, arg2RawType.class).getGenericParameterTypes()`

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
beikov: Gotcha

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
gavinking: Nope that is a mistake, thanks.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
gavinking: If we were going to do something like that, we could just expose the whole `Member` object and be done. But the purpose of this stuff isn't to allow runtime reflection on repository methods. Perhaps one day we will find a use for that, and we can prototype it in our implementation and maybe eventually add it to the spec. But today we don't have such a need for the problems we're trying to solve here. And I think it would be very premature to add such a thing for M1.

gavinking: Note that annotations are even more useful than full generic parameter types. But we don't need them to either.

gavinking: Actually if we gave them the class object where the annotation occurs, that would be enough to look up the member. I suppose we could go that far.

gavinking: Hold on I will push something in 20 mins.

beikov: Exposing the `Member` or anything else would imply that a type containing `@StaticQuery` has to be fully available for reflection at _runtime_, whereas exposing just the `java.lang.reflect.Type` of the method parameters means that only the parameter types have to be available at _runtime_, which they naturally are anyway.

>Note that annotations are even more useful than full generic parameter types. But we don't need them to either.

Unless we specify parameter annotations in JPA, I don't see a need. We could easily add `Annotation[][] getParameterAnnotations()` later if we need, but how would those be interpreted?

api/src/main/java/jakarta/persistence/TypedQueryReference.java
gavinking: I mean, sure, we can introduce a single `record` type like that. I thought you were asking for subtypes representing named parameters vs positional parameters. I guess it makes things slightly easier on the client, since Java doesn't have a god damned `zip()` function. The thing is, I don't know where to put that. An inner type belonging to `TypedQueryReference`? Don't really love that at all. Or do we demote all this stuff about arguments down to `StaticQueryReference`? That might make the most sense, since that is all starting to get more complicated than I hoped.

beikov: Having separate lists is totally fine. We can store this static information in the generated class as static final fields.

api/src/main/java/jakarta/persistence/query/StaticQueryReference.java
gavinking: fixed.

api/src/main/java/jakarta/persistence/TypedQueryReference.java
gavinking: > Exposing the Member or anything else would imply that a type containing @StaticQuery has to be fully available for reflection at runtime, whereas exposing just the java.lang.reflect.Type of the method parameters means that only the parameter types have to be available at runtime, which they naturally are anyway.

So whether this is an actual advantage is *entirely* speculative, because we don't know how this would actually be used. Which is precisely why I don't like designing things without an actual use case and a proof of concept solution.

> Unless we specify parameter annotations in JPA, I don't see a need. 

I mean, Data now puts quite a lot of info in parameter annotations, and perhaps a Persistence implementation could make use of that to do interesting runtime validation/optimization/whatever. I don't know. Which is why I would prefer to leave stuff out for now.

That said, I added a `getMember()` to `StaticQueryReference` because it was extremely easy to do. If you like it, I'll leave it in. Otherwise I would be just as happy to leave it out at least for now.

beikov: >That said, I added a getMember() to StaticQueryReference because it was extremely easy to do. If you like it, I'll leave it in. Otherwise I would be just as happy to leave it out at least for now.

I think `getMember()` is not useful. At best, it would have to be `getMethod()`, since `@StaticQuery` has `@Target(METHOD)`, but I'm concerned about reflection usage for e.g. native image generation, which is why I'd prefer to store only the bits we actually make use of.

>So whether this is an actual advantage is entirely speculative, because we don't know how this would actually be used. Which is precisely why I don't like designing things without an actual use case and a proof of concept solution.

Imagine a user writes a native query e.g. `select * from tbl t where t.col = coalesce(:array, t.col)` and wants to use `List<X>` instead of `X[]` as argument type. No problem with the `setParameter` API, because the user can pass the `java.lang.reflect.Type` to `BasicTypeRegistry#getRegisteredType` to get a `jakarta.persistence.metamodel.Type`. With the `@StaticQuery`, it wouldn't work though, because we'd only use `Class` instead of `java.lang.reflect.Type`.

Now, you could simply say that this would just be a limitation of this API and users will have to use `X[]` instead of `List<X>`, but it's IMO a waste to not make the API allow exposing such type information.

gavinking: > I think getMember() is not useful. 

Then I will probably remove it.

> At best, it would have to be `getMethod()`, since `@StaticQuery has @Target(METHOD)`

Again, I'm anticipating reasonable (ab)usages of this API which go somewhat beyond (but not too far beyond) what is specified here. I don't want to prevent this stuff being used with fields. `Method` is a subclass of `Member`, and it's easy to downcast it.

> but I'm concerned about reflection usage for e.g. native image generation, which is why I'd prefer to store only the bits we actually make use of.

I mean, this would be of concern if this `getMember()` method were needed for the specified functionality or if native image were the main target platform of the Persistence spec. Neither of those things are true.

> Imagine a user writes a native query e.g. `select * from tbl t where t.col = coalesce(:array, t.col)` and wants to use `List<X> `instead of `X[]` as argument type. No problem with the `setParameter` API, because the user can pass the `java.lang.reflect.Type` to `BasicTypeRegistry#getRegisteredType` to get a `jakarta.persistence.metamodel.Type`. 

Yes problem, because what you just described is functionality of Hibernate, not of Jakarta Persistence. JPA doesn't even require named parameter support -- let alone collection-valued parameters -- with native SQL. That code is not required to work in a different provider. And, indeed, there are significant downsides to the SQL preprocessing that Hibernate does.

Anyway, we are getting way off into "boil the ocean" territory here.

gavinking:
I have made some further changes to the spec sections 5.1.1 and 3.11.9 for readability and a little bit of completeness. There only impact on semantics is to say that static metamodel generation works for any class annotated `@NamedQuery` or `@NamedNativeQuery`, which was previously unclear.

I've also updated the Javadoc for `@StaticMetamodel` to make some of this info more accessible.

gavinking:
I'm going too merge this now, because it's already really big, and I didn't want it to be big. I definitely don't want it to get any bigger in one huge PR.

Further changes can be done in followup issues. I already have an issue I'm going to open about a possible change to the API for options annotations, but because everyone was very happy with the API proposed here, it's better to have that discussion separately.

PR #791, opened 2025-11-30
remove deprecated support for Character[] and Byte[] as basic types

### LINKED ISSUE #790: remove required support for Character[] and Byte[]
In #549 we deprecated the use of `Character[]` and `Byte[]` as basic types in JPA 3.2 because of real problems with the handling of null elements.

We should completely remove this in 4.0.

[Of course, individual providers can continue to provide support for these types as an optional feature; that doesn't need to be mentioned in the spec.]

gavinking:
fixes #790


gavinking:
Thanks, @scottmarlow !

PR #794, opened 2025-12-02
clarify usage of @OrderColumn and @MapKeyColumn

gavinking:
I've realized that there's huge confusion here, because the Javadoc for these annotations fails to mention an important thing.

### DIFF
--- a/api/src/main/java/jakarta/persistence/MapKeyJoinColumn.java
+++ b/api/src/main/java/jakarta/persistence/MapKeyJoinColumn.java
@@ -90,11 +90,10 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * }
  * }
  *
- * <p>The {@code MapKeyJoinColumn} annotation is specified on the owning side
- * of a {@link OneToMany} or {@link ManyToMany} relationship or on an
- * element collection. The {@code MapKeyJoinColumn} annotation is specified
- * on the side of the relationship that references the map. The map key
- * is not visible as part of the state of the entity or embeddable class.
+ * <p>This annotation is specified on the owning side of a {@link OneToMany}
+ * or {@link ManyToMany} relationship or on an element collection&mdash;and
+ * always on the side of the relationship that references the map. The map
+ * key is not visible as part of the state of the entity or embeddable class.
  *
  * <p>This annotation is used with owned collections.
  *
--- a/api/src/main/java/jakarta/persistence/OrderColumn.java
+++ b/api/src/main/java/jakarta/persistence/OrderColumn.java
@@ -30,16 +30,17 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * database to reflect any insertion, deletion, or reordering
  * affecting the list.
  *
- * <p>The {@code OrderColumn} annotation is specified on the owning side
- * of a {@link OneToMany} or {@link ManyToMany} relationship or on an
- * element collection. The {@code OrderColumn} annotation is specified
- * on the side of the relationship that references the collection that
- * is ordered. The order column is not visible as part of the state of
- * the entity or embeddable class.
+ * <p>This annotation is specified on the owning side of a
+ * {@link OneToMany} or {@link ManyToMany} relationship or on an
+ * element collection&mdash;and always on the side of the relationship
+ * that references the collection that is ordered. The order column is
+ * not visible as part of the state of the entity or embeddable class.
  *
  * <p>This annotation is used with owned collections. For unowned
  * collections, that is, for association mappings which specify
  * {@code mappedBy}, the {@link OrderBy} annotation should be used.
+ * Use of the {@code OrderColumn} annotation on the unowned side of a
+ * relationship is not portable between persistence providers.
  *
  * <p>The {@link OrderBy} annotation should be used for ordering that
  * is visible as persistent state and maintained by the application.
--- a/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
+++ b/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
@@ -3323,11 +3323,13 @@ type. If the `name` element is not specified, it defaults to the
 concatenation of the following: the name of the referencing relationship
 field or property; "`_`"; "`KEY`".
 
-The `MapKeyColumn` annotation is used with owned collections. It is specified
-on the owning side of a one-to-many or many-to-many relationship or on an
-element collection. The `MapKeyColumn` annotation is specified on the side of
-the relationship that references the map. The map key is not visible as part
-of the state of the entity or embeddable class.
+`MapKeyColumn` is used with owned collections. It is specified on the
+owning side of a one-to-many or many-to-many relationship or on an
+element collection--and always on the side of the relationship that
+references the map. The map key is not visible as part of the state of
+the entity or embeddable class. Use of the `MapKeyColumn` annotation on
+the unowned side of a relationship is not portable between persistence
+providers.
 
 [source,java]
 ----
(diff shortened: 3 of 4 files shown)


api/src/main/java/jakarta/persistence/OrderColumn.java
```diff
@@ -30,12 +30,16 @@
  * database to reflect any insertion, deletion, or reordering
  * affecting the list.
  *
- * <p>The {@code OrderColumn} annotation is specified on a
- * {@link OneToMany} or {@link ManyToMany} relationship or on an
+ * <p>The {@code OrderColumn} annotation is specified on the owning side
+ * of a {@link OneToMany} or {@link ManyToMany} relationship or on an
```
gavinking: The only really semantically significant change here is to make explicit that these annotations are supposed to be used on *owned* collections.

gavinking:
Damn, the TCK actually has tests which test the use of these annotations on unowned collections, where the order or key column doesn't exist at all in the other entity, creating a requirement for "partially owned" collections. I really think this is an incredibly bad feature, typically resulting in inefficient SQL, and that we should strongly consider deprecating it.

Thoughts, everyone?

gavinking:
I mean, the spec also says:

> The owning side of a relationship determines the updates to the relationship in the database.

and:

> It is the developer's responsibility to keep the in-memory references held on the owning side and those held on the inverse side consistent with each other when they change. ... It is particularly important to ensure that changes to the inverse side of a relationship result in appropriate updates on the owning side, so as to ensure the changes are not lost when they are synchronized to the database.

This notion of a "partially owned" relationship just doesn't exist in the text of the specification, as far as I can tell. It's a pure invention of the TCK.

beikov:
The `@OrderColumn` probably doesn't make a lot of sense on the inverse/unowned side, since the order will be determined based on the owning side, so +1 for that clarification.
The `@MapKeyColumn` might make sense on the inverse/unowned side though, so -1 on changing that text.

gavinking:
> The `@MapKeyColumn` might make sense on the inverse/unowned side though

It only makes sense if the child entity has a field holding the value of the map key. In which case, you should use `@MapKey` to do the mapping, not `@MapKeyColumn`.

beikov:
It seems to me that with a `@ManyToMany` where the map key column is placed on the join table, it makes sense to map the inverse side with

```
@MapKeyColumn(name = "map_key")
@ManyToMany(mappedBy = "theMap")
```

gavinking:
> It seems to me that with a `@ManyToMany` where the map key column is placed on the join table

I mean, OK, in principle yes ... but that "map key column" would then wind up being a unique key of the join table. I doubt any JPA implementation handles this mapping elegantly (Hibernate doesn't).

If some JPA implementation wants to support such mapping, that's fine, it can. But I don't think the spec should require support for that.

(I'm assuming you mean that it's a bidirectional many-to-many with a `Map` on both sides.)

beikov:
>I mean, OK, in principle yes ... but that "map key column" would then wind up being a unique key of the join table. I doubt any JPA implementation handles this mapping elegantly (Hibernate doesn't).

Sure, it would form a unique constraint along with the foreign key columns pointing to the owning side. If Hibernate doesn't behave correctly here, that would be a bug that we should simply fix IMO.

>If some JPA implementation wants to support such mapping, that's fine, it can. But I don't think the spec should require support for that.

Right, so maybe change the language to say that support for this kind of mapping is non-portable or something like that.

>(I'm assuming you mean that it's a bidirectional many-to-many with a Map on both sides.)

Obviously `@MapKeyColumn` implies usage of `Map`, so yeah. I'm not sure if it is possible to map the owning side as `Map` and the unowned side with `@ManyToMany(mappedBy = "...")` as `Set` or `Collection`/`List`.

gavinking:
>  I'm not sure if it is possible to map the owning side as `Map` and the unowned side with `@ManyToMany(mappedBy = "...")` as `Set` or `Collection`/`List`.

That is possible, and completely normal.

What is *extremely* debatable is whether you should be able to have things like a `Set` on the owning side, but a `MapKeyColumn` on the unowned side. I claim that all **mapping** information should be defined on the side which is pointed to by the `mappedBy` element. There should be no mixing in of mapping info on a field which says it is `mappedBy` something else.

api/src/main/java/jakarta/persistence/MapKeyJoinColumn.java
```diff
@@ -90,12 +90,20 @@
  * }
  * }
  *
+ * <p>The {@code MapKeyColumn} annotation is specified on the owning side
+ * of a {@link OneToMany} or {@link ManyToMany} relationship or on an
+ * element collection. The {@code MapKeyColumn} annotation is specified
```
njr-11:
```suggestion
 * <p>The {@code MapKeyJoinColumn} annotation is specified on the owning side
 * of a {@link OneToMany} or {@link ManyToMany} relationship or on an
 * element collection. The {@code MapKeyJoinColumn} annotation is specified
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -3512,6 +3518,12 @@ specify a mapping to an entity that is a map key. The map key join
 column is in the collection table, join table, or table of the target
 entity that is used to represent the map.
 
+The `MapKeyJoinColumn` is used with owned collections. It is specified
+on the owning side of a one-to-many or many-to-many relationship or on
+an element collection. The `MapKeyColumn` annotation is specified on the
```
njr-11:
```suggestion
an element collection. The `MapKeyJoinColumn` annotation is specified on the
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -4444,18 +4456,18 @@ a column that is used to maintain the persistent order of a list. The
...
-part of the state of the entity or embeddable
-class.footnote:[The `OrderBy`
-annotation should be used for ordering that is visible as persistent
-state and maintained by the application.]
-
-The `OrderBy` annotation is not used when
-`OrderColumn` is specified.
+insertion, deletion, or reordering affecting the list.
+
+The `OrderColumn` is used with owned collections. It annotation may be
```
njr-11:
```suggestion
The `OrderColumn` is used with owned collections. This annotation may be
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -4444,18 +4456,18 @@ a column that is used to maintain the persistent order of a list. The
...
+specified on an owned one-to-many or many-to-many relationship or on
+an element collection. The `OrderColumn` annotation is specified on the
+side of the relationship that references the collection that ordered.
```
njr-11:
```suggestion
side of the relationship that references the collection that is ordered.
```

gavinking:
Thanks @njr-11, I applied all your suggestions. Not going to merge this for M1, because I still want to get @sebersole's feedback on it, especially since @beikov was a bit skeptical of this one.

sebersole:
> This notion of a <XYZ> just doesn't exist in the text of the specification, as far as I can tell. It's a pure invention of the TCK.

Shocking

sebersole:
How about we carve out an exception in the clarification for the many-to-many case?
Yes, that would mean that we'd have to "fix" this in Hibernate, but its more of question of correctness I think.

gavinking:
> that would mean that we'd have to "fix" this in Hibernate

Hibernate actually *does* allow that IIRC, but it uses two SQL statements to set up the association.

gavinking:
This is not so much about what is possible, but about what the spec *requires*.

sebersole:
> > that would mean that we'd have to "fix" this in Hibernate
> 
> Hibernate actually _does_ allow that IIRC, but it uses two SQL statements to set up the association.

Hence "fix" in quotes.  "Improve" if you prefer ;)

beikov:
Just for your information, there seem to be people out there that really use `@OrderColumn` on unowned collections: https://hibernate.atlassian.net/browse/HHH-11594

gavinking:
I have weakened the language slightly, so that it's clear that an implementation _might_ allow these annotations on the unowned side, and what we're saying is just that it's unportable.

gavinking:
I'm going to pull the trigger on this one, because I believe it correctly states the original intent of these annotations.

gavinking:
> I believe it correctly states the original intent of these annotations

And because we have not defined anywhere in the spec how a "partially owned" collection mapping is supposed to behave. (And such a definition is completely nontrivial.)

Issue #724, opened 2025-05-30
version of find() which never returns null


gavinking:
Eons ago, before anyone knew how to design APIs in Java, we decided that `Session.find()` should just return `null` if the entity was not found. This was consistent with other Java APIs, like, for example, `Map.get()`.

I no longer think that this was a great call, and I think it would be nice to have an alternative which throws a `NoResultException` when there's no entity to return. The tricky part of this is: what to call the new operation. Possibilities include: `get()`, `load()`, `fetch()`, or `retrieve()`.

[This was already mentioned in the discussion on #479, and I was surprised to realize it didn't have its own issue yet.]

PR #795, opened 2025-12-02
proposal for a stateless EntityAgent

gavinking:
Redo of #675, to deal with conflicts.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityHandler.java
+++ b/api/src/main/java/jakarta/persistence/EntityHandler.java
@@ -39,13 +39,23 @@ import java.util.Map;
  * <li>For an {@link EntityManager}, every entity instance
  *     returned by any operation declared by this interface
  *     is in the managed state and belongs to the persistence
- *     context associated with the entity manager. An
- *     {@code EntityManager} is required to ensure that if an
- *     entity instance representing a given record already
- *     exists in its associated persistence context, then
- *     exactly that instance is returned by every operation
- *     that returns an entity representing the record.
+ *     context associated with the entity manager. The
+ *     instance remains in the managed state until that
+ *     entity manager is closed and its persistence context
+ *     is destroyed. An {@code EntityManager} is required
+ *     to ensure that if an entity instance representing a
+ *     given record already exists in its associated
+ *     persistence context, then exactly that instance is
+ *     returned by every operation that returns an entity
+ *     representing the record.
  * </ul>
+ * <p>Note that, in a Jakarta EE container environment, a
+ * transaction-scoped entity manager which is accessed
+ * outside the scope of a container transaction is destroyed
+ * immediately after it is called. And so such an entity
+ * manager may appear to return unmanaged instances to its
+ * client. But this is merely a consequence of the extremely
+ * short lifetime of its persistence context.
  *
  * <p>Operations which return entity instances are, by default,
  * permitted to retrieve state from the second-level cache, if


api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -0,0 +1,529 @@
...
+     * @since 2.0
+     */
+    <T> T unwrap(Class<T> cls);
+
+    /**
+     * Return the underlying provider object for the
+     * {@link EntityHandler}, if available. The result of this
+     * method is implementation-specific.
+     * <p>The {@code unwrap} method is to be preferred for new
+     * applications.
```
marko-bekhta: maybe let's deprecate it then and don't bring it to the common interface ?

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,262 @@
...
+import java.util.List;
+
+/**
+ * Interface enabling more direct control over interaction with the
+ * database than what is possible with {@link EntityManager}.
+ *
+ * <p>An instance of {@code EntityAgent} must be obtained from
+ * an {@link EntityManagerFactory}, and is able to manage
+ * persistence of entities belonging to the associated persistence
+ * unit.
```
marko-bekhta: > manage persistence

maybe it would be better to say something like: perform operation on entities beloning to ...
just wondering if we say "manage persistance" wouldn't it be brining us closer to a stateful EM?

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,262 @@
...
+    void insert(Object entity);
+
+    /**
+     * Insert every record in the given list.
+     *
+     * @param entities The entities to be inserted.
+     *
+     * @see #insert(Object)
+     */
+    void insertMultiple(List<?> entities);
```
marko-bekhta, line 92: how about going with

```suggestion
    void insertMultiple(Collection<?> entities);
```
my thinking is that list implies an order of entities, so do we guarantee that the entities are inserted in exactly the order that user provided?
also thinking if there could be a usecase where someone builds a set of entities to keep only "unique" records and then they'd need to do an unnecessary conversion to a list..

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -187,6 +187,36 @@ public interface EntityManagerFactory extends AutoCloseable {
...
+     * it is invoked.
+     * <p>The {@link EntityAgent#isOpen} method will return true
+     * on the returned instance.
+     * @return entity agent instance
+     * @throws IllegalStateException if the entity manager factory
+     * has been closed
+     *
+     * @since 4.0
+     */
+    EntityAgent createEntityAgent();
```
marko-bekhta, line 202: what if we'd go with a new `EntityHandlerFactory extends EMF` ? and have the new methods there ?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1368,13 +1428,27 @@ access, but must not be `static` or `final`.
...
+Annotations for callbacks corresponding to database operations:
+
+* `@PreInsert`
+* `@PostInsert`
+* `@PreUpdate`
+* `@PostUpdate`
+* `@PreUpsert`
+* `@PostUpsert`
+* `@PreDelete`
+* `@PostDelete`
```
marko-bekhta, line 1450: wondering if we need any `@PostSelect` for completeness ?

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,262 @@
...
+ * {@linkplain #update update} is an explicit operation. Since
+ * there's no {@linkplain EntityManager#flush flush operation},
+ * and since the entities themselves are detached, modifications to
+ * the entities are never automatically detected and made persistent.
+ *
+ * @since 4.0
+ *
+ * @see EntityManager
+ */
+public interface EntityAgent extends EntityHandler {
```
yrodiere: Some bikeshedding available in #675.

Reminder of proposals so far, along with main drawbacks:

1. `EntityAgent`: confusing, especially since people will think of AI.
2. `EntityStore`: not aligned on the "thing that does" approach to naming so far (manager, factory, handler, ...)
3. `EntityOperator`: I didn't see any comment against it? Though it may be more appropriate as an alternative to `EntityHandler`.
4. `EntityBroker`: a bit exotic at first sight...

I would suggest `EntityLoader` (as in upload/download), but... It's probably too imprecise and also likely will be understood as in "download" rather than upload.

I think we're looking for a word that makes you think "this thing moves entities to and from the database". `EntityBroker` is... that, I suppose. A bit odd, but I can't say what comes to my mind ("carrier", "transporter", "mover", "hauler") is better...

api/src/main/java/jakarta/persistence/EntityHandler.java
gavinking: If you want to deprecate it, I'm not against doing that, but you would have to open an issue :)

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: @marko-bekhta I updated that text. Is that better?

marko-bekhta: yes, thanks 👍🏻 🙂

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: > do we guarantee that the entities are inserted in exactly the order that user provided?

Yes, we should, probably. That's useful.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: Then you would have two objects instead of one. That just seems like unnecessary extra complexity. What does it buy you?

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: How would it be different to `@PostLoad`?

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: I have clarified that order is preserved.

api/src/main/java/jakarta/persistence/EntityHandler.java
marko-bekhta: https://github.com/jakartaee/persistence/issues/796 🙂

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
marko-bekhta: oh I meant it only in a sense of naming (to match the factory "responsibilities").
my thinking was: we have the new interface and make use of it moving forward while "deprecate" the original EMF, since we cannot just rename the current interface. But, can I think of other benefit beyond the naming ? no 🙂

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: we list `@PostLoad` under the 

> Annotations for callbacks corresponding to managed entity instance lifecycle operations

so if we'd want to use the new stateless `EntityAgent` we would't get that one executed (post load) as we won't be dealing with a managed entity, or I'm misinterpreting it?

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: I don't want to deprecate something that every JPA program is using *just to change its name*.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: `@PostLoad` would always be executed, just like `@PostInsert`, `@PostUpdate`, and `@PostDelete`.

gavinking: Oh I see, it's in the wrong list.

gavinking: Fixed, thanks!

api/src/main/java/jakarta/persistence/EntityAgent.java
beikov: I still like `EntityStore` most, but could probably live with `EntityOperator`. Did we consider `EntityClient` before?
Anyway, how are we going to agree on a name here? Put it out for vote?

gavinking: > Put it out for vote?

No way. This is not and has never been a democracy.

I have already explained with good reasons why `EntityStore` and `EntityOperator` are not acceptable to me. Forget about those ones.

I'm always very open to hear interesting new and creative proposals, but just repeating the same proposals is unlikely to change my mind.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+package jakarta.persistence;
+
+import java.util.List;
+
+/**
+ * Interface enabling more direct control over interaction with the
+ * database than what is possible with {@link EntityManager}.
+ *
+ * <p>An instance of {@code EntityAgent} must be obtained from
+ * an {@link EntityManagerFactory}, and is able to execute
```
FroMage: You mean an `EntityAgentFactory`?

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
FroMage: I was going to suggest the same as Marko, but upward.

It _is_ weird to get a XFactory to produce Y, for sure, we'll get ridiculed for that. OTOH I also agree that this would be a big change just for a name…

api/src/main/java/jakarta/persistence/EntityAgent.java
FroMage: 😬

api/src/main/java/jakarta/persistence/EntityAgent.java
FroMage: Alright, here goes _creative_ 🤣

If one type of entity requires a manager… the other type of entity could be working under a handler, agent or broker or operator… and they both work in an 🥁 `EntityFactory` 🥳

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: Of course the perfect and correct thing to call the EMF now would be `EntityStore`. But as I said to Marko, I don't want to deprecate one of the two most important interfaces in JPA just to change its name.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+// Contributors:
+//     Gavin King      - 4.0
+
+package jakarta.persistence;
+
+import java.util.List;
+
+/**
+ * Interface enabling more direct control over interaction with the
+ * database than what is possible with {@link EntityManager}.
```
njr-11: I tried to think of a summary that will give the user an overview of what this interface gives them,
```suggestion
 * Provides entity operations that are performed independently
 * of a persistence context. This interface enables more direct
 * control over interaction with the database than what is
 * possible with {@link EntityManager}.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
+/*
+ * Copyright (c) 2008, 2025 Oracle and/or its affiliates. All rights reserved.
```
njr-11: For a new file,
```suggestion
 * Copyright (c) 2025 Contributors to the Eclipse Foundation
```

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: minor grammar correction
```suggestion
 * an {@link EntityManagerFactory} and is able to execute
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+ * an {@link EntityManagerFactory}, and is able to execute
+ * persistence-related operations for entity types belonging to the
+ * associated persistence unit.
+ *
+ * <p>An {@code EntityAgent} has no associated persistence context,
+ * and works only with detached entity instances. When a method of
+ * this interface is called, any necessary interaction with the
+ * database happens immediately and synchronously. In particular,
+ * {@linkplain #update update} is an explicit operation. Since
+ * there's no {@linkplain EntityManager#flush flush operation},
```
```suggestion
 * there is no {@linkplain EntityManager#flush flush operation},
```

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: Choosing a good name is difficult. I would favor the approach of deferring the name decision to a separate PR and not holding this one up over it.  There is still a long time to think of good name ideas before EE 12 needs to be done and it would be valuable to have this interface with any name in a 4.0 Milestone to help the capability be experimented with.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * </ul>
+     *
+     * @param entity a new or removed entity instance
+     *
+     * @return The identifier of the inserted entity
+     *
+     * @throws IllegalArgumentException if the given instance is
+     *         determined to not be new or removed
+     */
+    void insert(Object entity);
```
njr-11, line 71: Matches up with `@return The identifier of the inserted entity` which will be very useful when the identifier is generated
```suggestion
    Object insert(Object entity);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Insert every record in the given list. The records are
+     * inserted in the order in which they occur in the given list.
+     *
+     * @param entities The entities to be inserted.
+     *
+     * @see #insert(Object)
+     */
+    void insertMultiple(List<?> entities);
```
njr-11, line 92:
```suggestion
     * @return The identifiers of the inserted entities, in the same order
     *         as the {@code entities} list.
     *
     * @see #insert(Object)
     */
    List<?> insertMultiple(List<?> entities);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     *     triggered.
+     * <li>The {@link jakarta.persistence.PostInsert} callback is
+     *     triggered if the operation is successful.
+     * </ul>
+     *
+     * @param entity a new or removed entity instance
+     *
+     * @return The identifier of the inserted entity
+     *
+     * @throws IllegalArgumentException if the given instance is
```
njr-11, line 62: EntityExistsException seems like a more descriptive exception for this. If we add it here, its Javadoc will also need an update so say it can happen on insert in addition to persist.
```suggestion
     * @throws EntityExistsException if the supplied entity has a unique
     *              identifier and already exists in the database.
     * @throws IllegalArgumentException if the given instance is
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     *         determined to not be new or removed
+     */
+    void insert(Object entity);
+
+    /**
+     * Insert every record in the given list. The records are
+     * inserted in the order in which they occur in the given list.
```
```suggestion
     * inserted in the order in which they occur in the given list
     * and according to the requirements of the {@link insert(Object)}
     * method.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * <p>
+     * If the entity {@link Id} field is declared to be generated,
+     * for example, if it is annotated {@link GeneratedValue}, the
+     * id is generated and assigned to the given instance.
+     * <p>
+     * <ul>
+     * <li>The {@link jakarta.persistence.PreInsert} callback is
+     *     triggered.
+     * <li>The {@link jakarta.persistence.PostInsert} callback is
+     *     triggered if the operation is successful.
```
njr-11, line 57: Should this be tied to transaction commit instead?  Also applies in other places.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * @param entity a detached entity instance
+     *
+     * @throws IllegalArgumentException if the given instance is
+     *         determined to not be detached
+     */
+    void update(Object entity);
+
+    /**
+     * Update every record in the given list. The records are
+     * updated in the order in which they occur in the given list.
```
```suggestion
     * updated in the order in which they occur in the given list
     * and according to the requirements of the {@link update(Object)}
     * method.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * <li>The {@link jakarta.persistence.PostUpdate} callback is
+     *     triggered if the operation is successful.
+     * </ul>
+     *
+     * @param entity a detached entity instance
+     *
+     * @throws IllegalArgumentException if the given instance is
+     *         determined to not be detached
+     */
+    void update(Object entity);
```
njr-11, line 115: It would be nice to have a way of knowing the entity is found for update.
```suggestion
     * @return whether a matching entity was found in the database.
     * @throws IllegalArgumentException if the given instance is
     *         determined to not be detached
     */
    boolean update(Object entity);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * <li>The {@link jakarta.persistence.PostDelete} callback is
+     *     triggered if the operation is successful.
+     * </ul>
+     *
+     * @param entity a detached entity instance
+     *
+     * @throws IllegalArgumentException if the given instance is
+     *         determined to not be detached
+     */
+    void delete(Object entity);
```
njr-11, line 160: It would be nice to have a way of knowing if the entity was already deleted,
```suggestion
     * @param entity a detached entity instance
     * @return whether the entity was found in the database and deleted
     * @throws IllegalArgumentException if the given instance is
     *         determined to not be detached
     */
    boolean delete(Object entity);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Delete every record in the given list. The records are
+     * deleted in the order in which they occur in the given list.
```
```suggestion
     * deleted in the order in which they occur in the given list
     * and according to the requirements of the {@link delete(Object)}
     * method.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     *
+     * @param entity a detached entity instance, or a new instance
+     *               with an assigned identifier
+     * @throws IllegalArgumentException if the entity has a null id
+     */
+    void upsert(Object entity);
+
+    /**
+     * Upsert every record in the given list. The records are
+     * upserted in the order in which they occur in the given list.
```
```suggestion
     * upserted in the order in which they occur in the given list
     * and according to the requirements of the {@link upsert(Object)}
     * method.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Retrieve a record.
+     *
+     * @param entityClass The class of the entity to retrieve
+     * @param id The id of the entity to retrieve
+     *
+     * @return a detached entity instance
+     */
+    <T> T get(Class<T> entityClass, Object id);
```
njr-11: Need to indicate what happens when not found. Maybe use Optional as well?
Later revised to use `null` instead of `Optional`,
```suggestion
     * @return a detached entity instance if a matching entity
     *             is found in the database. Otherwise, returns
     *             {@code null}
     */
    <T> T get(Class<T> entityClass, Object id);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Retrieve a record, obtaining the specified lock mode.
```
```suggestion
     * Retrieve a record, obtaining locks as needed per the specified lock mode.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     *
+     * @param entityClass The class of the entity to retrieve
+     * @param id The id of the entity to retrieve
+     * @param lockMode The lock mode to apply to the entity
+     *
+     * @return a detached entity instance
+     */
+    <T> T get(Class<T> entityClass, Object id, LockModeType lockMode);
```
```suggestion
     * @return a detached entity instance if a matching entity
     *             is found in the database. Otherwise, returns
     *             {@code null}
     */
    <T> T get(Class<T> entityClass, Object id, LockModeType lockMode);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Retrieve a record, fetching associations specified by the
+     * given {@link EntityGraph}.
+     *
+     * @param graph The {@link EntityGraph}
+     * @param graphSemantic a {@link GraphSemantic} specifying
+     *                      how the graph should be interpreted
```
njr-11: graphSemantic is not part of the method signature
```suggestion
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * @param id The id of the entity to retrieve
+     *
+     * @return a detached entity instance
+     *
+     * @since 6.3
+     */
+    <T> T get(EntityGraph<T> graph, Object id);
```
```suggestion
     * @return a detached entity instance if a matching entity
     *             is found in the database. Otherwise, returns
     *             {@code null}
     */
    <T> T get(EntityGraph<T> graph, Object id);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Retrieve a record, fetching associations specified by the
+     * given {@linkplain EntityGraph load graph}, and obtaining
+     * the specified {@link LockModeType lock mode}.
```
```suggestion
     * Retrieve a record, fetching associations specified by the
     * given {@linkplain EntityGraph load graph}, and obtaining
     * locks as needed, according to the specified
     * {@link LockModeType lock mode}.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     *
+     * @param graph The {@linkplain EntityGraph load graph}
+     * @param id The id of the entity to retrieve
+     * @param lockMode The lock mode to apply to the entity
+     *
+     * @return a detached entity instance
+     */
+    <T> T get(EntityGraph<T> graph, Object id, LockModeType lockMode);
```
```suggestion
     * @return a detached entity instance if a matching entity
     *             is found in the database. Otherwise, returns
     *             {@code null}
     */
    <T> T get(EntityGraph<T> graph, Object id, LockModeType lockMode);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Retrieve multiple rows, returning entity instances in a
```
```suggestion
     * Retrieve multiple records, returning entity instances in a
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     *         null elements representing missing entities
+     */
+    <T> List<T> getMultiple(Class<T> entityClass, List<?> ids);
+
+    /**
+     * Refresh the entity instance state from the database.
+     *
+     * @param entity The entity to be refreshed.
+     */
+    void refresh(Object entity);
```
njr-11, line 248: It might be nice to know if the entity was refreshed
```suggestion
     * @return {@code true} if the entity was found in the database,
     *             otherwise {@code false}.  
     */
    boolean refresh(Object entity);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     */
+    void refreshMultiple(List<?> entities);
+
+    /**
+     * Refresh the entity instance state from the database.
+     *
+     * @param entity The entity to be refreshed.
+     * @param lockMode The LockMode to be applied.
+     */
+    void refresh(Object entity, LockModeType lockMode);
```
njr-11, line 275:
```suggestion
     * @return {@code true} if the entity was found in the database,
     *             otherwise {@code false}.  
     */
    boolean refresh(Object entity, LockModeType lockMode);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+
+    /**
+     * Fetch an association or collection that's configured for lazy loading.
```
```suggestion
     * Fetch an association or collection that is configured for lazy loading.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * <pre>
+     * Book book = agent.get(Book.class, isbn);  // book is immediately detached
+     * agent.fetch(book.getAuthors());           // fetch the associated authors
+     * book.getAuthors().forEach(author -> ... );  // iterate the collection
+     * </pre>
+     *
+     * @param association a {@linkplain FetchType#LAZY lazy} association
+     */
+    void fetch(Object association);
```
njr-11: Usage might be more convenient if this returned the association, so you could do,
```
agent.fetch(book.getAuthors()).forEach(author -> ... );
```

```suggestion
     * @return the fetched association.
     */
    <T> T fetch(T association);
```

njr-11 (commented):
The EntityAgent is an excellent idea.  I started out by reviewing only the EntityAgent interface and posted a number of suggestions/clarifications/ideas to think about.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: I had this originally, but the feedback on the older PR was that this was unnecessary, and I guess I was inclined to agree.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: > All of the entities are inserted within the same transaction, such that none of the inserts are committed if any insert attempt fails.

Dunno what to think of this. I guess my knee-jerk is that for this more "bare-metal"  interface, if the user wants a tx then they should explicitly start a tx.

OTOH I'm not very sure what happens with JDBC batching. Can a JDBC batch partially succeed? If not, then perhaps that's an argument for why this operation should be atomic.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: No, because when this is channelled to a Jakarta Data `PostInsertEvent`, the *observer* chooses when they want to be notified. See the documentation of `jakarta.data.event.LifecycleEvent`.

Now, perhaps Jakarta Persistence should have some kind of pre- and post- completion events, but that's a whole different issue.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: Good catch. We do need to say **something** about this.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: > Need to indicate what happens when not found.

Yes, we do need to say something. But let's stay away from `Optional`, because Java is already moving away from it.

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: > feedback on the older PR was that this was unnecessary

The alternative would be updating the supplied entity instance to assign the new id value. I just re-read the javadoc, and it looks like you already have that covered, so I'll resolve this conversation and switch my review comment to cover removing the `@return` from the Javadoc

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: Not particularly keen on this change of wording, because you are permitted to pass in `LockModeType.NONE`.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,266 @@
...
+     * <ul>
+     * <li>The {@link jakarta.persistence.PreInsert} callback is
+     *     triggered.
+     * <li>The {@link jakarta.persistence.PostInsert} callback is
+     *     triggered if the operation is successful.
+     * </ul>
+     *
+     * @param entity a new or removed entity instance
+     *
+     * @return The identifier of the inserted entity
```
njr-11: This method returns `void`
```suggestion
```

gavinking:
Thanks, @njr-11, lots of suggestions here that we should discuss on a call. Too hard to do it via comments.

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: I'm fine with `null` instead of `Optional`.

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: Good point, I forgot about `NONE`.  I was trying to rephrase it to indicate obtaining a lock rather than obtaining a lock mode, which sounded awkward.  I'll make a better attempt at it.

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: > I'm not very sure what happens with JDBC batching. Can a JDBC batch partially succeed?

The [Javadoc for Statement.executeBatch](https://docs.oracle.com/en/java/javase/25/docs/api/java.sql/java/sql/Statement.html#executeBatch()) is complicated and a bit confusing.  It allows the JDBC driver to choose whether or not to continue processing batched SQL, but in that case, it also provides a way (BatchUpdateException.getUpdateCounts) for the user to know exactly which statements were successful and which failed.  It is somewhat unclear if that same pattern can be applied to the `executeBatch` `int[]` return value or if only BatchUpdateException can be used. But either way, a JDBC driver does have the option and there is a way for the user to know what was successful vs not.  That last part is important.  The `insertMultiple` method doesn't have any way of indicating partial success, so I had suggested the easy approach of stating it all rolls back.  If I were a user I wouldn't ever use an operation where I cannot know how much of it worked.  We should either roll it all back or have a way to indicate.

gavinking: I mean I would just never do write operations outside a transaction.

I'm sorta sympathetic to what you're saying, but I just don't see what scenario this would be useful for.

What you're saying is that when the user calls `insertMultiple()` outside a transaction, we have to implicitly start a tx, do the inserts, and then commit the tx for them. (I believe that's always at least two interactions with the database.)

But that gives you atomicity for just one method call. They would have been much better off to do *all* their work in a transaction, and amortize the cost of the commit call over all that work.

And it cuts against the grain of this API which is supposed to be all about putting the client in control.

njr-11: Sure that's fine to not go with the transaction approach. But in that case, we ought to indicate to the user which inserts succeeded vs which did not.  I'll update my suggestion accordingly.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Insert every record in the given list. The records are
+     * inserted in the order in which they occur in the given list.
+     *
+     * @param entities The entities to be inserted.
+     *
+     * @see #insert(Object)
+     */
+    void insertMultiple(List<?> entities);
```
njr-11, line 92:
```suggestion
     * @return a list, ordered the same as the given {@code entities}, with each
     *             element indicating whether the corresponding insert succeeded.
     *
     * @see #insert(Object)
     */
    List<Boolean> insertMultiple(List<?> entities);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Update every record in the given list. The records are
+     * updated in the order in which they occur in the given list.
+     *
+     * @param entities The entities to be updated.
+     *
+     * @see #update(Object)
+     */
+    void updateMultiple(List<?> entities);
```
njr-11, line 137:
```suggestion
     * @return a list, ordered the same as the given {@code entities}, with each
     *             element indicating whether the corresponding update succeeded.
     * @see #update(Object)
     */
    List<Boolean> updateMultiple(List<?> entities);
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,264 @@
...
+
+    /**
+     * Delete every record in the given list. The records are
+     * deleted in the order in which they occur in the given list.
+     *
+     * @param entities The entities to be deleted.
+     *
+     * @see #delete(Object)
+     */
+    void deleteMultiple(List<?> entities);
```
njr-11, line 182:
```suggestion
     * @return a list, ordered the same as the given {@code entities}, with each
     *             element indicating whether the corresponding delete succeeded.
     *
     * @see #delete(Object)
     */
    List<Boolean> deleteMultiple(List<?> entities);
```

njr-11 (commented):
I started switching the suggestions over to return a List of indicators of success, but after doing this, just realized that we instead ought to be reporting an error back to the application like BatchUpdateException does so that the user can know why a failure occurred in case it represents something they need to fix, like an invalid value for an entity attribute.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,264 @@
...
+     * <li>The {@link jakarta.persistence.PostInsert} callback is
+     *     triggered if the operation is successful.
+     * </ul>
+     *
+     * @param entity a new or removed entity instance
+     *
+     * @throws IllegalArgumentException if the given instance is
+     *         determined to not be new or removed
+     */
+    void insert(Object entity);
```
njr-11, line 71: This goes along with the other suggestions
```suggestion
     * @return {@code true} if the entity was successfully inserted;
     *         {@code false} if a matching entity already exists in the
     *         database
     * @throws IllegalArgumentException if the given instance is
     *         determined to not be new or removed
     */
    boolean insert(Object entity);
```

api/src/main/java/jakarta/persistence/EntityHandler.java
njr-11: I agree with omitting `getDelegate` from the common interface.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -494,4 +524,63 @@ <E> EntityListenerRegistration addListener(
...
+     * control to the client.
+     *
+     * @param handlerClass the type of {@link EntityHandler} to create, for
+     *                     example, {@link EntityAgent EntityAgent.class}.
+     * @param work a function to be called in the scope of the transaction
+     * @return the value returned by the given function
+     *
+     * @since 4.0
+     */
+    <R, H extends EntityHandler> R callInTransaction(Class<H> handlerClass, Function<EntityAgent,R> work);
```
njr-11:
```suggestion
    <R, H extends EntityHandler> R callInTransaction(Class<H> handlerClass, Function<H,R> work);
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2565,19 +2649,23 @@ The effect of applying `setMaxResults` or
...
-`getSingleResult`, and `getSingleResultOrNull` methods are not required
-to be invoked within a transaction context unless such a lock mode has
-been specified for the queryfootnote:[A lock mode is
-specified for a query by means of the `setLockMode` method or by
-specifying the lock mode in the `NamedQuery` annotation.]. If an entity manager with
-transaction-scoped persistence context is in use, the resulting entities
-will be detached; if an entity manager with an extended persistence
-context is used, they will be managed. See <<a11431>> for further
-discussion of entity manager use outside a transaction and persistence context types.
+Apart from `executeUpdate()`, methods `Query` and `TypedQuery` may be
```
```suggestion
Apart from `executeUpdate()`, methods of `Query` and `TypedQuery` may be
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2565,19 +2649,23 @@ The effect of applying `setMaxResults` or
...
+the methods `getResultList()`, `getSingleResult()`, and `getSingleResultOrNull()`
+may be called when no transaction is activefootnote:[A lock mode may be
+specified for a query by calling `setLockMode()` or by specifying the
+`lockMode` member of the `@NamedQuery` annotation.].
+
+- If the `Query` or `TypedQuery` object was obtained from an `EntityAgent`
+  or from an `EntityManager` with a transaction-scoped persistence context,
+  the entities returned by the query are detached;
+- otherwise, if the query object was obtained or from an `EntityManager`
+  with an extended persistence, the returned entities are managed.
```
```suggestion
  with an extended persistence context, the returned entities are managed.
```

spec/src/main/asciidoc/ch07-entitymanagers-and-persistence-contexts.adoc
```diff
@@ -123,33 +123,30 @@ public class MySessionBean implements MyInterface {
...
-More than one entity manager factory
-instance may be available simultaneously in the
-JVM.footnote:[This may be the case
-when using multiple databases, since in a typical configuration a single
-entity manager only communicates with a single database. There is only
-one entity manager factory per persistence unit, however.]
+More than one entity manager factory instance may be available simultaneously
+in the JVM.footnote:[This is usually the case when the application uses more
+than one database, since it's typical that each entity manager communicates
+which exactly one database. On the other hand, there's usually only one entity
```
njr-11:
```suggestion
than one database, since it is typical that each entity manager communicates
which exactly one database. On the other hand, there is usually only one entity
```

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: @njr-11 I have proposed a similar change.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: @njr-11 I have proposed a similar change. Are you happy with that?

gavinking:
> I started switching the suggestions over to return a List of indicators of success, but after doing this, just realized that we instead ought to be reporting an error back to the application like BatchUpdateException does so that the user can know why a failure occurred in case it represents something they need to fix, like an invalid value for an entity attribute.

@njr-11 Yeah, it's like I [commented on the issue](https://github.com/jakartaee/persistence/issues/374#issuecomment-3608850059). I think it's much better to notify failure via an exception, because in Java we have no way to force the client to check the return value.

I'm not against introducing a subtype of `OptimisticLockException` which would give implementations a way to notify the client as to which rows were successfully updated. On the other hand, such subtypes can also be defined by provider extensions so it's not absolutely critical to define this _now_.

An even bigger question is whether exceptions thrown by `EntityAgent` are _recoverable_. Exceptions thrown by an `EntityManager` are not, for good reasons, but those reasons don't apply to `EntityAgent`.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: Notifying failures via return codes is generally an anti-pattern in Java, since Java's type system does not provide a way to force the client to check the return codes. Checking that every value in a list is `true` is especially clumsy for the client.

So, as I commented on the issue, I think failures should be reported via an exception. And I think you already realized that. So I'm going to mark these suggestions as resolved.

api/src/main/java/jakarta/persistence/EntityHandler.java
gavinking: I have moved it back down, so that #797 may be applied cleanly after this work is merged.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: If you guys want to rename `EntityMangerFactory` to `EntityStore`, that's something we can discuss, but we can do that as part of a separate issue. It doesn't need to be done as an atomic part of this work.

So I'm going to mark these comments as resolved.

gavinking:
Based on feedback I have:

1. clarified how insert/update/delete/upsert/refresh report failure (@njr-11 please review!)
2. demoted `getDelegate()` back to `EntityManager`

I've also marked `EntityHandler` as `sealed`, with makes things a bit more convenient if you want to write generic code which abstracts over EHs.

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: I have (tentatively) implemented the solution proposed here: https://github.com/jakartaee/persistence/issues/374#issuecomment-3608850059

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: I have proposed this solution: https://github.com/jakartaee/persistence/issues/374#issuecomment-3608850059

gavinking:
Finally, I have tentatively moved the `find()` and `get()` methods up to `EntityHandler` and specified that:

- `get()` methods throw when the entity is not found, whereas
- `find()` methods return null.

as proposed here: https://github.com/jakartaee/persistence/issues/374#issuecomment-3608850059 and solving #724.

Thoughts?

gavinking:
Moved out of draft mode because I think this is pretty close to being mergeable now. (Merging this PR of does not of course imply that we won't make changes to it.)

api/src/main/java/jakarta/persistence/EntityAgent.java
njr-11: Yes, looks good

njr-11:
> An even bigger question is whether exceptions thrown by `EntityAgent` are _recoverable_. Exceptions thrown by an `EntityManager` are not, for good reasons, but those reasons don't apply to `EntityAgent`.

Right, at least in the cases I can think of having the exceptions be recoverable would be ideal and align with the direction of putting the user more in control, and it opens up some useful patterns that applications will be able to take advantage of.

> a subtype of `OptimisticLockException` which would give implementations a way to notify the client as to which rows were successfully updated

I think users will need something like that unless we limit usage to running within a transaction and rolling back.  That said, we could certainly defer this piece to a different PR and allow some more time to think through how best to externalize it.

gavinking:
> Right, at least in the cases I can think of having the exceptions be recoverable would be ideal and align with the direction of putting the user more in control, and it opens up some useful patterns that applications will be able to take advantage of.

Alright, so I think we should investigate this, but I need to chew on it for a bit and talk it over with folks from the Hibernate team. It's something I have not thought through at all. I will open an issue.

njr-11:
> 1. clarified how insert/update/delete/upsert/refresh report failure (@njr-11 please review!)

Looks good!  Raising an exception looks good for the multiple entity methods, and also for the single entity methods if the exception is recoverable.  In another PR, we can consider some ways of providing information on which, if any, entities succeeded when a multiple entity method raises the exception.  Subclassing the exception was one idea.  Another idea could be chaining an exception that provides the information.

gavinking:
See  #798.

gavinking:
I have done some work on adding more `@throws PersistenceException` to more clearly distinguish operations of the `Entity*` interfaces which are expected to hit the database.

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -0,0 +1,1071 @@
...
+ * depends on the semantics of the subtype.
+ * <ul>
+ * <li>For an {@link EntityAgent}, there is no associated
+ *     persistence context, and so every entity instance
+ *     returned by any operation is in the detached state.
+ *     Operations on an entity agent never return the same
+ *     entity instance twice.
+ * <li>For an {@link EntityManager}, every entity instance
+ *     returned by any operation declared by this interface
+ *     is in the managed state and belongs to the persistence
```
scottmarlow, line 41: I think each returned entity instance is only in managed state if there is an active Jakarta Transactions transaction (JTA).  

Perhaps change to something like the following:

```
*  For an {@link EntityManager}, every entity instance
 *     returned by any operation declared by this interface
 *     is in the detached state if there is no active JTA transaction.  
 *     If active JTA transaction instance is managed and belongs to the persistence

```

gavinking: > I think each returned entity instance is only in managed state if there is an active Jakarta Transactions transaction (JTA).

That's not really correct. What you're describing is the semantics of a transaction-scoped entity manager in an EE container, and is an aspect of how the container creates and manages instances of `EntityManager`. A given instance of `EntityManager` always returns objects in the managed state. It's just that for a transaction-scoped entity manager, the EM is then immediately closed by the container and the PC is destroyed.

scottmarlow: I agree that my feedback is based on an `EE container` aspect but the current wording kind of disagrees with this `Jakarta EE container` aspect.  

For reference, I see the term `clears` instead of `destroyed` in https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#a11894 but logically they both appear the same to the application in an EE container.

gavinking: @scottmarlow I have added words to clarify this confusion.

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -0,0 +1,1081 @@
...
+ *     returned by every operation that returns an entity
+ *     representing the record.
+ * </ul>
+ * <p>Note that, in a Jakarta EE container environment, a
+ * transaction-scoped entity manager which is accessed
+ * outside the scope of a container transaction is destroyed
+ * immediately after it is called. And so such an entity
+ * manager may appear to return unmanaged instances to its
+ * client. But this is merely a consequence of the extremely
+ * short lifetime of its persistence context.
```
gavinking, line 58: cc @scottmarlow

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -0,0 +1,291 @@
...
+     * agent.fetch(book.getAuthors());           // fetch the associated authors
+     * book.getAuthors().forEach(author -> ... );  // iterate the collection
+     * }
+     *
+     * @param association a {@linkplain FetchType#LAZY lazy} association
+     *
+     * @throws PersistenceException if a record could not be
+     *         read from the database
+     */
+    void fetch(Object association);
```
njr-11, line 290: Usage might be more convenient if this returned the association,

```suggestion
     * Book book = agent.get(Book.class, isbn);  // book is immediately detached
     * agent.fetch(book.getAuthors())            // fetch the associated authors
     *      .forEach(author -> ... );            // iterate the collection
     * }
     *
     * @param association a {@linkplain FetchType#LAZY lazy} association
     *
     * @return the fetched association.
     * @throws PersistenceException if a record could not be
     *         read from the database
     */
    <T> T fetch(T association);
```

gavinking:
> I also reposted one of my prior review comments, which is optional. I'm completely fine with not including, but I wasn't sure if it was noticed amid all the other comments from prior reviews.

@njr-11 Since Scott merged the PR, feel free to repurpose that change as a separate pull request. I'm guess I'm weakly in favor of it.

gavinking:
🎉🎉🎉

Thank you guys!

PR #799, opened 2025-12-04
link more directly to the relevant info about parameter binding

gavinking:


gavinking:
Merged because it's trivial.

Issue #79, opened 2014-05-19
Calling SQL functions via StoredProcedureQuery


lukasj:
JPA 2.1 provides a facility to call stored procedures, which greatly improves performance in many situations.

Unfortunately, it does not yet provide the same feature richess of JDBC. One partocular missing feature is the ability to read the result of a stored FUNCTION: The technical way to read those is defined by JDBC API, but not by JPA!

In JDBC, a return value can be simply queried by reading the OUT parameter at index 0, which is valid for all escaped functions when using the syntax {?=call myfunc(...)}. In JPA, providing a leading question mark leads to syntax errors. While using a native query using "SELECT myfunc(...)" certainly does work, it leads to unwanted performance overhead and code clutter, as it creates a CURSOR and wraps it using a JDBC ResultSet, possibly inducing additional network roundtrips to get that CURSOR's description and first row. JDBC prevents this by simply requesting the sole function result value as a side effect of the call, which is certainly available in the same roundtrip.

Hence, to spare overhead, improve performance and reduce application code size, it would be really great if JPA learns to deal with stored FUNCTION result values, just as JDBC can do it since many years.

A proposed syntax would be:

StoredFunctionQuery q = em.createStoredFunctionQuery("MyFunc(...)");
q.registerOutputParameter(0, Integer.class, OUT);
q.execute();
int = g.getParameter(0); <-- Returns value of '?=' in JDBC

gavinking:
This is functionality that Hibernate exposes as `markAsFunctionCall(resultClass)` method. It would be *extremely* trivial to add something similar to JPA's `StoredProcedureQuery`. Even though such function calls don't return entities, it's still a bit more convenient to call them via our APIs than via JDBC.

Therefore I'm inclined to consider this simple thing for JPA4.

gavinking:
In #800 I'm proposing a potential solution.

Borrowing language from the JDBC spec section 13.4.4, I'm calling the method `registerResultParameter()`, though an alternative would be `registerReturnType()` which might be easier for users to find.

gavinking:
@sebersole WDYT about adding a `reset()` method to `StoredProcedureQuery`?

Or, second possibility: what about making `StoredProcedureQuery` implement `AutoCloseable`?

gavinking:
Fixed, ten years later!

Issue #162, opened 2017-10-16
StoredProcedureQuery and statements leaks


lukasj:
_StoredProcedureQuery_ for a JDBC implementation of JPA must keep track of the JDBC statement instance used to call the procedure. Since _StoredProcedureQuery_ work in a stateful manner, the client process results returned by the query one by one, if at all, the query must be told at some time to close the statement.

Client code could be calling a procedure and not caring about all the results, since there is no way to "close" the _StoredProcedureQuery_ and as consequence, close the JDBC statement, the possibility of leaks is always present.

In our case we got [DB2 statements resource leaks][1] during testing, when migrating from Hibernate _ProcedureCall_ API to JPA _StoredProcedureQuery_. The only way to solve the problem was to unwrap the query instance to a [_ProcedureCall_][2] and close the ouputs object it provides.

_StoredProcedureQuery_ needs a way to be closed be it, for example, implementing _AutoCloseable_, that is different in the way other non procedural query instances work (stateless), or giving access to a Hibernate like object like [ProcedureOutputs][3]. The equivalent ProcedureOutputs could be _AutoCloseable_

Note: It is not common to leak too much JDBC statements this way, only on client code that requires to call one or more store procedures from some kind of loop, but still for code that only call one store procedure with one or more ouput result set, at least one statement could leak.

I have only tested this on Hibernate but I think other JPA implementations will leak too

1. EclipseLink has a [_close()_][4] method called from _executeUpdate()_ and _getSingleResult()_, any other way of retrieving results will leak (there is a [_finalize()_][5] method closing it too). Unwrapping and calling _close()_ should work as a workaround
2. OpenJPA: I don't find any statement closing on their [StoredProcedureQueryImpl class][6]
3. Hibernate: has a [_release()_][7] method only called from _executeUpdate()_, any other way of retrieving results will leak (there is no _finalize()_ method releasing leaked statements). Unwrapping and calling _getOutputs().release()_ work as a workaround.

[1]: http://www-01.ibm.com/support/docview.wss?uid=swg21504334
[2]: https://docs.jboss.org/hibernate/orm/5.2/javadocs/org/hibernate/procedure/ProcedureCall.html
[3]: https://docs.jboss.org/hibernate/orm/5.1/javadocs/org/hibernate/procedure/ProcedureCall.html#getOutputs--
[...]

PR #800, opened 2025-12-05
StoredProcedureQuery function and parameter handling

gavinking:
to allow calls to "stored functions"

see #79 and #824

### DIFF
--- a/api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
+++ b/api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
@@ -287,6 +287,9 @@ public interface StoredProcedureQuery extends Query {
      * The result may be retrieved after execution by calling
      * {@link #getOutputParameterValue(Parameter)}.
      * @param resultType the type of the result parameter
+     * @return an object representing the parameter, which may
+     *         be passed to {@link #setParameter(Parameter, Object)}
+     *         and {@link #getOutputParameterValue(Parameter)}
      * @since 4.0
      */
     <T> Parameter<T> registerResultParameter(Class<T> resultType);
@@ -299,7 +302,9 @@ public interface StoredProcedureQuery extends Query {
      * @param position the parameter position
      * @param type the type of the parameter
      * @param mode the parameter mode
-     * @return the same query instance
+     * @return an object representing the parameter, which may
+     *         be passed to {@link #setParameter(Parameter, Object)}
+     *         and {@link #getOutputParameterValue(Parameter)}
      * @since 4.0
      */
     <T> Parameter<T> registerParameter(int position, Class<T> type,
@@ -313,7 +318,9 @@ public interface StoredProcedureQuery extends Query {
      *                      or specified in metadata
      * @param type the type of the parameter
      * @param mode the parameter mode
-     * @return the same query instance
+     * @return an object representing the parameter, which may
+     *         be passed to {@link #setParameter(Parameter, Object)}
+     *         and {@link #getOutputParameterValue(Parameter)}
      * @since 4.0
      */
     <T> Parameter<T> registerParameter(String parameterName, Class<T> type,


gavinking:
@beikov @sebersole @loiclefevre Does this look right to you guys?

loiclefevre:
> @beikov @sebersole @loiclefevre Does this look right to you guys?

Hi Gavin,
My comments on the changes made in this PR:
- A stored procedure can have output parameters
- A function can have a result AND output parameters
- A stored *procedure* can return multiple result sets/cursors

Said otherwise, these 3 use cases are not covered yet.

loiclefevre:
image

https://docs.oracle.com/en/database/oracle/oracle-database/26/lnpls/formal-parameter-declaration.html#GUID-5BA8E033-96B9-439A-A4FC-4844FEC14AD8__CJAJDGCC

loiclefevre:
https://docs.oracle.com/en/database/oracle/oracle-database/26/lnpls/subprogram-parameters.html#GUID-518B8827-26CC-4734-B799-ACB038185638__BCFHBGBI

image

loiclefevre:
And finally: https://oracle-base.com/articles/12c/implicit-statement-results-12cr1#return_result

gavinking:
@loiclefevre OK, but what's the conclusion from that. The original issue asked for support for the JDBC escape syntax `{? = call fun(x,y)}` or equivalent, for which JPA currently has no API. The purpose of this new method would be to request use of that syntax (or equivalent) and register the "result parameter", as the JDBC spec seems to call it in section 13.4.4.

loiclefevre:
> @loiclefevre OK, but what's the conclusion from that. The original issue asked for support for the JDBC escape syntax `{? = call fun(x,y)}` or equivalent, for which JPA currently has no API. The purpose of this new method would be to request use of that syntax (or equivalent) and register the "result parameter", as the JDBC spec seems to call it in section 13.4.4.

Well, looks good to me :)

gavinking:
> Well, looks good to me :)

Haha OK :-)

Do you think this is the right language to use to describe it?

beikov (approved):

loiclefevre:
> > Well, looks good to me :)
> 
> Haha OK :-)
> 
> Do you think this is the right language to use to describe it?

yes, definitely

sebersole:
I mean this is exactly what I did for Hibernate, so +1 ;)

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -279,6 +280,17 @@ StoredProcedureQuery setParameter(int position,
      */
     StoredProcedureQuery setTimeout(Integer timeout);
 
+    /**
+     * Mark this as a call to a stored procedure with a result
+     * parameter and register the type of the result parameter
+     * This is typically required when calling a stored function.
```
sebersole, line 286: What would be another use?  This would *only* be used when calling a function, no?

gavinking: Probably: I'm not sure.

gavinking:
This change now also addresses #824. I now consider this a high priority change set, since it dramatically increases the usefulness of `StoredProcedureQuery`. I know that wasn't on the road map, but it is now.

gavinking:
@njr-11 @anija-anil @scottmarlow need a review please 🙏

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -279,6 +280,45 @@ StoredProcedureQuery setParameter(int position,
      */
     StoredProcedureQuery setTimeout(Integer timeout);
 
+    /**
+     * Mark this as a call to a stored procedure with a result
+     * parameter and register the type of the result parameter
```
```suggestion
     * parameter and register the type of the result parameter.
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -279,6 +280,45 @@ StoredProcedureQuery setParameter(int position,
...
+     * by calling {@link #getOutputParameterValue(Parameter)}.
+     * All parameters must be registered.
+     * @param position the parameter position
+     * @param type the type of the parameter
+     * @param mode the parameter mode
+     * @return the same query instance
+     * @since 4.0
+     */
+    <T> Parameter<T> registerParameter(int position, Class<T> type,
+                                       ParameterMode mode);
```
njr-11: There is a mismatch between the method return value `Parameter<T>` and what the Javadoc says is returned, "the same query instance".  The subsequent method has the same issue as well. And the prior method's Javadoc lacks `@return`.

gavinking: Ah, thanks, I will fix.

gavinking: Fixed now, thanks again.

njr-11 (approved):

gavinking:
Alright, great, thanks everyone who reviewed this. Next up is #162 😅

Issue #135, opened 2016-09-20
Unclear file handling for javax.persistence.schema-generation.scripts.drop-target and javax.persistence.schema-generation.scripts.create-target


lukasj:
It seems unclear in the JPA 2.1 specification whether files specified in the javax.persistence.schema-generation.scripts.drop-target and javax.persistence.schema-generation.scripts.create-target properties are expected to overwrite an existing version of such a file or to append to the file.

It seems that Hibernate has chosen to append to existing files resulting in a growing file if not deleted by other processes.
#### Affected Versions
[2.1.1]

m-reza-rahman:
This seems like low-hanging fruit that could be looked at in the next release and resolved one way or the other? It does appear to be a fairly low priority either way.

Reza Rahman
Jakarta EE Ambassador, Author, Blogger, Speaker

Please note views expressed here are my own as an individual community member and do not reflect the views of my employer.

gavinking:
I talked this over with @sebersole.

He doesn't like the idea of just overwriting files due to safety concerns.

I think we both would be OK with just failing fast in the case that a script target file already exists, but we don't think the spec really needs to be very prescriptive about this.

Therefore I've proposed #802.

Note that this is a *very* old issue with zero votes, and I don't think we've ever received complaints about this issue, in all these years.

gavinking:
We have "solved" this issue just by adding a NOTE warning that portable application should ensure that the generated script location is unoccupied.

gavinking:
Apparently I forgot to close this one.

PR #802, opened 2025-12-05
note about unoccupied script target locations

gavinking:
see #135

cc @sebersole


sebersole:
There is a (not necessarily common) case for *Hibernate users* where users collect create and update output in one file over time ("rolling").  I was worried about the wording here, but since schema update is not standardized and this is Hibernate-specific, I guess its not worth fussing over here.

So +1

gavinking:
Merging based on positive feedback from Steve.

PR #808, opened 2025-12-12
Review comment on possible enhancement with a return type for fetch(association)

njr-11:
An optional [review comment](https://github.com/jakartaee/persistence/pull/795#issuecomment-3647181286) that was deferred from #795

Usage of entityAgent.fetch(association) could be slightly more convenient in some cases if it returned the association.  See update to code example in method Javadoc.

If this isn't wanted, just close it.


gavinking:
Thanks.

PR #812, opened 2025-12-13
add some more useful info to the javadoc for @Entity

gavinking:
this information was not so easily discoverable

### DIFF
--- a/api/src/main/java/jakarta/persistence/Entity.java
+++ b/api/src/main/java/jakarta/persistence/Entity.java
@@ -90,8 +90,8 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * </ul>
  * <p>By default:
  * <ul>
- * <li>fields or properties of basic type, one-to-one associations, and
- *     many-to-one associations are fetched eagerly, but
+ * <li>fields or properties of basic or embedded type, one-to-one 
+ *     associations, and many-to-one associations are fetched eagerly, but
  * <li>element collections, one-to-many associations, and many-to-many
  *     associations are fetched lazily.
  * </ul>


gavinking:
Can someone find five minutes to proof-read this please? @sebersole @beikov @njr-11 🙏

sebersole:
Seems like an odd place to put this information.  I don't have a suggestion for a better place *on annotations*, but maybe at the root Javadoc page might be better?

gavinking:
> maybe at the root Javadoc page might be better

Maybe. But it's currently completely empty, so I wouldn't put *just* this information there on its own.

How about we make a plan to write proper package-level documentation at some point before the final release, and then consider moving this text there at that time if it fits?

I note that @njr-11 have been having a similar running debate in the Data spec; he is maintaining a huge list of information in the module descriptor that I *hate* because it's so difficult to navigate and maintain.

But when you go here and read the result, it's not so bad: https://jakarta.ee/specifications/data/1.0/apidocs/jakarta.data/module-summary.html

gavinking:
> How about we make a plan to write proper package-level documentation at some point before the final release

See #833

gavinking:
[That said, I don't think it's *that* strange of a place to put this info.]

api/src/main/java/jakarta/persistence/Entity.java
```diff
@@ -79,6 +79,43 @@
...
+ * <ul>
+ * <li>transparently on first access, if the entity is a managed instance
+ *     belonging to a persistence context, or
+ * <li>by calling the method {@link EntityAgent#fetch}, if the entity is
+ *     detached and not associated with any persistence context.
+ * </ul>
+ * <p>By default:
+ * <ul>
+ * <li>fields or properties of basic type, one-to-one associations, and
+ *     many-to-one associations are fetched eagerly, but
```
njr-11: We could also mention embeddables here

gavinking: Hrrrrm ... yeah I guess that makes sense. Depends how you look at it: are they fields or are they just containers? (They're both of course.)

gavinking:
```suggestion
 * <li>fields or properties of basic or embedded type, one-to-one 
 *     associations, and many-to-one associations are fetched eagerly, but
```

gavinking: Like this?

njr-11: That looks good. I was going back and forth on whether we needed to mention embedded explicitly. I like what you wrote up and it doesn't hurt to include mention of it, so I think we should.

gavinking:
Thanks folks.

PR #816, opened 2025-12-13
add a way to exclude a column from schema validation

### LINKED ISSUE #815: exclude column from schema validation
Sometimes users run into a case where the schema validation is too strict, and their program actually works as is; it would be really nice to given them a way to exclude a column from validation. Otherwise, the developer of the schema validation tool has to make some fairly arbitrary calls about how strict to be and about what to tolerate and what not to tolerate.

gavinking:
Sometimes users run into a case where the schema validation is too strict, and their program actually works as is; it's nice to have a way to exclude a column from validation.

Fixes #815.

WDYT?

### DIFF
--- a/api/src/main/java/jakarta/persistence/Column.java
+++ b/api/src/main/java/jakarta/persistence/Column.java
@@ -186,4 +186,15 @@ public @interface Column {
      * @since 3.2
      */
     String comment() default "";
+
+    /**
+     * Determines if this column is validated as part of the
+     * schema {@linkplain SchemaManager#validate validation}
+     * process. By default, the column is validated. If the
+     * column should be excluded from validation, specify
+     * {@code validated=false}.
+     *
+     * @since 4.0
+     */
+    boolean validated() default true;
 }
--- a/api/src/main/java/jakarta/persistence/DiscriminatorColumn.java
+++ b/api/src/main/java/jakarta/persistence/DiscriminatorColumn.java
@@ -99,4 +99,15 @@ public @interface DiscriminatorColumn {
      * @since 4.0
      */
     String comment() default "";
+
+    /**
+     * Determines if this column is validated as part of the
+     * schema {@linkplain SchemaManager#validate validation}
+     * process. By default, the column is validated. If the
+     * column should be excluded from validation, specify
+     * {@code validated=false}.
+     *
+     * @since 4.0
+     */
+    boolean validated() default true;
 }
--- a/api/src/main/java/jakarta/persistence/JoinColumn.java
+++ b/api/src/main/java/jakarta/persistence/JoinColumn.java
@@ -212,4 +212,15 @@ public @interface JoinColumn {
      * @since 3.2
      */
     String comment() default "";
+
+    /**
+     * Determines if this column is validated as part of the
+     * schema {@linkplain SchemaManager#validate validation}
+     * process. By default, the column is validated. If the
+     * column should be excluded from validation, specify
+     * {@code validated=false}.
+     *
+     * @since 4.0
+     */
+    boolean validated() default true;
 }
--- a/api/src/main/java/jakarta/persistence/MapKeyColumn.java
+++ b/api/src/main/java/jakarta/persistence/MapKeyColumn.java
@@ -171,4 +171,15 @@ public @interface MapKeyColumn {
      * @since 4.0
      */
     String comment() default "";
+
+    /**
+     * Determines if this column is validated as part of the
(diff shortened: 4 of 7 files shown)


api/src/main/java/jakarta/persistence/Column.java
```diff
@@ -186,4 +186,15 @@
      * @since 3.2
      */
     String comment() default "";
+
+    /**
+     * Determines if this column is validated as part of the
+     * schema {@linkplain SchemaManager#validate validation}
+     * process. By default, the column is validated. If the
+     * column should be excluded from validation, specify
+     * {@code validated=false}.
```
gavinking, line 195: There is an important ambiguity here. What does "excluded from validation" mean precisely? Does it mean we check that the column exists but perform no further validation beyond that? Or does it mean we don't even validate that it exists in the table? 🤔

yrodiere: I can't imagine a situation where one would want their program to run against a database where one of the column doesn't exists... ?

The original use case was to workaround limitations in validation type-checking. So I think we want that:

> we check that the column exists but perform no further validation beyond that?

api/src/main/java/jakarta/persistence/Column.java
yrodiere, line 1: Doesn't validation also make sense for things that are not related to columns?

I couldn't find a precise definition of schema validation in the JPA spec, so I suppose it's up to the implementation to pick what gets validated.

I can see that there are a few more things than columns which could impact the schema:

* `@Table.uniqueConstraints`
* `@Table.indexes`
* `@Table.check`

And maybe more but the rest seems... unlikely to be validated.

Maybe `@Table.validated` would make some sort of sense?

api/src/main/java/jakarta/persistence/Column.java
gavinking: Yeah, that's probably right.

api/src/main/java/jakarta/persistence/Column.java
gavinking: I mean, yeah, _in principle_, but I'm cautious about adding features to the spec for some purely-speculative feature that some JPA implementation _might_ have at some unspecified point in the future.

So unless we're making a decision to start validating the existence of constraints, I wouldn't add this.

That's not to say I don't think constraint validation would be useful. Just we don't have any implementation of it yet.

yrodiere: The other side of the coin is that if we don't add it now, then when an implementation decides to implement such validation (since that's up to the implementation), there will likely be a delay of months, likely years, before the corresponding "validation" exclusion is made available for tables in the same way it is available for columns. Leading to either a gap in functionality, or the temporary introduction of native annotations -- which, for `@Table` at least, we've been trying to get rid of.

One way to circumvent your position would be to decree that `validated = false` on a table means both the table's constraints *and its columns* don't get validated (regardless of what's configured at the column level). Then the attribute would have a use *right now*.

Otherwise, okay, if it's a conscious decision, fair enough.

gavinking: I guess I can sorta see that `validated=false` is a reasonably natural way to say "don't validate the existence of columns". (Though we agree that's not especially useful.) 

I'm not sure it's an especially good way to indicate that, say, a foreign key or unique key constraint should not be validated. I feel like that's something you would want to be able to set *per-constraint*, not once for the whole table. 

Also note that `@ForeignKey` actually already has something pretty close to that with `@ForeignKey(NO_CONSTRAINT)`. Ignore the particular syntax, which I've never liked.

gavinking:
```java
 * @since 2.1
 */
public enum ConstraintMode {
```

`@since 2.1` is this massive red flag in the persistence APIs. 😭

yrodiere: > I feel like that's something you would want to be able to set _per-constraint_, not once for the whole table.

True, but then the whole feature is a workaround to deal with false-positives of validation. Disabling validation of a specific constraint would be best, but failing that disabling validation of the whole table still has benefits (you can validate other tables).

yrodiere: > Maybe `@Table.validated` would make some sort of sense?

FWIW I suspect @maxandersen's use case with sqlite's FTS5 feature would benefit from that: https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/hibernate.20control.20over.20hbm2ddl.3F/with/622495316

PR #818, opened 2025-12-13
comment and check for @MapKeyColumn, @MapKeyJoinColumn, and @PrimaryKeyJoinColumn

### LINKED ISSUE #817: comment and check for @MapKeyColumn, @MapKeyJoinColumn, and @PrimaryKeyJoinColumn
Apparently when we added `column` and `check` to the different flavors of column annotation, `@MapKeyColumn`, `@MapKeyJoinColumn`, and `@PrimaryKeyJoinColumn` missed out. I'm wracking my head trying to imagine a reason why this was intentional but I just can't think of anything. I've gotta conclude I just _forgot about them_, which is worrying in itself.

Anyway, it seems to me that we should add them??

gavinking:
Fixes #817

### DIFF
--- a/api/src/main/java/jakarta/persistence/OrderColumn.java
+++ b/api/src/main/java/jakarta/persistence/OrderColumn.java
@@ -109,4 +109,12 @@ public @interface OrderColumn {
      * @since 3.2
      */
     String options() default "";
+
+    /**
+     * (Optional) A comment to be applied to the column.
+     * This is only used if table generation is in effect.
+     *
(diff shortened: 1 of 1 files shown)


gavinking:
I can't imagine anyone objecting to these changes. So I think they can be quickly merged.

gavinking:
> I can't imagine anyone objecting to these changes. So I think they can be quickly merged.

Going to merge this tomorrow, unless someone objects.

Issue #105, opened 2015-03-26
Allow AttributeConverter to be used for composite types by using @Embeddable


lukasj:
Right now, AttributeConverters can only be used for single-value types. They could be extended to use an @Embeddable type as a surrogate column type for composite types.

Imagine the immutable type Money(final BigDecimal amount, final Currency currency) and the following @Embeddable type:

```java
@Embeddable
public class EmbeddableMoney {

	private BigDecimal amount;
	private Currency currency;

	// getters & setters
 }
```
```java
@Converter(autoApply = true)
public class MoneyConverter implements AttributeConverter<Money, EmbeddableMoney> {

	@Override
	public EmbeddableMoney convertToDatabaseColumn(Money attribute) {
		if (attribute == null) {
			return null;
		}
		EmbeddableMoney columns = new EmbeddableMoney();
		columns.setAmount(attribute.getAmount());
		columns.setCurrency(attribute.getCurrency());
		return columns;
	}

	@Override
	public Money convertToEntityAttribute(EmbeddableMoney dbData) {
		if (dbData == null) { // check if empty? 			return null;
		}
		return new Money(dbData.getAmount(), dbData.getCurrency());
	}
}
```
```java
@Converter(autoApply = true)
public class CurrencyConverter implements AttributeConverter<Currency, String> {
	...
}
```

What do you think?

kalgon:
Another example of persistence which can use intermediate/surrogate objects is java standard serialization with `readResolve()` and `writeReplace()`.

kalgon:
With java 14 records, this feature could prove even more useful.

jwgmeligmeyling:
I agree. Embeddables are not always a viable option if you want to program against an interface from an external module (i.e. Java Money). A solution that is commonly used for this is composite user types (or multi column types). These however are going to be removed in Hibernate, so that option will disappear. I agree that being able to use attribute converters on embedded attributes would be a neat solution to this problem.

And I don't think it will be possible to add support for `javax.money.MonetaryAmount` directly, because that would involve the introduction of multi-column bindings in JPA.

beikov:
I don't think conversion would make much sense. JPA has no idea what the model structure of the converted type is, so you'd have no way of accessing individual properties in a JPQL query. IMO, there should be a way in JPA to define the structure of types programmatically or through XML mapping and a mechanism that allows to control the construction of objects of that type.

kalgon:
[...]

Issue #73, opened 2014-02-07
Parameterized AttributeConverter and/or AttributeConverter metadata access


lukasj:
It would be good if we were able to parameterize AttributeConverter instances and/or have access to property metadata within an AttributeConverter. This way we are able to reuse AttributeConverter code instead of building similar ones or even supply further required context information. An example:

Quite often we have to deal with ancient DB schemas, and just recently we started to migrate a large Cobol application to Java. We are unable to change the content of the database, and because of Cobol with have to support fixed width string columns containing left padded data. A 5 digit fixed width column will represent 1 as '00001' and 100 as '00100'. Nothing we can change here. Unless I am mistaken I have to create a dedicated converter for every required fixed length ending up with an LeadingZeroTwoDigitAttributeConverter, LeadingZeroFiveDigitAttributeConverter and so on.

Right now it seems the JPA spec does not define wether there is one converter instance per persistent property or just a global one (which would be OK by current spec requirements). I would propose that parameterized converter instances are bound to property fields in order to support parameter evaluation during JPA provider startup.

I believe we need access to both basic attribute information and optionally supplied converter attributes. The following example would assume access to the leading pad char, the length attribute and wether the attribute is nullable or not. (in my current project I have to store an empty string for null values in the database).

@Convert(converter = LeadingDigitAttributeConverter.class, metaData="padChar=0" )
@Column(name = "ACOLUM", length=5, nullable = false)

If you believe there is no need to support something like that: Unfortunately we are in the middle of migrating quite a few very very old applications to Java, and we can't change that stuff. And there is more to come.

gavinking:
```java
@MyAnnotation(padChar = "0")
@Convert(converter = LeadingDigitAttributeConverter.class)
private String myField;
```

This code example would be _much_ improved if the `@Convert` annotation could be applied as a meta-annotation of `@MyAnnotation`. Jus' sayin'. :)

gavinking:
> Jus' sayin'. :)

We could definitely take the same approach to this one that we took with `UserType` in Hibernate:

1. allow `@Convert` to be used as a meta-annotation, and
2. pass the annotation instance it annotates to the constructor of the `AttributeConverter` if there is one.

Issue #43, opened 2013-01-17
Allow type level annotations to be used as meta-annotations


lukasj:
I repeatedly find myself annotating my JPA entities with the very same set of annotations:

```
@Entity
@EntityListeners(AuditingEntityListener.class)
class Person {

}
```

If both <tt>@Entity</tt> and <tt>@EntityListener</tt> were allowed to be used as meta-annotations I could collapse them into:

```
@Target(TYPE)
@Retention(RUNTIME)
@Entity
@EntityListeners(AuditingEntityListener.class)
@interface @AuditedEntity {

}
```

Resulting in:

```
@AuditedEntity
class Person {

}
```

This is in line with the meta-annotation handling CDI exposes to introduce annotation with richer semantics in annotation code. The following changes would be required.

*   Add <tt>ElementType.ANNOTATION_TYPE</tt> to the relevant annotations
*   Specify that persistence providers have to evaluate the annotations from the meta-level as well using the first one found, so that locally defined annotations would be considered first.

michaelisvy:
hi,
I also wanted to add that it would be great to have `field-level` meta-annotation support. 

For instance, every time I create a new JPA entity, I need to declare the id as follows:
```java
@Id
@GeneratedValue(strategy = GenerationType.IDENTITY)
private Long id;
```
It would be nice to be able to create a single annotation that embeds both annotations. I would use it that way:
```java
@MyCompanyId
private Long id;
```
and declare the annotation as follows:
```java
@Target({ElementType.ANNOTATION_TYPE, ElementType.FIELD, ElementType.METHOD})
@Retention(RetentionPolicy.RUNTIME)
@Id
@GeneratedValue(strategy = GenerationType.IDENTITY)
public @interface MyCompanyId {
}
```

For it to be possible, `@Id` and `@GeneratedValue` annotations for the `jakarta.persistence` package would need to be updated so they include `ElementType.ANNOTATION_TYPE` as follows:

```java
@Target({METHOD, FIELD,ANNOTATION_TYPE})
@Retention(RUNTIME)
public @interface Id {}
```

Would it make sense?

darioseidl:
How is the status of this? I read that that meta-annotations are supported since JPA 2.2. Are all JPA annotations (class and field level) supported in meta-annotations now?

emivoz:
This wan is open, since it was planned for JAVA EE 8. Any plans, updates on this?

gavinking:
[...]

PR #819, opened 2025-12-13
add @Convert(to = ....class)

### LINKED ISSUE #619: consider adding JDBC type to @Column or a zoo of built-in AttributeConverters
Currently there are two ways to customize the column type mapped by a field:

1. using an `AttributeConverter`, or
2. using `@Column(columnDefinition)`.

An `AttributeConverter` works very well for basic types, but there are no built-in `AttributeConverter`s, and it's a pain for the user to have to write one. And it doesn't work at all for associations.

On the other hand `columnDefinition` doesn't work very well, for several reasons, including that it's not at all portable between databases, and also doesn't do any sort of type conversion at the Java level.

First approach
---

Since very ancient times, Hibernate allowed the column type to be customized in a portable way by specifying a JDBC type code from the list given in `SqlTypes`. Today this is usually done via the `@JdbcTypeCode` annotation.

The only thing about this approach that has always left me a little uncomfortable is the lack of typesafety, due to the `SqlTypes` being integers. A way to alleviate this discomfort would be to use the `JDBCType` enumeration instead. So we could add:

```java
public @interface Column {
    ...
    JDBCType type() default JDBCType.OTHER;
}
```

where we would be abusing `OTHER` to mean "default".

So I could write:

```java
@Column(type=JDBCType.DECIMAL)
int total;
```

The spec would need to define which combinations of (field type, JDBC type) an implementation is required to support, but just like with the JPQL `cast()` function, that list need not be extremely long. That is, it could, at least initially, just be basic numeric conversions and conversions to `varchar`.

Second approach
---

Alternatively, if this abuse of `OTHER` is unacceptable, we could add a new annotation:

```java
public @interface ColumnType {
    JDBCType value();
}
```

which might anyway be a cleaner approach.

I would write:

```java
@ColumnType(JDBCType.DECIMAL)
int total;
```

Third approach
---

Alternatively, the specification could provide a set of built-in attribute converter classes, for use with `@Convert`. The problem is that to provide converters for _n_ Java types to _m_ JDBC types requires _nm_ classes, which gets a bit messy.

It would look like:

```java
@Convert(converter=Converters.IntegerToDecimal.class)
int total;
```

which is a bit more verbose than the previous options.

The advantage of this approach is that it requires no new concept in the spec.

gavinking:
see #619

### DIFF
--- a/api/src/main/java/jakarta/persistence/Convert.java
+++ b/api/src/main/java/jakarta/persistence/Convert.java
@@ -280,7 +280,7 @@ public @interface Convert {
    *     {@link Double}, {@link Long}, {@code double}, or {@code long},
    *     and between {@link java.math.BigInteger} and {@link Long}
    *     or {@code long}, and
-   * <li>conversion from {@link Boolean} or {@code boolean} to any
+   * <li>conversion between {@link Boolean} or {@code boolean} and any
    *     integral numeric primitive or wrapper type, where {@code false}
    *     or {@code FALSE} is converted to {@code 0} and {@code true} or
    *     {@code TRUE} is converted to {@code 1}.
--- a/api/src/main/java/jakarta/persistence/Convert.java
+++ b/api/src/main/java/jakarta/persistence/Convert.java
@@ -208,6 +208,14 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * @Column(precision = 10, scale = 4)
  * // maps to column of type DECIMAL(10,4)
  * double magnitude;
+ *
+ * @Convert(to = long.class)
+ * // maps to column of type BIGINT for extra headroom
+ * int count;
+ *
+ * @Convert(to = byte.class)
+ * // maps to column of type TINYINT in legacy database
+ * boolean enabled;
  * }
  *
  * @see Converter
@@ -264,11 +272,15 @@ public @interface Convert {
    * <li>all conversions defined by {@code toString} and {@code parse}
    *     methods (for example, {@link Float#toString(float)} and
    *     {@link Float#parseFloat(String)}) between primitive or wrapped
-   *     numeric types and {@link String}, and
+   *     numeric types and {@link String},
    * <li>conversions between {@link java.math.BigDecimal} and
    *     {@link Double}, {@link Long}, {@code double}, or {@code long},
    *     and between {@link java.math.BigInteger} and {@link Long}
-   *     or {@code long}.
+   *     or {@code long}, and
+   * <li>conversion from {@link Boolean} or {@code boolean} to any
+   *     integral numeric primitive or wrapper type, where {@code false}
+   *     or {@code FALSE} is converted to {@code 0} and {@code true} or
+   *     {@code TRUE} is converted to {@code 1}.
    * </ul>
    * <p>A provider is encouraged to support additional vendor-specific
    * built-in conversions, but such conversions are not portable between
--- a/api/src/main/java/jakarta/persistence/Convert.java
+++ b/api/src/main/java/jakarta/persistence/Convert.java
@@ -238,7 +238,7 @@ public @interface Convert {
 
   /**
    * Specifies a basic type for the conversion. This element may be
-   * specified if the persistence provider feaures built-in support
+   * specified if the persistence provider features built-in support
    * for converting between the specified type and:
    * <ul>
    * <li>the declared type of the annotated program element, in the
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/Convert.java
```diff
@@ -219,6 +236,61 @@
...
+   * </ul>
+   * <p>When reading a value from the database:
+   * <ul>
+   * <li>the built-in conversion is first applied to the value read
+   *     from the database, and then
+   * <li>the resulting converted value is passed to the converter.
+   * </ul>
+   *
+   * <p>The Persistence provider is required to support the following
+   * built-in conversions:
```
yrodiere, line 271:
```suggestion
   * built-in conversions, in both directions:
```

I guess that's implicit, but given we're using `to = SomeClass.class` for something that will actually involve converting to `SomeClass` (when writing) and *from* `SomeClass` (when reading), that's maybe worth clarifying... ?

api/src/main/java/jakarta/persistence/Convert.java
```diff
@@ -219,6 +236,61 @@
...
+   *     {@link Float#parseFloat(String)}) between primitive or wrapped
+   *     numeric types and {@link String},
+   * <li>conversions between {@link java.math.BigDecimal} and
+   *     {@link Double}, {@link Long}, {@code double}, or {@code long},
+   *     and between {@link java.math.BigInteger} and {@link Long}
+   *     or {@code long}, and
+   * <li>conversion from {@link Boolean} or {@code boolean} to any
+   *     integral numeric primitive or wrapper type, where {@code false}
+   *     or {@code FALSE} is converted to {@code 0} and {@code true} or
+   *     {@code TRUE} is converted to {@code 1}.
```
yrodiere:
```suggestion
   * <li>conversion between {@link Boolean} or {@code boolean} and any
   *     integral numeric primitive or wrapper type, where {@code false}
   *     or {@code FALSE} is converted to {@code 0} and {@code true} or
   *     {@code TRUE} is converted to {@code 1}.
```

Unless you really meant `@Convert(to = boolean.class) int foo` is not legal?

api/src/main/java/jakarta/persistence/Convert.java
```diff
@@ -219,6 +236,61 @@
...
+   * </ul>
+   * <p>A provider is encouraged to support additional vendor-specific
+   * built-in conversions, but such conversions are not portable between
+   * providers and should not be relied upon by portable application
+   * programs.
+   *
+   * @since 4.0
+   */
+  Class<?> to() default void.class;
```
yrodiere, line 295: I wonder if something like `as()` wouldn't be less confusing here... because this is fundamentally not a one-way conversion.

But then `as` is a keyword in Groovy, so there's that.

gavinking: I think if it were `@Persist` or `@Treat` then `as` would be better. But "convert as string" isn't grammatically correct.

api/src/main/java/jakarta/persistence/Convert.java
gavinking: Does b6ed448d61302d10869a3dc570ef0ac69795a193 help?

yrodiere: Yes thank you. I'm not allowed to mark this as resolved, but it is.

yrodiere (approved):

gavinking:
If we're going to touch this stuff, we should take the opportunity to bless the use of `@Convert` as a meta-annotation, as discussed in #43.

gavinking:
@sebersole Are you in favor of doing this one in JPA4?

Or should we put it on ice for reconsideration as part of 4.1, where we're already considering other longstanding enhancements to converters (composite converters #105, and `@Convert` as a meta-annotation #73)?

sebersole:
Like we discussed in chat, I'd be inclined to (also?) expose this on `@Column`.  E.g.

```java
@Column(type=String.class)
UUID uuid;
```

Which, granted, implies some level of implicit conversion.  But you already called that out, and I don't think needs to be a huge set for the spec itself - providers could ofc provide extended set of conversions if they wish

gavinking:
Yeah so I'm trying to think this through. 

- There's something appealing about putting it on `@Column` because it really is just the column type you're specifying, and to the extend that there's a converter, it's entirely implicit and theoretical.
- On the other hand, it feels at least a bit strange to me to be specifying a column type as a Java type., and I feel like this sorta pushes me back in the direction of the place I actually started out at which was `@Column(jdbcType=Types.BIGINT)`.

PR #820, opened 2025-12-14
alternatives to ManagedEntityMode 

gavinking:
An alternative solution to #62, since I'm afraid that #668 might be getting out of hand.

### DIFF
--- a/api/src/main/java/jakarta/persistence/FlushModeType.java
+++ b/api/src/main/java/jakarta/persistence/FlushModeType.java
@@ -58,10 +58,11 @@ package jakarta.persistence;
  */
 public enum FlushModeType {
     /**
-     * Every flush is an explicit operation requested by the application
-     * program. The session is never automatically flushed. Modifications
-     * to entities held in the persistence context might not be visible
-     * to the processing of queries and might never be made persistent.
+     * Every flush is an explicit operation requested by the
+     * application program. The session is never automatically
+     * flushed. Pending modifications to entities held in the
+     * persistence context might not be visible to the processing
+     * of queries and might never be made persistent.
      *
      * @since 4.0
      */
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/ReadOnly.java
```diff
@@ -0,0 +1,65 @@
...
+ * in a significant cost to performance.
+ *
+ * @see EntityManager#enableFlush(Object)
+ *
+ * @since 4.0
+ */
+@Documented
+@Target(TYPE)
+@Retention(RUNTIME)
+public @interface ReadOnly {
```
beikov, line 64: Given this definition, I think the name `@ReadOnly` is odd since you can insert and remove entries. IMO `@Immutable` like we have in Hibernate ORM is a better name for this. It might be worth adding a section about inheritance i.e. polymorphic subclasses inherit the behavior.
Additionally, it might be interesting to allow the annotation on `FIELD`/`METHOD` to control which persistent attributes contribute to dirty checking, similar to what we define in Hibernate ORM already, which would allow us to get rid of the Hibernate ORM specific annotation.

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -793,6 +793,40 @@ void refresh(Object entity,
...
+     *               class
+     * @throws IllegalArgumentException if the instance is
+     *         not associated with this persistence context
+     *         or is not a non-null instance of a read-only
+     *         entity class
+     * @since 4.0
+     * @see ReadOnly
+     * @see #flush
+     */
+    void enableFlush(Object entity);
```
beikov, line 869: AFAIU, in Hibernate ORM it's impossible to make `@Immutable` entities non-read-only, so I would rather not do this. Also see `org.hibernate.engine.internal.AbstractEntityEntry#setReadOnly`.

api/src/main/java/jakarta/persistence/ReadOnly.java
gavinking: > I think the name @ReadOnly is odd since you can insert and remove entries.

I .... guess I don't find that very odd?

It's perhaps slightly odd to say that you can _delete_ a read-only thing, but I don't see how it's strange to say that you can create one.

>  IMO `@Immutable` like we have in Hibernate ORM is a better name for this.

"Immutable" to me implies a *much* higher level of un-modifiability than "read only". When I say something is "immutable", I usually mean in some sort of Platonic sense. Not just in the sense of "I'm not interested in modifying this thing".

api/src/main/java/jakarta/persistence/EntityManager.java
gavinking: What exists in Hibernate today grew organically and it shows. I don't think it's an especially perfect model.

api/src/main/java/jakarta/persistence/ReadOnly.java
gavinking: By the way, I consdered calling it `@Unflushable`, and I think maybe that's better.

Issue #821, opened 2025-12-14
Add fetch hints to EntityGraph


AleksNo:
EntityGraphs offer the description what data should be fetched. But if one wants to specify how the data should be fetched then the things start to become really messy. Either there are vendor specific annotations on entities or query hints or settings in the persistence.xml file etc. 
And unfortunately many of these options are far away from the actual bussiness logic.

My idea is to extend EntityGraphs in a way so they can take hints how the data could be fetched. 

For the EntityGraph interface there would be a new method:

```java
<X> EntityGraph <X> setFetchHints(Map<String, Object>)
```
for the SubGraph interface:

```java
<X> SubGraph <X> setFetchHints(Map<String, Object>)
```

and for the AttributeNode interface:

```java
void setFetchHints(Map<String, Object>)
````

This would allow to write EntityGraphs like this:

```java
EntityGraph<MyParentEntity> entityGraph = entityManager.createEntityGraph(MyParentEntity.class);

entityGraph.setFetchHints(Map.of("fetchType", "exists")).addSubgraph(MyParentEntity_.myChildEntity)
.setFetchHints(Map.of("fetchType", "in", "batchSize", "200")).addAttributeNodes(MyChildEntity_.otherEntity)
.setFetchHints(Map.of("fetchType", "in", "batchSize", "200"));

List<MyParentEntity> entityList = entityManager.findMultiple(entityGraph, List.of(1,2,3,4,5));
```

The benefit would be, more control how the data would be fetched. And fetching and bussiness logic would be very close together.

What do you think?

gavinking:
I agree that the inability to select between fetching strategies is a big limitation, and it's good to have an issue to track that.

On the other hand, I'm not at all a fan of passing strings around. If we did something like this, I would look to `FindOption` for inspiration: that is, the spec could have a `FetchOption` interface, and implementations could define vendor-specific options like `new BatchSize(200)`, `FetchType.SUBSELECT`, and so on.

gavinking:
So this looks suprisingly straightforward; we would add:

```java
public interface AttributeNode<T> {

    ...

    interface FetchOption {}

    AttributeNode<T> addOption(FetchOption option);

}
```

And its usage would look something like this:

```java
var graph = em.createEntityGraph(Book.class);
graph.addAttributeNode(Book_.publisher)
        .addOption(CacheRetrieveMode.BYPASS)
        .addOption(FetchMode.SELECT)
        .addOption(new BatchSize(200));
[...]

PR #822, opened 2025-12-14
add FetchOption and AttributeNode.setOption()

gavinking:
for #821

### DIFF
--- a/api/src/main/java/jakarta/persistence/CacheRetrieveMode.java
+++ b/api/src/main/java/jakarta/persistence/CacheRetrieveMode.java
@@ -32,7 +32,7 @@ package jakarta.persistence;
  * {@code jakarta.persistence.cache.retrieveMode}.
  *
  * @see EntityHandler#setCacheRetrieveMode(CacheRetrieveMode)
- * @see Query#setCacheRetrieveMode(CacheRetrieveMode)
+ * @see TypedQuery#setCacheRetrieveMode(CacheRetrieveMode)
  *
  * @since 2.0
  */
@@ -49,6 +49,10 @@ public enum CacheRetrieveMode implements FindOption, FetchOption {
     /**
      * Specifies that data must not be read from the second-level
      * cache and must be read from the database.
+     *
+     * @apiNote When used as a {@link FetchOption}, this value is
+     *          a hint to the persistence provider, and might be
+     *          ignored.
      */
     BYPASS
 }
--- a/api/src/main/java/jakarta/persistence/CacheStoreMode.java
+++ b/api/src/main/java/jakarta/persistence/CacheStoreMode.java
@@ -34,11 +34,11 @@ package jakarta.persistence;
  * {@code jakarta.persistence.cache.storeMode}.
  *
  * @see EntityHandler#setCacheStoreMode(CacheStoreMode)
- * @see Query#setCacheStoreMode(CacheStoreMode)
+ * @see TypedQuery#setCacheStoreMode(CacheStoreMode)
  *
  * @since 2.0
  */
-public enum CacheStoreMode implements FindOption, RefreshOption {
+public enum CacheStoreMode implements FindOption, RefreshOption, FetchOption {
 
     /**
      * Specifies that entity data may be inserted into the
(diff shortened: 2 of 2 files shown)


gavinking:
Wondering if we should make `FetchType` a `FetchOption`. That would be a bit of a change to the model though. We would then have `LAZY` nodes in the graph instead of just missing nodes. Not sure how I feel about that.

gavinking:
I believe it's time to get this one moving. @sebersole @njr-11 @anija-anil @AleksNo please review 🙏

gavinking:
Question (for @sebersole) ... should `CacheStoreMode` also be a `FetchOption`?

sebersole:
> We would then have LAZY nodes in the graph instead of just missing nodes. Not sure how I feel about that.

I think its just like `removeAttributeNode` in this regard.  Functionally you have to keep those nodes around (in some form) to be able to distinguish between missing and removed since that is an important distinction for load-graphs.  Missing just means "do the default thing".

sebersole:
> Question (for @sebersole) ... should `CacheStoreMode` also be a `FetchOption`?

Selfishly I'd say no since Hibernate does not operate that way.  Not saying its a bad idea, it does make sense in fact.

gavinking:
> I think its just like `removeAttributeNode` in this regard. Functionally you have to keep those nodes around (in some form) to be able to distinguish between missing and removed since that is an important distinction for load-graphs. Missing just means "do the default thing".

Right, that's what I've proposed in the linked PR.

sebersole:
> > Question (for @sebersole) ... should `CacheStoreMode` also be a `FetchOption`?
> 
> Selfishly I'd say no since Hibernate does not operate that way. Not saying its a bad idea, it does make sense in fact.

Like I said, this is something Hibernate does not currently support; but I think it logically makes sense and it is something I would be interested in implementing.  But ofc timing.

So either 

1. say that (some) options are "hints" and treated as such (aka optional)
2. don't make `CacheStoreMode` implement `FetchOption`

If other providers *do* support specificying different `CacheStoreModes` for different parts of a result graph then I'd hate to block support for this from them.  So my preference would be (1) though its an awkward precedence.

api/src/main/java/jakarta/persistence/AttributeNode.java
```diff
@@ -49,6 +50,26 @@ public interface AttributeNode<T> {
      */
     Attribute<?,T> getAttribute();
 
+    /**
+     * Add a subgraph rooted at this attribute node, which
+     * must be an embedded attribute or an association to
+     * an entity class.
+     * @see EntityGraph#addSubgraph(Attribute)
```
njr-11, line 58:
```suggestion
     * @return the added subgraph.
     * @see EntityGraph#addSubgraph(Attribute) 
```

api/src/main/java/jakarta/persistence/AttributeNode.java
```diff
@@ -49,6 +50,26 @@ public interface AttributeNode<T> {
...
+     */
+    Subgraph<T> addSubgraph();
+
+    /**
+     * Add a subgraph rooted at this attribute node, which
+     * must be an association to an entity class, for an
+     * entity subclass of the entity class.
+     * @param type The subclass of the entity class
+     * @param <S> The type of the subclass
+     * @see EntityGraph#addTreatedSubgraph(Attribute, Class)
```
njr-11, line 70:
```suggestion
     * @return the added subgraph.
     * @see EntityGraph#addTreatedSubgraph(Attribute, Class)
```

api/src/main/java/jakarta/persistence/FetchOption.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
njr-11:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

gavinking:
thanks, @njr-11

gavinking:
> say that (some) options are "hints" and treated as such

@sebersole As discussed, I've said that `CacheStoreMode` and `CacheRetrieveMode` are hints when used as `FetchOption`s. This makes sense since:

1. this is not something implementations currently support, and we might in principle run into corner cases where these hints are tricky to implement.,
2. we don't need this particular feature to be something that sits on the critical path to producing a compatible implementation, and
3. anyway, second-level caching in general has semantics which are already highly vendor-specific.

gavinking:
I believe we already have consensus on this one, so I'm pulling the trigger now.

Issue #420, opened 2023-05-30
streamline use of 'select new' with record types


gavinking:
I would like to propose that we let people write:

```java
record BookSummary(String title, String author) {}
em.createQuery("select book.title, book.author.name from Book book", BookSummary.class).getResultList();
```

instead of forcing them into the extra ceremony of:

```java
record BookSummary(String title, String author) {}
em.createQuery("select new BookSummary(book.title, book.author.name) from Book book", BookSummary.class).getResultList();
```

This isn't critical, of course, but it's a pretty easy simplification to the user experience.

gavinking:
Yes, sure, it would also be allowed for non-records. Sorry for being unclear.

trajano:
Nothing to be sorry about @gavinking  I presume the purpose of this repo is to discuss things to make sure things are clear and less *behaviour is unspecified and vendor dependent*

gavinking:
Another thing to consider as part of this is to just do certain sorts of implicit type conversions, for example, allow:

```java
int count = em.createSelectionQuery("select count(this) from Book", int.class).getSingleResult();
```

which would currently throw because `count(this)` is of type `Long`.

njr-11:
This would be nice to add to Jakarta EE 12 where Jakarta Data is adding the ability to select multiple entity attributes and retrieve them as a Java record.  With this feature from Jakarta Persistence, a Jakarta Persistence-backed Jakarta Data provider would not need to edit the SELECT clause to insert the `NEW org.example.MyRecord(` part and could just use the original query without modifying it.

gavinking:
@njr-11 Agreed.

gavinking:
I had almost forgotten about this one, but it's important for Jakarta Data. Need to start work on it.

gavinking:
I finally got around to writing this up.

> Also need to think about allowing this with `createNativeQuery()`.

I'm thinking of leaving that out for now.

gavinking:
This was done by #824.

PR #823, opened 2025-12-14
streamlined alternative to 'select new'

gavinking:
for #420

I have been promising everyone this for much more than 2 years now.

Note that this is a "minimal" option where we don't say that this works for native queries or stored procedure queries.

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -1435,8 +1435,7 @@ lifecycle operations:
 * `@PostPersist`
 * `@PreRemove`
 * `@PostRemove`
-* `@PreUpdate`
-* `@PostUpdate`
+* `@PreMerge`
 
 Annotations for callbacks corresponding to database operations:
 
@@ -1480,24 +1479,48 @@ An entity or entity listener may have multiple lifecycle callback methods:
 
 ==== Semantics of the Life Cycle Callback Methods for Entities [[a2202]]
 
-The `@PrePersist` and `@PreRemove` callbacks occur before the
-respective persist and remove operations for an entity are executed.
-For entities to which the `merge()` operation has been applied and
-causes the creation of newly managed instances, the `@PrePersist`
-callback occurs for the managed instance after the entity state has
-been copied to it. The `@PrePersist` and `@PreRemove` callbacks also
-occur for any entity to which these operations cascade. The `@PrePersist`
-and `@PreRemove` methods are always invoked as part of the synchronous
-`persist()`, `merge()`, and `remove()` operations of `EntityManager`.
+A `@PrePersist`, `@PreRemove`, or `@PreMerge` callback occurs before
+the respective persist, remove, or merge operation for an entity is
+executed.
 
-The `@PostPersist` and `@PostRemove` callbacks occur after an entity
-has been made persistent, or after it has been removed, respectively.
-These callbacks also occur for any entity to which these operations
-cascade. The `@PostPersist` and `@PostRemove` methods occur after the
-database insert and delete operations, respectively. The database
-operations might occur directly after invocation of `persist()`,
-`merge()`, or `remove()`, or they might occur during a `flush()`
-operation--which may not happen until the end of the transaction.
+- The `@PrePersist` callback for an entity being made persistent
+  occurs before the entity is marked as managed and associated with
+  the persistence context.
+- The `@PreRemove` callback occurs before the entity being removed
+  is marked as removed in the persistence context.
+- The `@PreMerge` callback for an entity being merged occurs before
+  state is copied from the entity being merged to the corresponding
+  managed entity, when copying is necessary.
+- When the `merge()` operation results in the creation of a new
+  managed instance, the `@PrePersist` callback for the managed
+  instance occurs after the merged entity state has been copied to
+  it.
+
+A `@PrePersist`, `@PreRemove`, or `@PreMerge` method is always
+invoked synchronously during execution of the `persist()`, `remove()`,
+or `merge()` operation of the `EntityManager`.
+
+A `@PostPersist` or `@PostRemove` callback occurs after the effect
+of a respective persist or remove operation for an entity has been
+realized.
+
+- The `@PostPersist` callback for an entity which has been made
+  persistent occurs after the entity has been marked as managed
+  and associated with the persistence context, and after the
+  database insert operations have been executed.
+- The `@PostRemove` callback for an entity which has been removed
+  occurs after the entity has been marked as removed in the
+  persistence context, and after the database delete operations
+  have been executed.
+
+The database insert or delete operations might occur immediately
+during execution of `persist()`, `merge()`, or `remove()`, or they
+might occur later when the persistence context is flushed. The
+persistence context might not be flushed until the end of the
+transaction, or, if the transaction rolls back, it might not be
+flushed at all. If the persistence context is never flushed, and
+as a result the insert or delete operations are never executed,
+the `@PostPersist` or `@PostRemove` callback never occurs.
 
 Generated primary key values might not be available during a `@PrePersist`
 callback.
@@ -1510,12 +1533,18 @@ callback.
 - Generated primary key values are always available in the `@PostPersist`
   method.
 
+When a persist, remove, or merge operation cascades to an associated
+entity, the corresponding `@PrePersist`, `@PreRemove`, or `@PreMerge`,
+and `@PostPersist` or `@PostRemove` callbacks also occur for the associated
+entity.
+
 The `@PreInsert`, `@PostInsert`, `@PreUpdate`, `@PostUpdate`, `@PreDelete`
 and `@PostDelete` callbacks occur before and after database insert, update,
 and delete operations to entity data, respectively. These callbacks occur
 whenever entity state is modified via `EntityManager` or `EntityAgent`,
 and are not tied to the lifecycle of managed entities associated with a
-persistence context.
+persistence context. These callbacks are not triggered when entity data is
+modified by execution of a `Query`.
 
 [NOTE]
 ====
@@ -1526,11 +1555,17 @@ removed within a single transaction. Portable applications should not rely
 on such behavior.
 ====
 
-The `@PostLoad` callback occurs after an entity instance has been loaded
-from the database by any means, or after the `refresh()` operation has
-been applied to an entity instance. The `@PostLoad` method is invoked
-before the newly-loaded query entity is returned or made accessible to
-the application.
+The `@PostLoad` callback occurs:
+
+- after an entity instance has been loaded from the database by any means,
+  or
+- after the `refresh()` operation of `EntityManager` or `EntityAgent` has
+  been applied to an entity instance.
+
+The non-relationship state of an entity is always set to the values just
+retrieved from the database before the `@PostLoad` method is invoked. The
+`@PostLoad` method is invoked before a newly loaded entity is returned
+or made accessible to the application.
 
 [NOTE]
 This specification does not define whether callback methods are invoked
@@ -1779,8 +1814,9 @@ lifecycle callback methods as specified in annotations.
 The `entity-listener` XML descriptor element
 is used to specify the lifecycle listener methods of an entity listener
 class. The lifecycle listener methods are specified by using the
-`pre-persist`, `post-persist`, `pre-remove`, `post-remove`,
-`pre-update`, `post-update`, and/or `post-load` elements.
+`pre-persist`, `post-persist`, `pre-remove`, `post-remove`,`pre-merge`,
+`pre-insert`, `post-insert`, `pre-delete`, `post-delete`, `pre-upsert`,
+`post-upsert`, `pre-update`, `post-update`, and/or `post-load` elements.
 
 ===== Specification of the Binding of Entity Listener Classes to Entities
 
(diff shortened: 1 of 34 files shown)


gavinking:
@njr-11 @otaviojava please review this. Does this look like it satisfied the needs of Jakarta Data?

I really want to get this into M1 so that we have it as a foundation for the work we're doing in Jakarta Data.

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -678,17 +678,32 @@ <T> List<T> findMultiple(EntityGraph<T> graph, List<?> ids,
...
-     * The select list of the query must contain only a single
-     * item, which must be assignable to the type specified by
-     * the {@code resultClass} argument.
+     * Jakarta Persistence query language SELECT statement and
+     * returning instances of the given result class. Either:
+     * <ol>
+     * <li>the select list of the query contains only a single
+     *     item, which must be assignable to the result class,
+     *     or
+     * <li>the result class must be a non-abstract class or
```
otaviojava, line 687: This one goes beyond, once, in Jakarta Data, we are working on the record, if I am not mistaken...

PS: I am not against it on Jakarta Persistence.

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -678,17 +678,32 @@ <T> List<T> findMultiple(EntityGraph<T> graph, List<?> ids,
...
+     *     record type with a constructor with the same number
+     *     of parameters as the query has items in its select
```
otaviojava, line 689: Question:

Given a class and a Projection:

```java
@Entity
class Book {
private Long isbn:

private String title:

private String category:

private String author:

private Year release:
}

record BookRelease(String isbn, String title, Year release ){}
```
Based on:

>  A record type with a constructor with the same number of parameters as the query has items in its select list, and the constructor parameter types must exactly match the types of the corresponding items in the select list

As written, the query:

```java
  @StaticQuery("from Book release = now()") //please ignore the funcion name here
 List<BookRelease> releaseOfTheYear();
```

**It is invalid** because it neither selects three fields nor matches the record constructor. Under the new rules, the provider would reject this at parse/compile time.


Question: 

Is the intention to forbid implicit “partial entity” mappings, or is this an acceptable (provider-defined) convenience when the projection clearly represents a portion of the entity?

gavinking: @otaviojava Yes I wondered about that but then I went and looked how this is defined in the JPQL spec, and it says that a missing `select` clause means that a `select` clause in "inferred". So there actually *is* `select`, it's just implicit.

Now, a wrinkle with that is I believe we characterize this a bit differently in Jakarta Query, and so this language (like many other things) might need adjustment once we remove the definition of JPQL from the spec. But it's OK for now, I think.

> s the intention to forbid implicit “partial entity” mappings, or is this an acceptable (provider-defined) convenience when the projection clearly represents a portion of the entity?

It is intentional. That is explicitly called out as forbidden by the spec, IIRC.

api/src/main/java/jakarta/persistence/EntityHandler.java
gavinking: I am fine either way. I thought it was maybe slightly better to accept classes here because in this context people might be migrating from the use of `select new`, which does accept a class. Of course, it's usually sensible to use a `record`. For an implementation, there's no difference either way.

gavinking:
I have extended this to cover native SQL queries and stored procedure queries. That involved nailing down some stuff which has simply never been well specified before.

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -784,15 +799,21 @@ <T> List<T> findMultiple(EntityGraph<T> graph, List<?> ids,
...
-     * <li>Otherwise, the first column of the result set is interpreted
-     *     via the specified basic type.
+     * <li>the result class is an entity class and is interpreted as a
+     *     managed {@linkplain EntityResult entity result} with implicit
+     *     field mappings determined by the names of the columns in the
+     *     result set and the object/relational mapping of the entity,
+     * <li>the result class is the class of a {@linkplain Basic basic}
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
```
njr-11:
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -868,12 +890,27 @@ <T> TypedQuery<T> createNativeQuery(String sqlString,
...
+     * invocation. For each given result class, either:
+     * <ul>
+     * <li>the result class is an entity class and is interpreted as a
+     *     managed {@linkplain EntityResult entity result} with implicit
+     *     field mappings determined by the names of the columns in the
+     *     result set and the object/relational mapping of the entity,
+     * <li>the result class is the class of a {@linkplain Basic basic}
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
```
njr-11:
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/NamedNativeQuery.java
```diff
@@ -96,26 +101,54 @@
...
+     * {@linkplain #resultSetMapping result set mapping} is not, either:
+     * <ul>
+     * <li>the result class is an entity class and is interpreted as a
+     *     managed {@linkplain EntityResult entity result} with implicit
+     *     field mappings determined by the names of the columns in the
+     *     result set and the object/relational mapping of the entity,
+     * <li>the result class is the class of a {@linkplain Basic basic}
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
```
njr-11:
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/NamedStoredProcedureQuery.java
```diff
@@ -94,13 +97,53 @@
...
+     * {@linkplain #resultSetMappings result set mappings} are not, either:
+     * <ul>
+     * <li>the result class is an entity class and is interpreted as a
+     *     managed {@linkplain EntityResult entity result} with implicit
+     *     field mappings determined by the names of the columns in the
+     *     result set and the object/relational mapping of the entity,
+     * <li>the result class is the class of a {@linkplain Basic basic}
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
```
njr-11:
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -445,19 +442,30 @@ StoredProcedureQuery registerStoredProcedureParameter(
...
+     * <p>Either:
+     * <ul>
+     * <li>the result class is an entity class and is interpreted as a
+     *     managed {@linkplain EntityResult entity result} with implicit
+     *     field mappings determined by the names of the columns in the
+     *     result set and the object/relational mapping of the entity,
+     * <li>the result class is the class of a {@linkplain Basic basic}
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
```
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -469,17 +477,15 @@ StoredProcedureQuery registerStoredProcedureParameter(
...
+     * Retrieve the list of results from the next result set, specifying
+     * a {@linkplain ResultSetMapping result set mapping} which overrides
+     * any mapping or result class already specified. The provider calls
+     * {@link #execute} if necessary. A {@code REF_CURSOR} result set,
+     * if any, is retrieved in the order the {@code REF_CURSOR} parameter
+     * was registered with the query.
      * @param mapping the result set mapping to apply to the results
-     * @return a list of the results or null is the next item is not
-     * a result set
+     * @return a list of the results or null is the next item is not a
```
```suggestion
     * @return a list of the results or null if the next item is not a
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -445,19 +442,30 @@ StoredProcedureQuery registerStoredProcedureParameter(
...
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
+     *     with a constructor with the same number of parameters as the
+     *     result set has columns, and is interpreted as a
+     *     {@linkplain ConstructorResult constructor result} including
+     *     all the columns of the result set.
+     * </ul>
      * @param resultClass the type of the query result
      * @return a list of the results or null is the next item is not
```
```suggestion
     * @return a list of the results or null if the next item is not
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -491,16 +497,27 @@ StoredProcedureQuery registerStoredProcedureParameter(
...
+     * <p>Either:
+     * <ul>
+     * <li>the result class is an entity class and is interpreted as a
+     *     managed {@linkplain EntityResult entity result} with implicit
+     *     field mappings determined by the names of the columns in the
+     *     result set and the object/relational mapping of the entity,
+     * <li>the result class is the class of a {@linkplain Basic basic}
+     *     type and the result set must have a single column which is
+     *     interpreted as a {@linkplain ColumnResult scalar result}, or
+     * <li>the result class must be non-abstract class or record type
```
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -541,16 +556,27 @@ StoredProcedureQuery registerStoredProcedureParameter(
...
```
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -566,14 +592,27 @@ StoredProcedureQuery registerStoredProcedureParameter(
...
```
```suggestion
     * <li>the result class must be a non-abstract class or record type
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2920,6 +2920,31 @@ parameter.
...
+----
+<T> TypedQuery<T> createQuery(String qlString, Class<T> resultClass);
+----
+
+The `TypedQuery` created by this method returns instances of the given
+result class. Either:
+
+- the select list of the query contains only a single item, which must
+  be assignable to the given result class, or
+- the result class must be non-`abstract` class or record type with a
```
```suggestion
- the result class must be a non-`abstract` class or record type with a
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -3418,6 +3443,32 @@ An instance of `ResultSetMapping` representing a result set mapping
 declared using the `@SqlResultSetMapping` annotation may be obtained
 from the `EntityManagerFactory` by calling `getResultSetMappings()`.
 
+==== Result Classes for SQL Queries
+
+The `createQuery()` method of `EntityHandler` accepts, among other
```
njr-11: This is in a section referring to SQL queries and the following example is of createNativeQuery, not createQuery
```suggestion
The `createNativeQuery()` method of `EntityHandler` accepts, among other
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -3418,6 +3443,32 @@ An instance of `ResultSetMapping` representing a result set mapping
...
+result class. Either:
+
+- the result class is an entity class and is interpreted as a managed
+  entity result with implicit field mappings determined by the names
+  of the columns in the result set and the object/relational mapping
+  of the entity, as specified above in <<entity-results>>,
+- the result class is the class of a basic type listed in <<a486>> and
+  the result set must have a single column which is interpreted as a
+  scalar result as specified in <<scalar-results>>, or
+- the result class must be non-`abstract` class or record type with a
```
```suggestion
- the result class must be a non-`abstract` class or record type with a
```

gavinking:
Thanks, @njr-11, I applied them all.

gavinking:
I have moved the PR out of draft mode.

njr-11 (approved):

api/src/main/java/jakarta/persistence/EntityHandler.java
gavinking: @otaviojava found it:

> When an entity is to be returned from a native query, the SQL statement should select all of the columns that
are mapped to the entity object. This should include foreign key columns to related entities. The results obtained when insufficient data is available are undefined.

gavinking: Oh and this too:

> The `SELECT` clause is optional. A query with a missing `SELECT` clause is interpreted as if it had the following single-item `SELECT` clause: `select this`, where `this` is the implicit identification variable.

So I believe we are good here, at least for now.

gavinking:
I think I'm happy with this for now. If we later find that this is insufficient when we circle back to this in the context of the Data and Query spec, then we can make adjustments at that time. But I had a look at what we're saying right now in the Data spec and this LGTM.

Issue #766, opened 2025-11-21
PersistenceConfiguration uses deprecated name "JTA"


arjantijms:
A few years ago we did a large sweep throughout the Jakarta EE APIs to replace deprecated names like "JTA", "JSF", etc for their new names like Jakarta Transactions, Jakarta Faces, etc.

Unfortunately `jakarta.persistence.PersistenceConfiguration` still uses the old and deprecated names:

```java
/**
     * Specify the JNDI name of a JTA {@code javax.sql.DataSource}.
     * @param dataSourceJndiName the JNDI name of a JTA datasource
     * @return this configuration
     */
    public PersistenceConfiguration jtaDataSource(String dataSourceJndiName) {
        this.jtaDataSource = dataSourceJndiName;
        return this;
    }

    /**
     * The JNDI name of a JTA {@code javax.sql.DataSource}.
     * @return the configured JTA datasource, if any, or null
     */
    public String jtaDataSource() {
        return jtaDataSource;
    }

    /**
     * Specify the JNDI name of a non-JTA {@code javax.sql.DataSource}.
     * @param dataSourceJndiName the JNDI name of a non-JTA datasource
     * @return this configuration
     */
    public PersistenceConfiguration nonJtaDataSource(String dataSourceJndiName) {
        this.nonJtaDataSource = dataSourceJndiName;
        return this;
    }

    /**
     * The JNDI name of a non-JTA {@code javax.sql.DataSource}.
     * @return the configured non-JTA datasource, if any, or null
     */
    public String nonJtaDataSource() {
        return nonJtaDataSource;
    }
```

"JTA" should ideally be changed here to "Jakarta Transactions".

arjantijms:
The term appears here too:

```java
/**
 * Enumerates the possible approaches to transaction management in Jakarta
 * Persistence. An {@link EntityManager} may be a JTA entity manager, where
 * transaction management is done via JTA, or it may be a resource-local
 * entity manager, where transaction management is performed via the
 * {@link EntityTransaction} interface.
 *
 * @since 3.2
 */
public enum PersistenceUnitTransactionType {
    /**
     * Transaction management via JTA.
     */
    JTA,
    /**
     * Resource-local transaction management.
     */
    RESOURCE_LOCAL
}
```

Interestingly, "JPA" was changed to "Jakarta Persistence" here, but somehow "JTA" wasn't changed.

gavinking:
[...]

PR #836, opened 2025-12-16
clarify that "JTA" means Jakarta Transactions

gavinking:
and fix up some sentences that should include `EntityAgent`

#766


gavinking:
@njr-11 WDYT?

njr-11 (approved):
That's amazing JTA was used in so many places without explicitly spelling it out as Jakarta Transactions.

gavinking:
> That's amazing JTA was used in so many places without explicitly spelling it out as Jakarta Transactions.

No link to the spec either. Better add that.

gavinking:
I added it to the references.

Issue #814, opened 2025-12-13
Bean Validation and new lifecycle callbacks


gavinking:
We need to think about the impact of the new lifecycle callback types on the integration with Bean Validation. At the very least we should define new properties allowing you to turn validation on for the new `pre-*` events.

@marko-bekhta you want to put together a proposal?

PR #838, opened 2025-12-16
Account for new lifecycle event types during bean validation

marko-bekhta:
- fixes https://github.com/jakartaee/persistence/issues/814

Hey 👋🏻 

I went through the section about Jakarta Validation integration and here's the first draft of where we should probably take this.

There is a breaking change in a sense that validation will be ignored by default for all "manage-related events" and only triggered for the DB-related events, i.e. we'd stop triggering it by default for pre-persist, but instead trigger it for pre-insert/pre-update/pre-upsert.
I went with this approach since we will always get those pre-db versions if we were to write the changes back to a DB and that's where validation is important. We won't be getting pre-persist if we are within the "stateless case", but if we are working with EM we'd be getting two validation rounds for persist for no good reason...

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -1856,14 +1856,16 @@ and specifies additional validation points for managed types
 (entities, mapped superclasses, and embeddable classes) at the
 entity lifecycle events described in <<a2202>>.
 
-The Jakarta Persistence provider may automatically enforce the
-Jakarta Validation constraints <<a19498>> defined on managed objects
-(<<a2374>>). This process occurs internally within the persistence
-lifecycle, ensuring that validation is performed without requiring
-additional application-side logic beyond the definition of the
-constraints themselves.
+When the Jakarta Persistence provider is
+<<a2394,aware of the Jakarta Validation provider's presence>>
+and <<a2374,enables automatic validation>>, it must automatically
+enforce the Jakarta Validation constraints <<a19498>> defined on
+managed objects. This process occurs internally within the persistence
+lifecycle (<<automatic-validation-lifecycle-events>>), ensuring
+validation without requiring additional application-side logic
+beyond the constraints' definitions.
 
-==== Automatic Validation Upon Lifecycle Events
+==== Automatic Validation Upon Lifecycle Events [[automatic-validation-lifecycle-events]]
 
 _Automatic validation_ of entities is specified in terms of their
 lifecycle events defined in <<a2202>>.
@@ -1963,7 +1965,7 @@ and must mark the transaction for rollback if the persistence context is joined
 to the transaction.
 
 The validation process must not have any side effects;
-i.e., validation must not result in the loading of any
+validation must not result in the loading of any
 attribute or association which is not already loaded.
 An entity can be validated at most once per flush cycle.
 
@@ -1987,14 +1989,13 @@ where `tr` is the resolver.
 In Jakarta EE environments, a `jakarta.validation.ValidatorFactory`
 instance is made available by the Jakarta EE container. The container
 is responsible for passing this validator factory to the persistence
-provider via the map that is passed as an argument to the
-`createContainerEntityManagerFactory` call. The map key used by the
-container must be the standard property name
-`jakarta.persistence.validation.factory`.
+provider via a map passed as the second argument to
+`createContainerEntityManagerFactory`. The map key must be the
+standard property name `jakarta.persistence.validation.factory`.
 
-In Java SE environments, the application can pass the `ValidatorFactory`
-instance via the map that is passed as an argument to the
-`Persistence.createEntityManagerFactory` call. The map key used must
+In Java SE environments, the application can supply the
+`ValidatorFactory` instance via a map passed as the second argument to
+`Persistence.createEntityManagerFactory`. The map key must
 be the standard property name `jakarta.persistence.validation.factory`.
 If no `ValidatorFactory` instance is provided by the application, and
 if a Jakarta Validation provider is present in the classpath, the


gavinking:
Yeah, makes sense. I was thinking along the same lines, except I was wondering about whether some users might like to get errors *earlier*. But I guess they still can, if they configure that explicitly, right?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1930,30 +1930,36 @@ validation.
...
 ===== Requirements for Automatic Validation upon Lifecycle Events [[a2380]]
 
-For each event type, a list of groups is
-targeted for validation. By default, the default Bean Validation group
-(the group `Default`) will be validated upon the pre-persist and
-pre-update lifecycle validation events, and no group will be validated
-upon the pre-remove event.
+Each lifecycle event type has a list of validation groups assigned to it.
+If the list of validation groups for a particular event is not empty,
+validation is applied; otherwise, no validation is applied during that particular event.
```
```suggestion
validation is applied; otherwise, no validation is applied during that 
particular event.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1930,30 +1930,36 @@ validation.
...
+
+By default, only the lists of validation groups
+for the pre-insert, pre-update and pre-upsert lifecycle events
+contain the `jakarta.validation.groups.Default` group (the default Jakarta Validation group).
```
```suggestion
By default, only the lists of validation groups for the pre-insert, pre-update 
and pre-upsert lifecycle events contain the `jakarta.validation.groups.Default` 
group (the default Jakarta Validation group).
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1930,30 +1930,36 @@ validation.
...
+The value of a configuration property must be a list of the targeted groups.
+A targeted group must be specified by its fully qualified class name.
+Names must be separated by a comma.
 
-When one of the above events occurs for an
-entity, the persistence provider must validate that entity by obtaining
-a `Validator` instance from the validator factory in use (see <<a2394>>) and
+When one of the lifecycle events occurs for an
+entity, the persistence provider must validate that entity by using
+a `Validator` instance obtained from the validator factory in use (see <<a2394>>) and
```
```suggestion
When a lifecycle events occurs for an entity, the persistence 
provider must validate the entity using a `Validator` instance 
obtained from the validator factory in use (see <<a2394>>) and
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1930,30 +1930,36 @@ validation.
 
 ===== Requirements for Automatic Validation upon Lifecycle Events [[a2380]]
 
-For each event type, a list of groups is
-targeted for validation. By default, the default Bean Validation group
-(the group `Default`) will be validated upon the pre-persist and
-pre-update lifecycle validation events, and no group will be validated
-upon the pre-remove event.
+Each lifecycle event type has a list of validation groups assigned to it.
+If the list of validation groups for a particular event is not empty,
```
```suggestion
If the list of validation groups for a particular event is nonempty,
```

gavinking:
@marko-bekhta since we're working on this section, please feel free to clean up the language in the whole section. The spec has quite a lot of clumsy writing that we've been trying to improve bit by bit when we have to touch something.

marko-bekhta:
I've updated the remaining part of the section and was looking into adding some tests to the TCK, but I found none for BV integration in there, soo.. how do you want to deal with it? Should we create a set of tests in the TCK for BV integration or not ?

gavinking:
> Should we create a set of tests in the TCK for BV integration or not ?

Is it a required feature of the spec?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1925,30 +1925,37 @@ validation.
...
+* `jakarta.persistence.validation.group.pre-update`
+* `jakarta.persistence.validation.group.pre-upsert`
+* `jakarta.persistence.validation.group.pre-delete`
 
-The value of a validation property must be a
-list of the targeted groups. A targeted group must be specified by its
-fully qualified class name. Names must be separated by a comma.
+The value of a configuration property must be a list of the targeted groups.
+A targeted group must be specified by its fully qualified class name.
+Names must be separated by a comma.
```
```suggestion
Each pair of names must be separated by a comma.
```

Is whitespace tolerated?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1958,27 +1965,24 @@ to the returned set of `ConstraintViolation` objects, and must mark the
...
-The validator instance that is used for
-automatic validation upon lifecycle events must use a
-`TraversableResolver` that has the following behavior:
-
-* Attributes that have not been loaded must not
-be loaded.
-* Validation cascade (`@Valid`) must not
-occur for entity associations (single-valued or multi-valued).
+The validation process must not have any side effects;
+i.e., it must not load any additional attributes or associations.
```
```suggestion
i.e., validation must not result in the loading of any 
attribute or association which is not already loaded.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1925,30 +1925,37 @@ validation.
...
+A targeted group must be specified by its fully qualified class name.
+Names must be separated by a comma.
 
-When one of the above events occurs for an
-entity, the persistence provider must validate that entity by obtaining
-a `Validator` instance from the validator factory in use (see <<a2394>>) and
+When an entity lifecycle event occurs, the persistence
+provider must validate the entity using a `Validator` instance
+obtained from the validator factory in use (see <<a2394>>) and
 invoking its `validate` method with the targeted groups. If the list of
```
```suggestion
invoke its `validate` method with the targeted groups. If the list of
```

marko-bekhta:
> > Should we create a set of tests in the TCK for BV integration or not ?
> 
> Is it a required feature of the spec?

I'm reading this section:

https://github.com/jakartaee/persistence/blob/394a047b0c4ff6a9fcf81c6215d02a5ded41e037/spec/src/main/asciidoc/ch03-entity-operations.adoc?plain=1#L1847-L1865

and ti doesn't say that it's optional ... so I can only assume that yeah, it's required...

(and while reading it I noticed that I should update the `Automatic Validation Upon Lifecycle Events` section too...)

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: I'd leave it to the implementation to define how flexible it wants to go in terms of parsing the sting... but I guess a `split` and `trim` aren't that complex, so let's say:

```suggestion
Each pair of names must be separated by a comma.
Optional whitespace characters shall be permitted both preceding and following each group name.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-provider present in the environment, and the provider must throw the
-`PersistenceException` if the `jakarta.persistence.validation.mode`
-property value _"callback"_ has been passed to the
-`Persistence.createEntityManagerFactory` method.
-
-If the none validation mode is specified by
-the `validation-mode` element or the `jakarta.persistence.validation.mode`
-property, the persistence provider must not perform lifecycle event
-validation.
+This specification defines the following validation modes:
```
gavinking: You're defining the term "validation mode", so it goes in italics.

```suggestion
This specification defines the following _validation modes_:
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+
+* `auto` -- if the Jakarta Validation _provider_ is obtainable,
+then the persistence provider must perform the automatic validation of entities
+as described in <<a2380>>.
+Otherwise, no lifecycle event validation takes place.
+This is the default validation mode.
```
gavinking: You're not defining the term "provider", so no italics.

```suggestion
* `auto` -- if the Jakarta Validation provider is obtainable, then the 
  persistence provider must perform automatic validation of entities
  as described in <<a2380>>. Otherwise, no lifecycle event validation 
  takes place. This is the default validation mode.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+* `callback` -- the persistence provider must perform the automatic validation of entities
+as described in <<a2380>>. If no Jakarta Validation _provider_ is available,
+the persistence provider must throw the `PersistenceException`.
```
```suggestion
* `callback` -- the persistence provider must perform the automatic validation 
  as described in <<a2380>>. If no Jakarta Validation provider is available,
  the persistence provider must throw a `PersistenceException`.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+* `none` -- the persistence provider must not perform any lifecycle event validation.
```
gavinking: You changed the language to "automatic" validation in the other bullet point, so be consistent here, otherwise people might wonder what is the difference between "automatic validation" and "lifecycle event validation". Linguistic variety is great in most kinds of writing, but not in specifications.

```suggestion
* `none` -- the persistence provider must not perform any automatic validation.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+
+[NOTE]
+====
+The Jakarta Validation _provider_ is obtained via the `jakarta.validation.ValidatorFactory` (<<a2394>>).
```
```suggestion
The Jakarta Validation provider is obtained via the `jakarta.validation.ValidatorFactory`,
as specified in<<a2394>>.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+====
+
+The validation mode may be specified via the `validation-mode` element
+of the `persistence.xml` file.
+Allowed values for the `validation-mode` element are `AUTO`, `CALLBACK`, `NONE`.
+
+The validation mode value derived from the `persistence.xml`
+may be overridden by the `jakarta.persistence.validation.mode` configuration property
+(e.g. passed to the  `Persistence.createEntityManagerFactory` factory method).
+Expected values for this property are _"auto"_, _"callback"_, _"none"_.
```
gavinking: Trying to get away from weird nonstandard use of quotes.

```suggestion
Expected values for this property are `auto`, `callback`, `none`.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+The Jakarta Validation _provider_ is obtained via the `jakarta.validation.ValidatorFactory` (<<a2394>>).
+====
+
+The validation mode may be specified via the `validation-mode` element
+of the `persistence.xml` file.
+Allowed values for the `validation-mode` element are `AUTO`, `CALLBACK`, `NONE`.
+
+The validation mode value derived from the `persistence.xml`
+may be overridden by the `jakarta.persistence.validation.mode` configuration property
+(e.g. passed to the  `Persistence.createEntityManagerFactory` factory method).
```
gavinking: It's always better to given an explicit list of possibilities than leave an "e.g." hanging in the air. I believe that in this case, there are only two possibilities.

```suggestion
The validation mode value specified in `persistence.xml` may be overridden by 
the configuration property `jakarta.persistence.validation.mode` which may be
passed to `Persistence.createEntityManagerFactory` or 
`PersistenceProvider.createContainerEntityManagerFactory`, as defined in
<<a13443>> and <<a12802>>.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1958,31 +1950,29 @@ to the returned set of `ConstraintViolation` objects, and must mark the
...
+the validator instance in use with the traversable resolver
+(`jakarta.validation.TraversableResolver`)
+that matches these behavior requirements, in particular:
 
-Embeddable attributes must be validated only
-if the `Valid` annotation has been specified on them.
+* `TraversableResolver#isReachable(..)` only returns `true`
+for already initialized attributes and associations.
+* `TraversableResolver#isCascadable(..)` only returns `true`
+for properties that are not associations.
```
gavinking: The term "loaded" is used in the passage above this list.
Also, please prefer the syntax of Java to the syntax of Javadoc.

```suggestion
* `TraversableResolver.isReachable(..)` only returns `true`
  for attributes and associations which are already loaded.
* `TraversableResolver.isCascadable(..)` only returns `true`
  for attributes that are not associations.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1958,31 +1950,29 @@ to the returned set of `ConstraintViolation` objects, and must mark the
...
+attribute or association which is not already loaded.
+An entity can be validated at most once per flush cycle.
 
-These requirements guarantee that no unloaded
-attribute or association will be loaded by side effect and that no
-entity will be validated more than once during a given flush cycle.
+To achieve this, persistence provider must initialize
+the validator instance in use with the traversable resolver
+(`jakarta.validation.TraversableResolver`)
+that matches these behavior requirements, in particular:
```
```suggestion
that matches these behavior requirements. In particular:
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
 
-=== Bean Validation [[a2366]]
+=== Jakarta Validation [[a2366]]
 
-This specification defines support for use of
-Bean Validation <<a19498>> within Jakarta Persistence
-applications.
+Jakarta Persistence integrates with Jakarta Validation <<a19498>>
+and specifies additional validation points for managed types
+(i.e. entities, mapped superclasses, and embeddable classes)
```
gavinking: "i.e." does no work here

```suggestion
(entities, mapped superclasses, and embeddable classes)
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-is achieved by specifying that Jakarta Persistence delegate validation to
-the Bean Validation implementation upon the pre-persist, pre-update, and
-pre-remove entity lifecycle events described in <<a2202>>.
-
-Validation can also be achieved by the
-application calling the `validate` method of a `Validator` instance upon
-an instance of a managed class, as described in the Bean Validation
-specification <<a19498>>.
+Jakarta Persistence provider may automatically trigger validation of the managed objects
+as defined by the Jakarta Validation specification <<a19498>>.
```
gavinking: Not really very sure what this sentence is trying to say. Also, it's missing an article.
Be careful with "may", it's often quite ambiguous, as it is here.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-which all the `PrePersist`, `PreUpdate`, and `PreRemove` lifecycle
-callback method invocations respectively have been completed, or
-immediately after the point at which such lifecycle callback methods
-would have been completed (in the event that such callback methods are
-not present).
+If enabled, the automatic validation of entities
+upon lifecycle validation events must occur immediately after the point at
+which all the corresponding lifecycle callback method invocations have been completed,
+or immediately after the point at which such lifecycle callback methods
+would have been completed (if such callback methods are absent).
```
gavinking: "after the point at which some things have been completed" is a clumsy use of the perfect tense. Perfect tense already comes with after-ness built in.

 I think it's better to just say:

-  "after the point at which some things complete", or
-  "when something things have been completed".

```suggestion
_Automatic validation_ of entities is specified in terms of their lifecycle events
defined in <<a2202>>.

If enabled, automatic validation of entities must occur immediately after the point at
which all corresponding lifecycle callback method invocations complete, or immediately 
after the point at which such lifecycle callback methods would have completed if there
were any such callback methods present.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-In the case where an entity is persisted and
-subsequently modified in a single transaction or when an entity is
-modified and subsequently removed in a single transaction, it is
-implementation dependent as to whether the pre-update validation event
-occurs. Portable applications should not rely on this behavior.
+In the same way as with entity lifecycle events (<<a2202>>),
+the number of times automatic validation is triggered
+for an entity within a single transaction,
+when multiple operations are performed on the same entity,
+depends on the Jakarta Persistence provider.
```
gavinking: I had to read this sentence three times to understand what it was saying. Think of a different way to structure it. (Also, "In the same way as with" is inelegant.)

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1844,112 +1844,104 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+
+For example, whether a pre-update validation event occurs
+after an entity is persisted and then modified,
+or after it is modified and then removed within the same transaction,
+is implementation-dependent.
+
+Portable applications must not rely on the occurrence or order
+of such automatic validation events.
```
```suggestion
Portable applications must not rely on the specific timing or ordering
of automatic validation events.
```

gavinking (changes requested):
Thanks, @marko-bekhta. I've left a bunch of nitpicks because I'm just generally a pedantic asshole.

But also because it's our job to be pedants here :)

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: Applied your suggestion 🙂 
And just so you know why I made the change here in the first place ... when I was going through this validation section, it felt like we were constantly repeating the same thing: we can put a property in XML or pass it through the map. So at the time, I wished we had a "configuring the persistence provider" section somewhere that explained the available ways to configure things and their priorities for overriding one another 🙂.

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: Yeah... 🙈 the idea was to say that the validation can be enabled/disabled based on the validation mode and that things happen internally so that users only need to add constraints to their entities, but I forgot to add the second part, and the first one is confusing in a way it's written 🤦🏻 🙂 
Hopefully the updated version is better 🙂 🤞🏻

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: Sure, refactoring out copypasta into a proper definition is very often better than laboriously reciting the same phrase over and over. You're allowed to add new sections.

marko-bekhta:
I've added basic tests around the validation logic to the TCK. I think we shouldn't bring any implementation for the validation and just use the API, we don't really care if the validation is performed correctly or not, all we need is to make sure that validation was requested (i.e. `validator.validate(entity, groupsWeNeed)` was called). So having a throwing `Validator` impl with `validator.validate(...)` recording the calls to it should be enough.
I also added a note to the change log.

gavinking:
>  we don't really care if the validation is performed correctly or not, all we need is to make sure that validation was requested 

Agreed.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1849,162 +1849,158 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-Automatic validation using these constraints
-is achieved by specifying that Jakarta Persistence delegate validation to
-the Bean Validation implementation upon the pre-persist, pre-update, and
-pre-remove entity lifecycle events described in <<a2202>>.
-
-Validation can also be achieved by the
-application calling the `validate` method of a `Validator` instance upon
-an instance of a managed class, as described in the Bean Validation
-specification <<a19498>>.
+The Jakarta Persistence provider may automatically enforce the
```
gavinking: What does "may" here mean?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1849,162 +1849,158 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-key used must be the standard property name
-`jakarta.persistence.validation.factory`. If no `ValidatorFactory`
-instance is provided by the application, and if a Bean Validation
-provider is present in the classpath, the persistence provider must
-instantiate the `ValidatorFactory` using the default bootstrapping
-approach defined by the Bean Validation specification
-<<a19498>>, namely `Validation.buildDefaultValidatorFactory()`.
+In Java SE environments, the application can pass the `ValidatorFactory`
+instance via the map that is passed as an argument to the
+`Persistence.createEntityManagerFactory` call. The map key used must
```
```suggestion
In Java SE environments, the application can supply the `ValidatorFactory`
instance via a map passed as the second argument to
`Persistence.createEntityManagerFactory`. The map key must
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1849,162 +1849,158 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+in use (see <<a2394>>) and invoke its `validate` method with the targeted groups.
+If the list of targeted groups is empty, no validation is performed. If the
+set of `ConstraintViolation` objects returned by the `validate` method is not
+empty, the persistence provider must throw the `jakarta.validation.ConstraintViolationException`
+containing a reference to the returned set of `ConstraintViolation` objects,
+and must mark the transaction for rollback if the persistence context is joined
+to the transaction.
+
+The validation process must not have any side effects;
+i.e., validation must not result in the loading of any
```
```suggestion
validation must not result in the loading of any
```

i.e. is doing no work

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1849,162 +1849,158 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
-In Jakarta EE environments, a `ValidatorFactory`
-instance is made available by the Jakarta EE container. The container is
-responsible for passing this validator factory to the persistence
+In Jakarta EE environments, a `jakarta.validation.ValidatorFactory`
+instance is made available by the Jakarta EE container. The container
+is responsible for passing this validator factory to the persistence
 provider via the map that is passed as an argument to the
 `createContainerEntityManagerFactory` call. The map key used by the
 container must be the standard property name
 `jakarta.persistence.validation.factory`.
```
```suggestion
provider via a map passed as the second argument to
`createContainerEntityManagerFactory`. The map key
must be the standard property name
`jakarta.persistence.validation.factory`.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: let me know if that updated intro works better 🤞🏻 🙂

gavinking (approved):

gavinking:
Thank you Marko!!

PR #839, opened 2025-12-16
clarify use of Object[].class as a result class

### LINKED ISSUE #835: clarify use of Object[].class as a result class
I assume that it has always been the case in all implementations that you were allowed to pass `Object[].class` as an explicit result class to `createQuery()` but I don't actually see anywhere we say that explicitly. It should at minimum be mentioned in the Javadoc. Not sure if we should also say that `Object.class` is allowed. That's much less useful.

gavinking:
Fixes #835

The semantics of query results was extremely under-specified, especially in the Javadoc, leaving gaping holes in terms of portability between implementations.

The pull request fixes that. It probably introduces a new feature (a bit hard to say) which is the ability to pass `Object[].class` as an argument to `createNativeQuery()`. But that's useful and at least some implementations (Hibernate, ahem) have accepted it since before JPA 1.0 existed.

@sebersole @njr-11 @anija-anil Please review and proofread 🙏

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityHandler.java
+++ b/api/src/main/java/jakarta/persistence/EntityHandler.java
@@ -732,7 +732,7 @@ public sealed interface EntityHandler extends AutoCloseable
 
     /**
      * Create an instance of {@link TypedQuery} for executing
-     * a Jakarta Persistence query language {@code SELECT }
+     * a Jakarta Persistence query language {@code SELECT}
      * statement, specifying an {@link EntityGraph} which is
      * interpreted as a load graph. The select list of the
      * query must contain only a single item, which must be


api/src/main/java/jakarta/persistence/NamedQuery.java
```diff
@@ -90,19 +90,22 @@
...
      * <ol>
      * <li>the select list of the query contains only a single
-     *     item, which must be assignable to the given result
-     *     class, or
-     * <li>the result class must be a non-abstract class or
-     *     record type with a constructor with the same number
-     *     of parameters as the query has items in its select
-     *     list, and the constructor parameter types must exactly
+     *     item, which must be assignable to the result class,
+     *     <li>the result class is {@code Object[].class}, or
```
njr-11:
```suggestion
     * <li>the result class is {@code Object[].class}, or
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -511,12 +517,18 @@ StoredProcedureQuery registerStoredProcedureParameter(
      *     result set and the object/relational mapping of the entity,
      * <li>the result class is the class of a {@linkplain Basic basic}
      *     type and the result set must have a single column which is
-     *     interpreted as a {@linkplain ColumnResult scalar result}, or
-     * <li>the result class must be a non-abstract class or record type
+     *     interpreted as a {@linkplain ColumnResult scalar result},
+     * <li>the result class os a non-abstract class or record type
```
```suggestion
     * <li>the result class is a non-abstract class or record type
```

njr-11 (approved):
I think I have seen EclipseLink with the same behavior of packaging multiple SELECT items into an Object[].

gavinking:
Thanks so much for the proofread Nathan. Applied your fixes.

> I think I have seen EclipseLink with the same behavior of packaging multiple SELECT items into an Object[].

Yeah yeah it's totally standard when you _don't_ specify a result class. It just wasn't clear whether you're allowed to specify `Object[].class` *explicitly*. But I would be very surprised if EL doesn't let you do it instead of doing the typecast afterward. intuitively, if the query was already going to return `Object[]`s all by itself without an explicit result class, it should still return`Object[]`s if you explicitly specify `Object[].class`.

api/src/main/java/jakarta/persistence/NamedNativeQuery.java
```diff
@@ -116,23 +116,38 @@
...
+     * <ul>
+     * <li>it is inferred from the {@linkplain #resultSetMapping result
+     *     set mapping}, if any, or
+     * <li>it defaults to {@code Object} if the result set has a single
+     *     column or {@code Object[]} if the result set has multiple
+     *     columns.
+     * </ul>
+     * <p>When neither result class nor result set mapping is specified,
+     * column values are obtained according to the default type mappings
+     * defined by the JDBC specification.
```
gavinking, line 147: I have added this additional clarification.

api/src/main/java/jakarta/persistence/NamedQuery.java
```diff
@@ -80,29 +80,40 @@
...
+     * automatically packaged in a new instance of the result
+     * class by calling the matching constructor.
+     *
+     * <p>If the result class of a named query is not specified:
+     * it defaults to {@code Object} if the query has a single
+     * item in its select list or to {@code Object[]} if the
+     * query has multiple items in its select list. When
+     * generating the canonical static metamodel for a named
+     * query, the generator is permitted to infer a more precise
+     * type from the items in the select list.
```
gavinking, line 111: And also this clarification.

PR #840, opened 2025-12-16
make StoredProcedureQuery extend AutoCloseable

### LINKED ISSUE #162: StoredProcedureQuery and statements leaks

gavinking:
and add a note about something undefined (AFAICT) in the spec

fixes #162

### DIFF
--- a/api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
+++ b/api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
@@ -103,7 +103,7 @@ import java.util.List;
  *
  * @since 2.1
  */
-public interface StoredProcedureQuery extends Query {
+public interface StoredProcedureQuery extends Query, AutoCloseable {
 
     /**
      * Set a query property or hint. The hints elements may be used 
@@ -656,4 +656,17 @@ public interface StoredProcedureQuery extends Query {
      */
     int getUpdateCount();
 
+    /**
+     * Immediately destroy all resources associated with this query.
+     * If the client does not call this method before the end of the
+     * current transaction, the behavior is undefined. A provider
+     * might release resources when all query results are exhausted,
+     * at the end of the current transaction, or when the entity
+     * manager or entity agent is closed.
+     * <p>After invocation of {@code close()}, every method of the
+     * {@code StoredProcedureQuery} throws {@code IllegalStateException}.
+     * @since 4.0
+     */
+    @Override
+    void close();
 }


gavinking:
@sebersole Does this look right to you?

gavinking:
I reviewed how `CallableStatement` works today, and this still seems to be a very important missing feature.

sebersole:
Given the way StoredProcedureQuery is defined I think this is fine.  A method to close/release it was definitely an important missing piece.

gavinking:
Thanks Steve. I think this is good to go.

PR #842, opened 2025-12-17
introduce stronger distinction between Query and TypedQuery

### LINKED ISSUE #718: distinction between typed (selection) queries and update queries
We currently don't have a very clean or well-defined split between `TypedQuery`, which I claim should always represent a `select` query, and `update`/`delete` queries, which I claim should always be executed via the untyped supertype `Query`.

This sorta reflects the fact that `TypedQuery` was hacked on later in JPA 2.0.

In fact, it's quite arguable that Persistence 3.2 lets you execute a named update query using `TypedQuery<Void>` obtained via a `TypedQueryReference`. I don't recall if we really intended that, but even if we did, I don't think it's great.

As a first step to cleaning this up I would:

1. override and deprecate `executeUpdate()` on `TypedQuery`
2. introduce the (unparameterized) type `QueryReference` to represent untyped queries

A further optional step might be to introduce a dedicated interface for executing `update`/`delete` queries (the equivalent of Hibernate's `MutationQuery`) which subtracts all the operations of `Query` which are only relevant to `select` queries. I'm not sure of the best way to do this, and I'm concerned about introducing an explosion of `createQuery()` operations, so perhaps it's not worth going this far.

gavinking:
Redo of #719

fixes #718

Please see the description of what this does in comments on #718

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -2828,13 +2828,15 @@ returned by a stored procedure query invocation is determined according
 to the rules for `Query` or `TypedQuery` as described above.
 
 If the persistence context is joined to the transaction, any runtime
-exception thrown by any method not explicitly listed below of `Query`,
-`Statement`, `TypedQuery`, or `StoredProcedureQuery` causes the current
-transaction to be marked for rollback, unless the exception is an
-instance of one of the following exception types:
+exception thrown by any method of `Query`, `Statement`, `TypedQuery`,
+or `StoredProcedureQuery` causes the current transaction to be marked
+for rollback, unless the exception is an instance of one of the following
+exception types:
 
 - `NonUniqueResultException` or `NoResultException`, or
-- `QueryTimeoutException` or `LockTimeoutException`.
+- `QueryTimeoutException` or `LockTimeoutException`,
+
+or unless the method is explicitly excepted by the rules given below.
 
 On any database platform on which a query timeout causes an immediate
 transaction rollback, the persistence provider must throw a


gavinking:
> Unless the Data provider parses the JPQL to the degree that it understands which entity attribute is selected and what type it is such that it could send a valid class to `TypedQuery`, it will need to use `Query` for this.

We have now clearly specified that this works:

```java
List<Object> entities = em.createQuery(someJpqlString, Object.class).getResultList();
```

and it probably always worked. (Certainly did always work in Hibernate.)

And notice that you get back a `List<Object>` from this query instead of a raw `List`. There's no way we can ever migrate away from that 'orribe signature `List getResultList()` without deprecating it fist, since just changing it to `List<Object>` is a breaky change.

gavinking:
@anija-anil your thoughts on this one?

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -185,7 +213,10 @@ default Stream getResultStream() {
      *         conflict is detected during the flush
      *
      * @since 3.2
+     * @deprecated Use {@link TypedQuery#getSingleResult}
```
njr-11, line 266:
```suggestion
     * @deprecated Use {@link TypedQuery#getSingleResultOrNull}
```

api/src/main/java/jakarta/persistence/QueryReference.java
```diff
@@ -0,0 +1,97 @@
...
+     * Any mutation of the returned list results in an
+     * {@link UnsupportedOperationException}.
+     *
+     * @since 4.0
+     */
+    List<String> getParameterNames();
+
+    /**
+     * The arguments supplied to the query parameters,
+     * or {@code null} if no arguments were supplied.
```
njr-11: Would it be better to use the empty list for no arguments rather than a `null` list?
```suggestion
     * or the {@linkplain java.util.Collections#emptyList empty list}
     * if no arguments were supplied.
```

api/src/main/java/jakarta/persistence/query/StaticTypedQueryReference.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(resultType, name, parameterTypes, parameterNames, arguments, options);
+    }
+
+    @Override
+    public String toString() {
+        return "StaticQueryReference["
```
njr-11, line 254:
```suggestion
        return "StaticTypedQueryReference["
```

api/src/main/java/jakarta/persistence/QueryReference.java
gavinking: I really mean `null` here, because I'm allowing for the case of `@NamedQuery` occurring on a class.

I'm not talking about the case of a `@StaticQuery` on a method with no parameters.

api/src/main/java/jakarta/persistence/QueryReference.java
```diff
@@ -0,0 +1,97 @@
...
+//     Gavin King      - 4.0
+
+package jakarta.persistence;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A reference to a named query declared via the {@link NamedQuery}
+ * or {@link NamedNativeQuery} annotations.
```
gavinking: Need to mention `@StaticQuery` here, I believe.

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -36,6 +55,32 @@
...
+     * which must be a supertype of the result type of this query.
+     * This query must be a Jakarta Persistence {@code SELECT} query
+     * or a native SQL query which returns a result set.
+     * @param resultType The Java class of the query result type
+     * @param <R> The query result type
+     * @throws IllegalArgumentException if the given result type is
+     *         not a supertype of the result type of this query
+     * @since 4.0
+     */
+    <R> TypedQuery<R> ofType(Class<R> resultType);
```
gavinking, line 116: Who prefers `withResultType()` here?

njr-11: I would say either is fine.  `withResultType` is more precise.  `ofType` works because the method transforms a `Query` into a `TypedQuery`.

api/src/main/java/jakarta/persistence/QueryReference.java
gavinking: Done.

gavinking:
I have decided that I like this after all, and I would like to move this forward for JPA 4 M1, if other people also agree that this is a good direction to take. The motivation behind prioritizing it for inclusion in M1 is that this is *definitely* one we need to seek broad feedback on, especially from implementors.

@sebersole @njr-11 @beikov Please take a careful look over what is proposed here. This is a different approach to the one we took in Hibernate, but the reason for that is that the starting point is different: here we have a generic interface `TypedQuery<R>` extending a non-generic interface `Query` with raw types. Therefore, the solution of mixing in `Statement` and `SelectionQuery` to get `Query` at the bottom of a diamond-shaped hierarchy doesn't apply here, and we have to do something different. Also, I've attempted to do this while (a) not breakage old code, and (b) aggressively minimizing the number of new flavors of query interface.

gavinking:
Actually, this approach now has competition from the much less extreme approach taken by #858. We need to make a choice between the two of them.

beikov:
I'd personally prefer to mark `Query` as deprecated and not try to add anything further to it, but direct people to the `Statement` or `MutationQuery` (if naming is still open for debate) instead.
Since `TypedQuery` overrides everything from `Query` already (I think), just dropping `Query` is an option in a JPA 5.0 without causing too much trouble when people already migrated to the new interface.

gavinking:
We are definitely not going to call anything `MutationQuery`, because that collides with a name already used in Hibernate.

gavinking:
Also, to be clear I'm not even *contemplating* deprecating `Query` and I'm certainly not contemplating removing it in JPA 5 or any other JPA.

*Every single program using JPA* has a dependency to `Query`, not to mention that our own APIs descend from it. Removing it would cause chaos.

So I'm looking for something much more subtle: deprecating *certain operations* of `Query`, in particular anything involving `getResultList()` which returns a raw type. The justification is that that operation has already been producing a warning in the client for almost as long as it has existed, and so users have pretty fair warning to stay away from it.

This PR goes a bit further than that in also deprecating `executeUpdate()`, and that alone might simply be too much, which is why I'm leaning toward doing something less ambitious/disruptive in  https://github.com/jakartaee/persistence/pull/858.

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -749,6 +772,23 @@ <T> List<T> findMultiple(EntityGraph<T> graph, List<?> ids,
      */
     <T> TypedQuery<T> createQuery(String qlString, EntityGraph<T> resultGraph);
 
+    /**
+     * Create an instance of {@link Statement} for executing a
+     * named {@code UPDATE} or {@code DELETE} of statement written
```
njr-11:
```suggestion
     * named {@code UPDATE} or {@code DELETE} statement written
```

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -27,8 +27,55 @@
 import java.util.stream.Stream;
 
 /**
- * Interface used to control query execution.
+ * Declares common operations of interfaces used to control the
+ * execution of statements and queries written in the Jakarta
+ * Persistence query language or in native SQL.
```
njr-11: This seemed long for the summary sentence that appears in the Javadoc listing what each class does. Tried to shorten it a little,
```suggestion
 * Declares common operations for controlling the
 * execution of statements and queries written in the Jakarta
 * Persistence query language or in native SQL.
```

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -37,6 +84,52 @@
...
 
+    /**
+     * Obtain a {@link Statement} representing this query, which
+     * must be some kind of executable statement, that is, a
+     * Jakarta Persistence {@code UPDATE} or {@code DELETE}
+     * statement, or any native SQL statement that returns a row
+     * count. The executable statement may be executed by calling
+     * {@link Statement#execute}.
+     * @throws UnsupportedOperationException if this query is a
+     *         Jakarta Persistence {@code SELECT} query
```
njr-11: Would IllegalStateException be more appropriate here?

```suggestion
     * @throws IllegalStateException if this query is a
     *         Jakarta Persistence {@code SELECT} query
```

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -37,6 +84,52 @@
...
+     * Obtain a {@link TypedQuery} with the given query result type,
+     * which must be a supertype of the result type of this query.
+     * This query must be a Jakarta Persistence {@code SELECT} query
+     * or a native SQL query which returns a result set.
+     * @param resultType The Java class of the query result type
+     * @param <R> The query result type
+     * @throws IllegalArgumentException if the given result type is
+     *         not a supertype of the result type of this query
+     * @throws UnsupportedOperationException if this query is a
+     *         Jakarta Persistence {@code UPDATE} or {@code DELETE}
```
njr-11: goes with prior comment if this seems like a better choice
```suggestion
     * @throws IllegalStateException if this query is a
     *         Jakarta Persistence {@code UPDATE} or {@code DELETE}
```

api/src/main/java/jakarta/persistence/Reference.java
```diff
@@ -0,0 +1,107 @@
...
+     * {@link jakarta.persistence.query.StaticNativeQuery}.
+     */
+    String getName();
+
+    /**
+     * A map keyed by hint name of all hints specified via
+     * {@link NamedQuery#hints}, {@link NamedNativeQuery#hints},
+     * {@link jakarta.persistence.query.ReadQueryOptions#hints}, or
+     * {@link jakarta.persistence.query.WriteQueryOptions#hints}.
+     */
```
njr-11, line 52: To match the other methods,
```suggestion
     * <p>
     * Any attempted mutation of the returned map results in an
     * {@link UnsupportedOperationException}.
     */
```

api/src/main/java/jakarta/persistence/Reference.java
```diff
@@ -0,0 +1,107 @@
...
+     * {@linkplain #getArguments arguments} to parameters of the
+     * statement or query, or {@code null} if no arguments were
+     * supplied. Arguments are present when this is a reference
+     * to a statement or query declared using an annotation of a
+     * method.
+     * <p>
+     * Any mutation of the returned list results in an
+     * {@link UnsupportedOperationException}.
+     *
+     * @since 4.0
```
njr-11, line 66: `@since 4.0` is present on the class, so not needed here.
```suggestion
```

api/src/main/java/jakarta/persistence/Reference.java
```diff
@@ -0,0 +1,107 @@
...
+     * supplied. Arguments are present when this is a reference
+     * to a statement or query declared using an annotation of a
+     * method. If the query has named parameters, these are
+     * interpreted as the parameter names. Otherwise, if the
+     * query has positional parameters, they are ignored.
+     * <p>
+     * Any mutation of the returned list results in an
+     * {@link UnsupportedOperationException}.
+     *
+     * @since 4.0
```
njr-11, line 83:
```suggestion
```

api/src/main/java/jakarta/persistence/Reference.java
```diff
@@ -0,0 +1,107 @@
...
+     *     {@linkplain #getParameterNames parameter names} to
+     *     obtain an assignment of arguments to parameters.
+     * </ul>
+     * <p>
+     * Any mutation of the returned list results in an
+     * {@link UnsupportedOperationException}.
+     *
+     * @see Query#setParameter(int, Object)
+     * @see Query#setParameter(String, Object)
+     * @since 4.0
```
njr-11, line 107:
```suggestion
```

api/src/main/java/jakarta/persistence/Reference.java
```diff
@@ -0,0 +1,107 @@
...
+     * <p>
+     * Any mutation of the returned list results in an
+     * {@link UnsupportedOperationException}.
+     *
+     * @since 4.0
+     */
+    List<String> getParameterNames();
+
+    /**
+     * The arguments supplied to the parameters if the statement
```
njr-11:
```suggestion
     * The arguments supplied to the parameters of the statement
```

api/src/main/java/jakarta/persistence/Statement.java
```diff
@@ -0,0 +1,234 @@
...
+     * @return the same query instance
+     * @throws IllegalArgumentException if the parameter does not
+     *         correspond to a parameter of the query
+     * @since 2.0
+     * @deprecated Newly-written code should use the date/time types
+     *             defined in {@link java.time}.
+     */
+    @Deprecated(since = "3.2") @Override
+    Statement setParameter(Parameter<Date> param, Date value, TemporalType temporalType);
+
```
njr-11, line 148: Avoid deprecated methods on a new interface if possible.
```suggestion
```

api/src/main/java/jakarta/persistence/Statement.java
```diff
@@ -0,0 +1,234 @@
...
+     * @return the same query instance
+     * @throws IllegalArgumentException if the parameter name does
+     *         not correspond to a parameter of the query or if
+     *         the value argument is of incorrect type
+     * @deprecated Newly-written code should use the date/time types
+     *             defined in {@link java.time}.
+     */
+    @Deprecated(since = "3.2") @Override
+    Statement setParameter(String name, Date value, TemporalType temporalType);
+
```
njr-11, line 190:
```suggestion
```

api/src/main/java/jakarta/persistence/Statement.java
```diff
@@ -0,0 +1,234 @@
...
+     * @param temporalType  temporal type
+     * @return the same query instance
+     * @throws IllegalArgumentException if position does not
+     *         correspond to a positional parameter of the query or
+     *         if the value argument is of incorrect type
+     * @deprecated Newly-written code should use the date/time types
+     *             defined in {@link java.time}.
+     */
+    @Deprecated(since = "3.2") @Override
+    Statement setParameter(int position, Date value, TemporalType temporalType);
```
njr-11, line 233:
```suggestion
```

api/src/main/java/jakarta/persistence/StatementReference.java
```diff
@@ -0,0 +1,87 @@
...
+ * SPDX-License-Identifier: EPL-2.0 OR BSD-3-Clause
+ */
+
+// Contributors:
+//     Gavin King      - 4.0
+
+package jakarta.persistence;
+
+/**
+ * A reference to an executable named query declared via the
```
njr-11: Shorter summary sentence and mention statement. The prior wording was general an could have applied to other queries as well.
```suggestion
 * A reference to an executable named statement for an operation
 * that returns a row count. The named statement is declared via the
```

api/src/main/java/jakarta/persistence/StatementReference.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ *     @StaticQuery("delete from Record where temporary = true")
+ *     int purgeTemporaryRecords() {
+ *         return entityManager.createStatement(Filer_.purgeTemporaryRecords())
+ *                 .execute();
+ *     }
+ * }
+ * }
+ *
+ * <p>In this example, it is the entity class that is annotated,
```
njr-11:
```suggestion
 * <p>In this example, the entity class is annotated,
```

api/src/main/java/jakarta/persistence/StatementReference.java
```diff
@@ -0,0 +1,87 @@
...
+ * <li>A reference representing a query declared using an
+ *     annotation of a method holds information about the
+ *     arguments passed to the method, making it available via
+ *     {@link #getArguments()}, {@link #getParameterNames()},
+ *     and {@link #getParameterTypes()}.
+ * </ul>
+ *
+ * <p>In the Jakarta Persistence query language, a DELETE or
+ * UPDATE statement is not a typed query, and is always
+ * represented by an instance of {@link StatementReference}.
```
njr-11:
```suggestion
 * UPDATE statement is not a typed query and is always
 * represented by an instance of {@link StatementReference}.
```

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -26,16 +26,47 @@
 
 /**
  * Interface used to control the execution of typed queries.
+ * In the Jakarta Persistence query language, only SELECT
+ * queries are typed queries, since only a SELECT query can
+ * return a result. A DELETE or UPDATE query is not a typed
+ * query, and is always represented by an untyped instance
```
```suggestion
 * query and is always represented by an untyped instance
```

api/src/main/java/jakarta/persistence/TypedQueryReference.java
```diff
@@ -78,38 +78,37 @@
  *     and {@link #getParameterTypes()}.
  * </ul>
  *
+ * <p>In the Jakarta Persistence query language, only SELECT
+ * queries are typed queries, since only a SELECT query can
+ * return a result. A DELETE or UPDATE statement is not a
+ * typed query, and is always represented by an untyped
```
njr-11:
```suggestion
 * typed query and is always represented by an untyped
```

api/src/main/java/jakarta/persistence/query/StaticStatementReference.java
```diff
@@ -0,0 +1,121 @@
...
+package jakarta.persistence.query;
+
+import jakarta.persistence.StatementReference;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * A reference to a query declared using a {@link StaticQuery}
```
njr-11: Should not be identical to StaticTypedQueryReference's summary sentence,
```suggestion
 * A reference to a statement declared using a {@link StaticQuery}
```

api/src/main/java/jakarta/persistence/query/StaticStatementReference.java
```diff
@@ -0,0 +1,121 @@
...
+        this.hints = hints;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public Map<String, Object> getHints() {
```
njr-11, line 123: These methods should probably have Javadoc copied over, but this PR already has enough in it. That can be done separately.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2744,35 +2757,45 @@ specified for a query by calling `setLockMode()` or by specifying the
...
+In particular, problems may occur if the stored procedure initiates a
+transaction and a transaction is already active. The state of entities
+returned by a stored procedure query invocation is determined according
+to the rules for `Query` or `TypedQuery` as described above.
+
+If the persistence context is joined to the transaction, any runtime
+exception thrown by any method not explicitly listed below of `Query`,
+`Statement`,`TypedQuery`, or `StoredProcedureQuery` causes the current
+transaction to be marked for rollback, unless the exception is an
+instance of one of the following exception types:
```
njr-11: I was confused by this because "any method not explicitly listed below" does not immediately follow and initially seemed to be missing. Further on, I spotted the list of methods and then it made sense. This suggestion tries to adjust the wording to better prepare the reader,
```suggestion
exception thrown by any method of `Query`, `Statement`,`TypedQuery`,
or `StoredProcedureQuery` not explicitly listed later within this section
causes the current transaction to be marked for rollback, unless the
exception is an instance of one of the following exception types:
```

api/src/main/java/jakarta/persistence/Query.java
gavinking: I dunno, that's a question I had myself. I feel like `UnsupportedOperationException` is appropriate when "you can't do this operation with this object because it's the wrong kind of thing".

api/src/main/java/jakarta/persistence/Reference.java
gavinking: I need a way to distinguish the "new" operations inherited by `TypedQueryReference` from the ones which existed in 3.2.

api/src/main/java/jakarta/persistence/Statement.java
gavinking: They're covariant overrides. I think they're needed.

gavinking:
Thanks Nathan, so, now that you've looked over this, would you please also take a look at https://github.com/jakartaee/persistence/pull/858 and try to form some sort of opinion on which is a better path?

api/src/main/java/jakarta/persistence/Query.java
njr-11: I'm fine with either one.  I posted the comment after seeing that another method of the class used IllegalStateException.  The more important thing is to remain consistent across methods (same comment also posted to #858)

njr-11:
> Thanks Nathan, so, now that you've looked over this, would you please also take a look at #858 and try to form some sort of opinion on which is a better path?

At this point, I'm leaning toward #858 if we can do it without adding the `getResults()` method that duplicates `getResultList()`.

gavinking:
> At this point, I'm leaning toward #858 if we can do it without adding the `getResults()` method that duplicates `getResultList()`.

I don't think we really can. I struggled with this for days, and I found two self-consistent approaches:

1. Migrate `Query` toward being an abstract supertype of "result queries" (i.e. `TypedQuery`) and "mutations" (i.e. `Statement`). And that's what we have here. We deprecate half the methods of `Query` and introduce new `createStatement()` methods, etc.
2. Or we just focus on those two bad methods,`getResultList()` and `getResultStream()`, and avoid changing the whole nature of `Query`, at the cost of one dupe method.

I think if `getResults()` were a worse name than `getResultList()`, I would like option 2 a lot less. But since I actually think that `getResults()` is a *better* name, I can sorta maybe live with the duplication.

gavinking:
After I originally did this work, I was a bit horrified at the blast radius, and it turned me against the whole concept, and sent me looking for other alternatives like #858 and #872.

But last night I sat down and looked over what I did here, and I didn't hate it anymore. I'm still not sure that this is where I want to spend our complexity budget, but it might be. 

What I'm not sure is what is the path forward. We need some sort of a process to:

1. reach a consensus that one of #842, #858, or #872 is worth doing, and
2. evaluate its impact on existing implementations.

I don't want to do anything here that causes chaos for vendor APIs which extend `jakarta.persistence.Query`. Of course I'm thinking especially of Hibernate here @sebersole. Usually take the point of view that we should do what is best and makes most sense for the Persistence spec, and that Hibernate has to adapt. But this is a rather special case, and here I *am* going to pay close attention to the impact on Hibernate users (because there are simply so many of them).

@sebersole @njr-11 @beikov I'm going to need to guys to think carefully about this stuff. Steve, I think we need to try doing some spikes to evaluate the impact on Hibernate.

gavinking:
P.S. If we're going to do anything here, then I would like *something* to go into M1, so that we have time to 

- properly evaluate it and be sure it's not a mistake and
-  change or reverse course if necessary.

api/src/main/java/jakarta/persistence/Query.java
gavinking: I changed to `IllegalStateException`.

sebersole:
I like this approach generally over the others I've seen here.  Maybe because this is largely the direction we went in Hibernate breaking these down into `SelectionQuery` and `MutationQuery` which would now subtype `TypedQuery` and `Statement` nicely (I prefer the name `MutationQuery`, but meh).  But to be fair, we went that direction because we thought that made the most sense. 

I'll spend some time next week looking this over closer.

sebersole:
> We are definitely not going to call anything `MutationQuery`, because that collides with a name already used in Hibernate.

To be fair, we have the same "problem" with `Query`.  Its not a huge deal imo.

api/src/main/java/jakarta/persistence/Query.java
beikov: I'm a bit torn on this method, since it might require a re-interpretation of the underlying query by the implementor. I would prefer to not have the method and guide users to using the `createQuery` method that takes the result class directly.
Maybe it's no big deal and we can do this somewhat efficiently, but is there an actual need for this method or is it just a "convenience"?

api/src/main/java/jakarta/persistence/Reference.java
```diff
@@ -0,0 +1,121 @@
...
+// Contributors:
+//     Gavin King      - 4.0
+
+package jakarta.persistence;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * A reference to a named statememt or query declared via the
```
beikov:
```suggestion
 * A reference to a named statement or query declared via the
```

api/src/main/java/jakarta/persistence/Query.java
gavinking: > I'm a bit torn on this method, since it might require a re-interpretation of the underlying query by the implementor.

That is a very good point, and one which I had not thought of.

So we do need to think through the corner cases.

That said, I think what we're saying here is that use of `Query` for directly executing methods is now deprecated. Given that, it would be perfectly OK if some validation doesn't happen until execution time if you execute directly via `Query`.

But I need more time to think about this more carefully.

>  I would prefer to not have the method and guide users to using the `createQuery` method that takes the result class directly.

The thing is that I actually hate the signature of that method, because it takes the result class as the second parameter, making it really difficult to format code nicely. So I actually really like this method because it makes the code prettier.

sebersole:
Any chance to get this merged?  Its the last (hopefully) thing holding me up for integrating 4.0 into Hibernate.

gavinking:
I guess I was mostly waiting for your go-ahead. If you're telling me "go" then I guess I will go ahead and merge this.

sebersole:
Overall I really like the change.  I thought I had commented to that effect here, but I cannot see it (was probably on one of the other "change Query" issues).

There are some things I would do different following what I did in Hibernate - mainly I'd get away from `withResultType()` and simply have proper methods to create the correct things up front.   But that would require more overloads and we seem to shy away from that in persistence.

gavinking:
Alright, I will merge it. As I said, this is still a trial balloon to see what feedback we get and what the fallout looks like.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: @njr-11 See b914f75. Hope that helps a bit.

PR #845, opened 2025-12-18
javadoc about threadsafety of EM and EA

gavinking:
It's really bad that this information has always been missing from the documentation of `EntityManager`.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityAgent.java
+++ b/api/src/main/java/jakarta/persistence/EntityAgent.java
@@ -24,7 +24,7 @@ import java.util.List;
  * possible with {@link EntityManager}.
  *
  * <p>An instance of {@code EntityAgent} must be obtained from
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -24,9 +24,24 @@
  * possible with {@link EntityManager}.
  *
  * <p>An instance of {@code EntityAgent} must be obtained from
- * an {@link EntityManagerFactory} and is able to execute
- * persistence-related operations for entity types belonging to the
- * associated persistence unit.
+ * an {@link EntityManagerFactory}, and is only able to manage
```
```suggestion
 * an {@link EntityManagerFactory} and is only able to manage
```

njr-11 (approved):
Good idea to add this! I'm surprised that was not already stated.  I noted one minor grammar improvement if you want it.  Approving either way.

PR #848, opened 2025-12-18
TCK updates

marko-bekhta:
Since I saw we are getting (https://ci.eclipse.org/jpa/job/jakarta-persistence-api-publish-snapshot/job/main/5/pipeline-overview/?selected-node=49) 

>  - Namespace 'jakarta.tck' is not allowed

I realized that we aren't using the `jakarta.persistence` package for the tck subproject... if we want to publish it as part of the Persistence project we should keep the same group id....

While at it I've noticed a few more inconsistencies and made a few other changes (can be reviewed by commit)

- move the flattened poms in TCK to `/target` so that we don't end up with these poms in the work dir and make them "visible" to Git
- changed the group and artifact ids for all TCK modules to follow the same structure.
- bumped JUnit to 6, while unnecessary, it's the recent version that people are gradually moving towards, so we might as well bump the version so it would be easier to match it in the spec implementation projects.
- updated the GitHub workflow to build the TCK subproject (skipping the test run) so that we at least can see we can build it when we submit changes
- bumped the Jakarta Persistence version in TCK to 4.0. there were a few changes already that affected the TCK code. I think it would be good to see the effect of the spec/api updates on the TCK + we can also start gradually add new tests.

### DIFF
--- a/tck/common/src/main/java/ee/jakarta/tck/persistence/common/pluggability/altprovider/implementation/EntityManagerFactoryImpl.java
+++ b/tck/common/src/main/java/ee/jakarta/tck/persistence/common/pluggability/altprovider/implementation/EntityManagerFactoryImpl.java
@@ -16,21 +16,27 @@
 
 package ee.jakarta.tck.persistence.common.pluggability.altprovider.implementation;
 
+import java.lang.annotation.Annotation;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
 import jakarta.persistence.Cache;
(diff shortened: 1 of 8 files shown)


gavinking:
Shall I just merge this Marko?

marko-bekhta:
if you think that  the groupid/artifact id in TCK are reasonable, yes 👍🏻(I matched the format of Jakarta Validation TCK)
once we get the TCK snapshots published, I'll try to set up something on the impl (Hibernate) side for us to run the TCK, at least in the similar way we've done it for 3.2

PR #851, opened 2025-12-18
improve javadoc for EntityGraph

gavinking:


api/src/main/java/jakarta/persistence/EntityGraph.java
```diff
@@ -16,14 +16,54 @@
...
  * <p>
- * The methods to add subgraphs implicitly create the corresponding
- * attribute nodes as well; such attribute nodes should not be
- * redundantly specified.
+ * An entity graph is usually used to control the data fetched from
+ * the database. The {@linkplain AttributeNode attribute nodes} and
+ * {@linkplain Subgraph subgraphs} of an entity graph determine the
+ * limits of the graph of associated attributes and entities fetched
+ * when an operation which retrieves an instance or instances of the
+ * root entity of the graph is executed.
```
njr-11: This is a very long sentence that is difficult to read correctly.  The following might help.  Otherwise splitting into 2 sentences might be an option, but I didn't think of a good way to do that.
```suggestion
 * the database. The {@linkplain AttributeNode attribute nodes} and
 * {@linkplain Subgraph subgraphs} of an entity graph determine the
 * limits of the graph of associated attributes and entities fetched
 * when a query retrieves an instance or instances of the root entity
 * of the graph.
```

gavinking: I want to avoid the word "query" here because the spec generally uses "query" as a synonym for `Query`/`TypedQuery`, whereas here we're usually talking about the `find()` method.

gavinking: On the other hand, I agree with your suggestion that "is executed" isn't doing any useful work here.

```suggestion
 * the database. The {@linkplain AttributeNode attribute nodes} and
 * {@linkplain Subgraph subgraphs} of an entity graph determine the
 * limits of the graph of associated attributes and entities fetched
 * when an operation retrieves an instance or instances of the root 
 * entity of the graph.
```

gavinking:
Thanks Nathan.

PR #852, opened 2025-12-18
added missing comment/check-constraint elements to XML

gavinking:


gavinking:
@anija-anil would you do the favor of reviewing this one for me please?

sebersole:
Haha, I just started a branch for this too.

Your changes are matching up closely to what I have so far.  I noticed a few additional ones that I caan send separately after this is applied.

sebersole:
One general thing I'll mention is regarding comment.  The JPA XSD uses a frustrating mix of element and attribute for this.

Your PR takes the approach of using element all over.  In my brief look, the way the XSD has historically handled this is to use element for comments associated with tables and attribute for comments aassociated with columns.  To be honest, I prefer the element approach so I am not necessarily arguing about it; but we might want to remain consistent in that approach.

gavinking:
I think we should merge this, because @sebersole is already working with this stuff on the implementation side and, well, it's essentially just mechanical work of syncing with the annotations. Steve has already reviewed it now (and spotted something missing  in `<named-query>`). And I have looked over it several times carefully myself.

sebersole (approved):

sebersole:
I almost hate to mention this one, but another discrepency is in regards to `@MapKey#value` / `#name`.  We have not changed the XSD to reflect this change in the annotation.

To be clear, I do not think we should - `#value` is simply because of a "feature" of annotations.

At the least though, we *should* update the annotation copy in the XSD and maybe make a note of this in the XSD.

gavinking:
OK

gavinking:
Done.

anija-anil:
The PrimaryKeyJoinColumn Java annotation includes the attributes comment() and check(). These attributes are currently missing from the XSD changes in this PR.

api/src/main/resources/jakarta/persistence/orm_4_0.xsd
```diff
@@ -2146,13 +2179,18 @@
...
         }
 
       </xsd:documentation>
     </xsd:annotation>
     <xsd:sequence>
+      <xsd:element name="comment" type="xsd:string" minOccurs="0" />
       <xsd:element name="foreign-key" type="orm:foreign-key"
                    minOccurs="0"/>
+      <xsd:element name="check-constraint" type="orm:check-constraint"
+                   minOccurs="0" maxOccurs="unbounded"/>
```
gavinking, line 2193: @anija-anil They are here. Just hard to see from the diff.

anija-anil: Oh okay .Sorry for the noise on my end

anija-anil (approved):
LGTM

gavinking:
Great, thanks for the review, Anija!

PR #858, opened 2025-12-20
a less radical proposal for the Query hierarchy

gavinking:
Competes with #842

This one:

- Adds `ofType()` and `withEntityGraph()` to Query, because they're nice.
- Deprecates `getResultList()` and `getResultStream()` on Query.
- Adds a new method `getResults()` returning `List<?>` on `Query`, overridden to `return List<T>` on `TypedQuery`.
- Introduces `QueryReference`, without saying it always represents a statement.


api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -37,6 +53,67 @@
...
+    <R> TypedQuery<R> withEntityGraph(EntityGraph<R> graph);
+
+    /**
+     * Execute a SELECT query and return the query results as an untyped
+     * {@link List}. If necessary, first synchronize changes with the
+     * database by flushing the persistence context.
+     * @return a list of the results, or an empty list if there are
+     *         no results
+     * @throws IllegalStateException if called for a Jakarta
+     *         Persistence query language UPDATE or DELETE statement
```
njr-11: The prior two methods are using UnsupportedOperationException for this, and then this and the following method use IllegalStateException. It doesn't matter to me which type is chosen between the two, but ideally we would be consistent across methods.

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -26,7 +26,12 @@
 
 /**
  * Interface used to control the execution of typed queries.
- *
+ * In the Jakarta Persistence query language, only SELECT
+ * queries are typed queries, since only a SELECT query can
+ * return a result. A DELETE or UPDATE query is not a typed
+ * query, and is always represented by an untyped instance
+ * of {@link Query}. On the other hand, a native SQL query
+ * is considered a typed query if it returns a result set. *
```
```suggestion
 * is considered a typed query if it returns a result set.
 *
```

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -61,9 +66,42 @@ public interface TypedQuery<X> extends Query {
      * @throws PersistenceException if the flush fails
      * @throws OptimisticLockException if an optimistic locking
      *         conflict is detected during the flush
+     * @deprecated This method is deprecated in the supertype.
+     *             Use {@link #getResults()} instead.
```
njr-11: This doesn't seem very nice.  I'd rather not break applications that have been correctly using the typed getResultList method of TypedQuery. 

```suggestion
```

gavinking:
> I have a hard time justifying the removal of `TypedQuery.getResultList`, which is currently the pattern that we want applications using. Do we really need the new `List<?> getResults()` method?

The question is the following: how can we migrate people away from that raw `List`?

No, we don't *have* to deprecate `TypedQuery.getResultList`, we can deprecate the method on the supertype, and leave in place the override in an undeprecated form. I toyed with that.

The only problem is that we then end up, long term, with two identical methods `getResults()` and `getResultList()` on `TypedQuery`. That's not the *worst* thing ever. (It's better than the situation we have today.) And it might be the best thing to do. I just wasn't sure.

api/src/main/java/jakarta/persistence/TypedQuery.java
gavinking: We don't have to do this.

api/src/main/java/jakarta/persistence/Query.java
gavinking: I don't much care which one it is either. But looking at the docs of the two exception types, which could *really* be a lot better, it really does seem to better fit  UOE.

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -37,6 +53,67 @@
...
+     */
+    <R> TypedQuery<R> withEntityGraph(EntityGraph<R> graph);
+
+    /**
+     * Execute a SELECT query and return the query results as an untyped
+     * {@link List}. If necessary, first synchronize changes with the
+     * database by flushing the persistence context.
+     * @return a list of the results, or an empty list if there are
+     *         no results
+     * @throws IllegalStateException if called for a Jakarta
```
gavinking:
```suggestion
     * @throws UnsuportedOperationException if called for a Jakarta
```

api/src/main/java/jakarta/persistence/Query.java
gavinking: Oh, well, I guess we have to start consistent with `executeUpdate()`:

```java
     * @throws IllegalStateException if called for a Jakarta
     *         Persistence query language SELECT statement or for
     *         a criteria query
```

api/src/main/java/jakarta/persistence/TypedQuery.java
gavinking: Hah, actually I had removed the `@Deprecated` annotation but accidentally left the `@deprecated` tag there. Fixed.

api/src/main/java/jakarta/persistence/Query.java
gavinking: Fixed.

gavinking:
@njr-11 I have made both of your suggested changes. WDYT now?

njr-11:
> The only problem is that we then end up, long term, with two identical methods `getResults()` and `getResultList()` on `TypedQuery`. That's not the _worst_ thing ever. (It's better than the situation we have today.) And it might be the best thing to do. I just wasn't sure.

I'm not sure either.  I'm still wondering if we can avoid adding `getResults()` to avoid the duplication.  But otherwise, if we need to keep it, this PR looks good now.

gavinking:
> I'm still wondering if we can avoid adding `getResults()` to avoid the duplication

That's what I tried with the competing PR, but I feel like that one just got out of hand. Removing `getResultList()` from `Query` without replacing it with something equivalent just changes the nature of `Query` into something different to what it was before. And that pushes you toward something that's more than I really want to do here.

The only other option I see here is to just bite the bullet and change that return type to `List<?>`, and break everyone's programs. I mean, they kinda deserve it for ignoring the warnings about the raw type for two decades but ... I dunno, I still don't like it.

gavinking:
After several backchannel discussions with various people, the sense I took away was: "if we're going to do this, do #842".

I'm therefore closing this PR.

gavinking:
> Do we really need the new `List<?> getResults()` method?

FTR, this was the strongest argument against this option. This, and the fact that `TypedQuery` continues to have an `executeUpdate()` method into the infinite future.

PR #859, opened 2025-12-20
clean up some writing in the spec (ch1, ch2, ch11)

gavinking:
This pull request contains a raft of changes to improve the readability and understandability of the spec.

It focuses on important parts of ch2 and ch11.

A future pull request will do the same work for chapter 3, and, possibly, for ch7.

Note that:

-  ch4 is going away; 
- chs 6 and 12 are uninteresting; 
- chs 8 and 9 seem fine more or less as they are, even though ch9 has a weird structure.

So I don't plan to invest time in rewriting these chapters.

But chapters 2 and 3 in particular are very important and deserve a long-overdue refresh.

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -324,11 +324,11 @@ only at the inverse end of an association. Failure to modify the owning
 side of a relationship typically results in the change being lost.
 ====
 
-The persistence provider runtime is permitted to perform synchronization
-to the database at other times, as long as a transaction is active and
-the persistence context is joined to the transaction. The application
-can influence the timing of synchronization via the `flush()` method of
-`EntityManager` or by setting the `FlushModeType`.
+When a transaction is active and the persistence context is joined to
+the transaction, the persistence provider runtime is permitted to flush
+pending modifications to the database. The application can influence the
+timing of synchronization via the `flush()` method of `EntityManager` or
+by setting the `FlushModeType`.
 
 - When the `flush()` method is called, the provider must synchronize
   every pending modification made to every entity currently associated
@@ -344,7 +344,7 @@ can influence the timing of synchronization via the `flush()` method of
 
 If there is no transaction active or if the persistence context has not
 been joined to the current transaction, the persistence provider must not
-flush to the database.
+flush modifications to the database.
 
 The semantics of the flush operation, applied
 to an entity X are as follows:


spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -86,17 +90,16 @@ property access or field access is used:
   via the property accessor methods. Every property not annotated with the
   `Transient` annotation is persistent.
 
-Mapping annotations must not be applied to fields or properties marked
-`transient` or `Transient`, since those fields and properties are not
-persistent.
+Mapping annotations must not be applied to a field or property marked `static`,
+`transient`, or `@Transient`, since the field or property is not persistent.
```
marko-bekhta:
```suggestion
`transient`, or `@jakarta.persistence.Transient`, since such field or property is not persistent.
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -112,38 +115,116 @@ Use of the generic variants of these collection types is strongly encouraged,
...
 (or detached), subsequent access to the collection must be through the
-interface type.
-
-==== Persistent Attribute Type
+interface type. The persistence provider runtime is permitted to replace,
+at any time, any reference held by an entity to a collection instantiated by
+the application with a different provider-specific object implementing the
+same interface type.
+
+==== Persistent Attribute Type [[categories]]
```
marko-bekhta: maybe 
```suggestion
==== Persistent Attribute Type [[attribute-type-categories]]
```
just thinking that `categories` is too ambiguous especially if you think of it as an anchor in the shared URL. 
btw on a related note, how does this play with all those other `a12345` anchors ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -112,38 +115,116 @@ Use of the generic variants of these collection types is strongly encouraged,
...
+
+Every persistent field or property of an entity is either:
+
+- an attribute of **basic type**, as defined in <<a486>>, or of a type
+  converted to a basic type through use of a _converter_, as defined
```
marko-bekhta, line 139:
```suggestion
  converted to a basic type through use of a converter, as defined
```
just checking, since I think you were saying the _italics_  were for defining a term 🫣

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -112,38 +115,116 @@ Use of the generic variants of these collection types is strongly encouraged,
...
+- An attribute annotated `@Embedded` is an embedded attribute.
+- An attribute annotated `@ElementCollection` is an element collection.
+- An attribute annotated `@ManyToOne`, `@OneToOne`, `@OneToMany`, or
+  `@ManyToMany` is an association of the indicated multiplicity.
+
+If none of the above annotations occur on a persistent attribute,
+the defaulting rules specified below in <<a511>> determine which
+category it belongs to.
+
+An attribute of an entity must not have with mapping annotations
```
marko-bekhta:
```suggestion
An attribute of an entity must not have mapping annotations
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -112,38 +115,116 @@ Use of the generic variants of these collection types is strongly encouraged,
...
+the assigned value, preserves the identity of the instance.
+
+In general, operations performed by a persistence provider are not
+required to preserve the object identity of a value stored in an
+attribute of an entity. The provider is permitted to modify the
+value held in the attribute to a new value with a distinct object
+identity. Such modifications should never alter the _persistent_
+representation of the entity because the Java object identity of
+a value held in an attribute is never part of the persistent
+state of the entity.
```
marko-bekhta:
```suggestion
identity. Such modifications must never alter the _persistent_
representation of the entity, which is possible because the Java
object identity of a value held in an attribute is never part of the
persistent state of the entity.
```
maybe ^ ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -112,38 +115,116 @@ Use of the generic variants of these collection types is strongly encouraged,
...
+
+In particular:
+
+- A basic type, an embeddable class, or any collection type is
+  considered a _value type_, and the provider is not required
+  to respect the integrity of shared references to an instance
+  of such a type. If multiple fields referenced the same instance
+  of a value type when an entity was made persistent, those fields
+  hold references to distinct instances of the value type when the
+  entity is reloaded from the database.
```
marko-bekhta:
```suggestion
  of such a type. If multiple fields referenced the same instance
  of a value type when an entity was made persistent, those fields
  may hold references to distinct instances of the value type when 
  the entity is reloaded from the database.
```
since I don't think that we are enforcing that those references must be distinct, no ?

spec/src/main/asciidoc/ch02-entities.adoc
marko-bekhta, line 1330: since we are providing a list of basic types that are allowed for the version how would application be able to declare it on any other type ? The way I read this is that the app would just fail so there's no concern about "portability" no ?
or if persistent providers are allowed to extend that list of types ... maybe we should change the wording to:
> The version should be of one of the following basic types:

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+- an _embedded attribute_, that is, single-valued field or property
+  whose type is an embeddable class, or even
+- a collection-valued attribute whose element type is an embeddable
+  class.
+
+An embeddable type may even be used to represent the key or value of
+a map.
+
+An embedded object belongs strictly to its owning entity instance and
+is not sharable across persistent entity instances; sharing of an
```
marko-bekhta:
```suggestion
is not shareable across persistent entity instances; sharing of an
```
no?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
-An embeddable class (including an
-embeddable class within another embeddable class) may contain a
-collection of a basic type or other embeddable
-class.footnote:[Direct or indirect
-circular containment dependencies among embeddable classes are not
-permitted.]
+- An embeddable class may have an embedded field or property. However,
+  an embeddable class may not occur, directly or indirectly, as the type
+  of one of its own embedded attributes. Circular containment dependencies
+  among embeddable classes are not permitted.
```
marko-bekhta, line 1410: > Circular containment dependencies among embeddable classes are not permitted.

any specific expectations from the provider ? throw a particular exception maybe ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+An element collection must have declared type `java.util.Collection`,
+`java.util.Set`, `java.util.List`, or `java.util.Map`. Collections of
+type `Map` are special, and are described below in <<map-collections>>.
+
+When an element collection has declared type `Collection<E>`, `Set<E>`,
+or `List<E>`, the type argument `E` is taken, by default, to be its
+element type. The element type may be explicitly specified using the
+`targetClass` member of the `ElementCollection` annotation. If an
+element collection is declared with a raw type, the `targetClass`
+member must be explicitly specified.
```
marko-bekhta: should we specify what's the expectation when the target class and type variable are incompatible?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
-persistent field or property of the entity that is the value of the map.
-The `MapKeyClass` annotation is not used when `MapKey` is specified.
+An element collection may be of type `java.util.Map`. Such a map
+collection has two different element types:
+
+- a _key type_ and
+- a _value type_.
+
+A key or value type may be a basic type, an embeddable class,
+or an entity type. The key type and map type are completely
```
marko-bekhta:
```suggestion
or an entity type. The key type and value type are completely
```
?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+independent of one another.
+
+A map collection must be declared using an `ElementCollection`,
+`OneToMany`, or `ManyToMany` annotation:
+
+- when the map value type is a basic type or embeddable class,
+  the `ElementCollection` annotation is used;
+- but when the map value type is an entity, a `OneToMany` or
+  `ManyToMany` annotation is used.
```
marko-bekhta, line 1487: should we state explicitly that map key has no influence ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+- if the map represents the collection-valued side of a bidirectional
+  one-to-many/many-to-one relationship, then, by default, the map maps
+  to the primary table of the value entity type.
+
+In this case, default mappings are described in <<a538>>.
+
+==== Map Keys
+
+When an element collection has declared type `Map<K,V>`, the type
+argument `A` is taken, by default, to be its key type. The key type
```
marko-bekhta:
```suggestion
argument `K` is taken, by default, to be its key type. The key type
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+an element collection is declared with a raw type, the `MapKeyClass`
+annotation is required.
+
+When the map value type is an entity type, the `MapKey` annotation
+specifies that the map key is itself the primary key or a persistent
+field or property of the map value entity.
+
+The `MapKey` and `MapKeyClass` annotations are mutually exclusive;
+an element collection may have either a `@MapKey` or `@MapKeyClass`
+annotation, or neither, but not both.
```
marko-bekhta, line 1538: so idk how you feel about consistency in these annotation types mentioned in the docs. I noticed that throughout the spec we have a mix of `@AnnotationName` / `AnnotationName` / `@AnotationFQCN`

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+
+Dedicated annotations are provided for specifying column mappings
+for map keys:
+
+- If the map key type is a basic type, the `MapKeyColumn` annotation
```
marko-bekhta, line 1543: e.g. here we are introducing the annotation so I'd probably go with the fqcn.. but I'll leave it up to you to decide how you prefer to deal with the annotations 🙂

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+has undefined semantics and should be avoided.].
 
 === Mapping Defaults for Non-Relationship Fields or Properties [[a511]]
 
-If a persistent field or property other than a relationship property is
-_not_ annotated with one of the mapping annotations defined in <<a13915>>
-(and no equivalent mapping information is specified in any XML descriptor),
-the following default mapping rules are applied in order:
+If a persistent field or property other than a relationship property
+is _not_ annotated with one of the mapping annotations listed above in
```
marko-bekhta, line 1572:
```suggestion
is not annotated with one of the mapping annotations listed above in
```

? (but again ... I may have misinterpreted the rule on italics 🙂 🫣)

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+<<categories>>, and if no equivalent mapping information is supplied in
+any XML descriptor, the following defaulting rules are applied in order:
```
marko-bekhta:
```suggestion
any XML descriptor, or in a provider-specific way, the following defaulting rules are applied in order:
```
maybe ? (that way providers may have their own ways like let's say programmatic mapping)

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
-in the database (e.g., by foreign key constraints).
+Every relationship (or association) between two entity types has
+well-defined _multiplicity_, either:
+
+- **one-to-one**,
+- **one-to-many** / **many-to-one** (depending on which side the relationship
+  is viewed from), or
+- **many-to-many**.
+
+An attribute which represents one side of such an association must
```
marko-bekhta:
```suggestion
An attribute that represents one side of such an association must
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
+additional mapping strategies:
+
+* A unidirectional one-to-many relationship may be mapped to a foreign key
+  column of the primary table of the target entity. The `JoinColumn`
+  annotation found in <<a14922>> or equivalent XML element must be used to
+  indicate this mapping strategy.
+
+* A unidirectional or bidirectional one-to-one relationship,
+  bidirectional many-to-one/one-to-many relationship, or unidirectional
+  many-to-one relationships may be mapped to a join table. The `JoinTable`
```
marko-bekhta:
```suggestion
  many-to-one relationship may be mapped to a join table. The `JoinTable`
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2066,33 +2262,32 @@ primary key column of table `PATENT`.
...
-`Entity` annotation, mapped as entities, and queried for as entities.
+- any association targeting the second entity class is by nature a
+  polymorphic association, and may reference an instance of the first
+  entity class, and
+- any Jakarta Persistence query against the second entity class is also
+  polymorphic and may return an instance of the first entity class.
 
-Entities can extend non-entity classes and
-non-entity classes can extend entity classes.
+An abstract class may be annotated `Entity` or declared as an `entity`
```
marko-bekhta, line 2279:
```suggestion
An abstract class may be annotated with `Entity` or declared as an `entity`
```
?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2066,33 +2262,32 @@ primary key column of table `PATENT`.
 
 === Inheritance [[a765]]
 
-An entity may inherit from another entity
-class. Entities support inheritance, polymorphic associations, and
-polymorphic queries.
+An entity class might inherit another entity class. If this is the case,
```
marko-bekhta, line 2269:
```suggestion
An entity class might extend another entity class. If this is the case,
```
? (to be more within Java lang)

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2066,33 +2262,32 @@ primary key column of table `PATENT`.
...
+then:
 
-Both abstract and concrete classes can be
-entities. Both abstract and concrete classes can be annotated with the
-`Entity` annotation, mapped as entities, and queried for as entities.
+- any association targeting the second entity class is by nature a
+  polymorphic association, and may reference an instance of the first
+  entity class, and
+- any Jakarta Persistence query against the second entity class is also
+  polymorphic and may return an instance of the first entity class.
```
marko-bekhta: maybe we should use extended/super (or similar) rather than first/second (it's not 100% clear who's first and second are here 🙂)?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2066,33 +2262,32 @@ primary key column of table `PATENT`.
...
-Entities can extend non-entity classes and
-non-entity classes can extend entity classes.
+An abstract class may be annotated `Entity` or declared as an `entity`
+in its XML descriptor. An abstract entity class may participate in a
+relationship, and may occur in the `from` clause of a query.
 
-These concepts are described further in the
-following sections.
+An entity might extend a non-entity class, and a non-entity class might
+extend an entity class.
```
marko-bekhta: any expectations/limitations when we do so? (maybe worth pointing to some further sections `Non-Entity Classes in the Entity Inheritance Hierarchy` ?)

marko-bekhta: ahh ok, maybe pull the next paragraph into this one?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2142,40 +2337,45 @@ public class PartTimeEmployee extends Employee {
 
 ==== Mapped Superclasses
 
-An entity may inherit from a superclass that
-provides persistent entity state and mapping information, but which is
-not itself an entity. Typically, the purpose of such a mapped superclass
-is to define state and mapping information that is common to multiple
-entity classes.
+An entity might inherit a superclass that declares persistent entity state and
```
marko-bekhta, line 2349: inherit vs extend ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2142,40 +2337,45 @@ public class PartTimeEmployee extends Employee {
...
+mapping information, but which is not itself an entity. Typically, the purpose
+of such a _mapped superclass_ is to define state and mapping information that
+is common across multiple entity classes but which maps to disparate, unrelated
+database tables.
+
+A mapped superclass must be annotated with the `MappedSuperclass` annotation or
+declares as a `mapped-superclass` in its XML descriptor.
```
marko-bekhta:
```suggestion
declared as a `mapped-superclass` in its XML descriptor.
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2142,40 +2337,45 @@ public class PartTimeEmployee extends Employee {
...
-`Query` operations. Persistent relationships defined by a mapped
-superclass must be unidirectional.
+A mapped superclass, unlike an entity, is not queryable and must not be passed
+as an argument to `EntityManager`, `EntityAgent`, or `Query` operations.
 
-Both abstract and concrete classes may be
-specified as mapped superclasses. The `MappedSuperclass` annotation (or
-`mapped-superclass` XML descriptor element) is used to designate a
-mapped superclass.
+Persistent relationships involving by a mapped superclass must be unidirectional.
```
marko-bekhta:
```suggestion
Persistent relationships involving a mapped superclass must be unidirectional.
```

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2239,27 +2439,26 @@ public class PartTimeEmployee extends Employee {
...
-annotations on such superclasses are ignored.
+When a non-entity class is annotated with an annotation type defined by this
+specification, the behavior is undefined. The persistence provider might simply
+ignore such annotations.
 
-Non-entity classes cannot be passed as
-arguments to methods of the `EntityManager`, `EntityAgent`, or `Query`
-interfacesfootnote:[This includes
-instances of a non-entity class that extends an entity class.] and cannot bear mapping information.
+A non-entity class cannot be passed as an argument to a method of the `EntityManager`, `EntityAgent`, or `Query` interfacefootnote:[This includes
```
marko-bekhta: it's also not queryable ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2305,106 +2504,105 @@ public class ShoppingCart extends Cart {
...
+* For the **joined subclass** strategy, any field which is declared by
   a subclass is mapped to a separate table from the fields which are
   declared by its superclass. Table joins may be used to retrieve all
   fields declared and inherited by the subclass.
 
-An implementation is required to support the single table per class
-hierarchy inheritance mapping strategy and the joined subclass strategy.
+These mapping strategies are enumerated by `jakarta.persistence.InheritanceType`,
+and the mapping strategy for an entity class hierarchy may be specified by
+the `Inheritance` annotation or `inheritance` element of an XML descriptor.
```
marko-bekhta: where in the hierarchy the `@Inheritance` is expected to be ?

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -2426,26 +2624,34 @@ Similarly, the following must be treated as equivalent:
...
-** Using annotations, a name is specified as a
-delimited identifier by enclosing the name within double quotes, whereby
-the inner quotes are escaped, e.g., `@Table(name="\"customer\"")`.
-** When using XML, a name is specified as
-a delimited identifier by use of double quotes, e.g., `<table name="\&quot;customer\&quot;"/>` footnote:[If
-<delimited-identifiers> is specified and individual annotations or XML
-elements or attributes use escaped double quotes, the double-quotes
-appear in the name of the database identifier.]
+This behavior relieves the developer of the need to explicitly specify
+a table or column name when the name of differs only in case from the
```
marko-bekhta:
```suggestion
a table or column name when the name differs only in case from the
```

spec/src/main/asciidoc/ch09-container-provider-contracts.adoc
```diff
@@ -248,58 +248,64 @@ classloader hierarchy at a given time.
...
+The following standard properties are defined for configuring the schema
+generation process.
+
+- In Jakarta EE environments these properties are passed by the container
+  in the `Map` argument to either `PersistenceProvider.generateSchema()`
+  or `createContainerEntityManagerFactory()`.
+- In Java SE environments, they are passed in the `Map` argument to either
+  the `Persistence.generateSchema()` or `createEntityManagerFactory()`.
+
+In Jakarta EE environments, any strings identifying file URLs of script
```
marko-bekhta, line 305:
```suggestion
In Jakarta EE environments, any strings identifying file URLs for/to script
```
?

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -1502,27 +1547,36 @@ public class ValuedCustomer extends Customer { ... }
...
+mapping strategy is used]. On the other hand, the discriminator value, if
+not defaulted, must be specified for every entity class in the hierarchy.
+
+The `DiscriminatorValue` annotation has only a `value` member, as listed
+below in <<a14585>>. A discriminator value must be consistent in type with
+the discriminator type (see <<a14530>>) of the discriminator column for the
+entity class hierarchy. The `value` member of `DiscriminatorValue` accepts
+a string which is interpreted as a literal value of the discriminator type.
+If the discriminator type is `INTEGER`, the specified `value` string must
+convertable to an integer value by calling `Integer.parseInt()`.
```
marko-bekhta:
```suggestion
be convertable to an integer value by calling `Integer.parseInt()`.
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -1749,29 +1805,30 @@ public EmploymentPeriod getEmploymentPeriod() { ... }
...
+The `EmbeddedId` annotation identifies a persistent field or property of
+an entity class or mapped superclass as holding a composite primary key
+represented as an embeddable class, as defined in <<a132>>. The embeddable
+class must be annotated as `Embeddable`.footnote:[The `Id` annotation is
+not applied to any field or property of the embeddable class.]
 
-There must be only one `EmbeddedId`
-annotation and no `Id` annotation when the `EmbeddedId` annotation is
-used.
+An embedded id class must not declare a field or property representing an
```
marko-bekhta:
```suggestion
An embedded id class must not declare a field or property representing a
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -1749,29 +1805,30 @@ public EmploymentPeriod getEmploymentPeriod() { ... }
...
+relationship, in the sense of <<a516>>, with another entity.
 
-The `AttributeOverride` annotation may be
-used to override the column mappings declared within the embeddable
-class.
+In a given entity class:
+
+- there must be at most one field or property annotated `EmbeddedId`, and
+- if there is a field or property annotated `EmbeddedId`, there must be
+  no field or property annotated `Id`.
```
marko-bekhta, line 1821: `annotated with` rather than just annotated ? no?

gavinking:
> well that was a quick read 🙂

Hahahah, thanks @marko-bekhta 🙏

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: In the rest of the spec, the names of APIs in the root back are not qualified. I think that's completely fine, though it definitely should be mentioned in _Document Conventions_ and currently it's not.

But changing "the" -> "such s" is fine.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: > just thinking that `categories` is too ambiguous

Sure. That change is perfectly fine by me.

> how does this play with all those other `a12345` anchors 

Well at some point somebody should go through and change them all. But not while we have multiple pull requests open :)

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Well yeah use of italics there is debatable. The thing is that it's the first time that the term occurs in the specification, and I *do* link to its definition, so I sorta am defining it.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: I think it works better to just remove the word "should".

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -112,38 +115,116 @@ Use of the generic variants of these collection types is strongly encouraged,
...
+well-defined object identity unique to the instance and is passed
+_by reference_, meaning that assignment of the instance to a field,
+local variable, or method parameter, followed by later access to
+the assigned value, preserves the identity of the instance.
+
+In general, operations performed by a persistence provider are not
+required to preserve the object identity of a value stored in an
+attribute of an entity. The provider is permitted to modify the
+value held in the attribute to a new value with a distinct object
+identity. Such modifications should never alter the _persistent_
```
gavinking: How about this:

```suggestion
identity. But such modifications never alter the _persistent_
```

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: > any specific expectations from the provider ?

I dunno. What does Hibernate do today? 

> throw a particular exception maybe ?

JPA doesn't define any exceptions for reporting problem with mappings.

spec/src/main/asciidoc/ch02-entities.adoc
marko-bekhta: yeah 👍🏻 🙂

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: This work?

```suggestion
`targetClass` member of the `ElementCollection` annotation. If an
element collection is declared with a raw type, the `targetClass`
member must be explicitly specified. Otherwise, if the element
collection is not declared with a raw type, the `targetClass` is
optional, and it must specify an entity type which is assignable 
to the type argument `E`.
```

spec/src/main/asciidoc/ch02-entities.adoc
marko-bekhta: >. What does Hibernate do today?

ah, I dunno 🙂 (haven't tested that scenario)

> JPA doesn't define any exceptions for reporting problem with mappings.

ah got it, so then no need to do it here 👍🏻

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: It's not completely necessary but if you want to, add a suggestion.

spec/src/main/asciidoc/ch02-entities.adoc
marko-bekhta: yes, thanks 🙏🏻 🙂

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: I guess I would like to gradually migrate to `@AnnotationName`.

I think fully-qualified names are only used for annotations not in the package `jakarta.persistence`.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Oh, italics for _emphasis_ are fine!

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Well remember I also need to stay a bit consistent with the surrounding text. So I didn't just change everything to my preferred style.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: How about this:

```suggestion
any XML descriptor, nor via any provider-specific mechanism, the 
following defaulting rules are applied in order:
```

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Awww gawd, you and IntelliJ trying to force by Australian-English-speaking-ass to conform to some claimed wrinkly of the American dialect!

gavinking:
```suggestion
An attribute representing one side of such an association must
```

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Naaah. "Inherit" is a good word.

gavinking: Also, "inherit" has "directly or indirectly" built in. "Extend" doesn't, precisely because it collides with the Java keyword.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Sure, make suggestion.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: I insist that annotations are adjectives. And so this is good how it is. I also find it reads more naturally.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: inherit (I don't mean "directly extend")

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1279,296 +1366,405 @@ Persistence for basic types is defined in <<a14205>> and <<a14719>>.
...
-in the database (e.g., by foreign key constraints).
+Every relationship (or association) between two entity types has
+well-defined _multiplicity_, either:
+
+- **one-to-one**,
+- **one-to-many** / **many-to-one** (depending on which side the relationship
+  is viewed from), or
+- **many-to-many**.
+
+An attribute representing one side of such an association must
```
gavinking:
```suggestion
Every attribute representing one side of such an association must
```

spec/src/main/asciidoc/ch09-container-provider-contracts.adoc
gavinking: I say "the address of a thing". Not "the address for a thing" or "address to a thing".

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
gavinking: No :)

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: So yeah. I read the "A portable application .... blah blah" bit as modifying the meaning of the previous "must". I think the spec does that in several places. It's not perfect, but I think it's clear what's going on?

And I'm not quite sure how else to say it. People hate to see the word "should" in specs.

spec/src/main/asciidoc/ch09-container-provider-contracts.adoc
marko-bekhta: ah ok 🙂 , I was thinking in terms that URL point to something 🙂

spec/src/main/asciidoc/ch02-entities.adoc
marko-bekhta: nice 🙂

spec/src/main/asciidoc/ch02-entities.adoc
marko-bekhta: maybe ..

> While persistence providers are allowed to accept other basic types as version, a portable application .......

yeah idk 🙂 🤷🏻

gavinking: I mean we can recite the whole "A persistence provider is required to support versions of ....".

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: I guess we're missing a link to a14891.

gavinking:
```suggestion
the `@Inheritance` annotation or `inheritance` element of an XML descriptor,
as defined by <<a14891>>.
```

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: @marko-bekhta Look at what I just pushed.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: @marko-bekhta See 51ae429 and eab5d4e.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: See fffef34.

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: See [2e42aff](https://github.com/jakartaee/persistence/commit/2e42affc226b73b298655f2d9b1e909c53b85e16).

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: See [1d4eaa5](https://github.com/jakartaee/persistence/commit/1d4eaa599ef809fae3645638db2e440e0bbe08fe).

spec/src/main/asciidoc/ch02-entities.adoc
gavinking: Answer:

```
Exception in thread "main" org.hibernate.MappingException: Recursive embeddable mapping detected for property 'foo.Book.emd.emb' of class 'foo.Emb'
```

which is nice.

But I think as far as the spec goes, this is covered.

gavinking:
Thank you @anija-anil. 

(And thanks again for @marko-bekhta.)

I have two reviews now, so I'm going to merge it.

PR #860, opened 2025-12-22
Add annotations which were missing from the spec

gavinking:
This is barely necessary, since Javadoc is sufficient specification, but for completeness we should mention them along with the other annotations.

### DIFF
--- a/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
+++ b/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
@@ -686,7 +686,7 @@ native SQL dialect of the database.
 |(Optional) Constraint name.
 |A provider-chosen name.
 
-|String[]
+|String
 |constraint
 |(Required) The native SQL expression to be checked.
 |


api/src/main/java/jakarta/persistence/query/ReadQueryOptions.java
```diff
@@ -34,7 +34,7 @@
  * <li>a method with a {@link StaticQuery} or Jakarta Data
  *     {@code jakarta.data.repository.Query} annotation
  *     whose {@code value} member specifies a {@code SELECT}
- *     statement,</li>
+ *     statement,
```
marko-bekhta, line 37: Entirely unrelated to the change 🙂 
But just in case... since we are now building with JDK 25 we could start using markdown for the javadoc 🫣 🙂

spec/src/main/asciidoc/ch10-metadata-annotations.adoc
```diff
@@ -852,4 +852,109 @@ annotation (or corresponding XML element).
...
 If there is more than one converter defined for the same target type,
 the `Convert` annotation must be used to explicitly specify which
-converter applies.
\ No newline at end of file
+converter applies.
+
+=== Annotations for Static Queries
+
+These annotations are used to declare queries associated with methods.
+In the context of Jakarta Data, they are used to declare repository
```
marko-bekhta: should we have a reference to JD in 
https://github.com/jakartaee/persistence/blob/6ee431ee9e1fbc5e36b33d80e86f15a5f54ad0e3/spec/src/main/asciidoc/related-documents.adoc?plain=1#L15 ?

api/src/main/java/jakarta/persistence/query/ReadQueryOptions.java
gavinking: Sure but we're not going to do that piecemeal. It's all or nothing.

spec/src/main/asciidoc/ch10-metadata-annotations.adoc
gavinking: Yes.

gavinking: Done.

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -655,6 +655,48 @@ be cached.
...
+[#a1224,options="header"]
+|===
+|Type |Name |Description |Default
+
+|String
+|name
+|(Optional) Constraint name.
+|A provider-chosen name.
+
+|String[]
```
njr-11:
```suggestion
|String
```

gavinking:
@njr-11 Thanks, fixed.

Issue #701, opened 2025-02-27
Configure EMF without XML in EE


hantsy:
Persistence 3.2 introduce programmatic APIs for SE env, but for Jakarta EE, we still require persistence.xml to configure Persistence.

Add an annotation similar to the `DataSourceDefinition`:

```java
@EntityManagerFactoryDefinition(
  name = "" // unique for inject by name
  datasource ="",  // or username/password/jdbcUrl
  txType ="LOCAL/JTA",
  qulififiers=[], // for inject by quilifier classes
  properties ={}
)
```

gavinking:
My knee-jerk reaction is that I hate it, just as I similarly hate `@DataSourceDefinition`, but I'm struggling to come up with a _strong_ argument against it.

It's true that we could introduce a `@PersistenceUnitConfiguration` annotation with essentially the same members as the `PersistenceConfiguration` class, but with the addition of the ability to select the CDI scope and qualifiers.

I'm not sure @lukasj would be very keen on this, however, because I don't think he likes seeing more stuff that's specific to the Jakarta EE environment leak into the Persistence spec. I guess if we provided a way to obtain a `PersistenceConfiguration` from a `@PersistenceUnitConfiguration`, then that might deal with that concern.

Dunno, I'm not in love with it at all. Hard to see how it's better than `persistence.xml`.

hantsy:
`@PersistenceUnitConfiguration` is a bit confusing due the existing `PersistenceConfiguration`. I am not sure, but for the current naming convention in the common annotations and other specifications, a `Definition`postfix is used everywhere.

gavinking:
> Ha, I hope all configurations support both annotations and programmatic APIs.

Let me prefix by saying I've been a champion of annotations since before just about anyone else. I mean, we designed EJB3, JPA, and CDI when annotations were still new, and copped a lot of flak at the time from e.g. the Spring community who were at the time incredibly addicted to XML.

But I guess I've never taken it as a requirement that *everything* should be settable in annotations. At least _some things_, for example, database connection properties, just seem to belong naturally in a `.properties` or `.xml` file, or whatever, and I've never really seen a need to move them into the Java code.

If we consider this annotation, the first thing I would ask is: where would you put it? And the natural answer would seem to be: on the module or perhaps package descriptor. It's not clear that that's a more convenient place to have the information than in a `persistence.xml` file.

[...]

Issue #237, opened 2019-12-05
Auto-Quote database identifiers to avoid use of SQL keywords


gergelymolnarpro:
Standardise hibernate.globally_quoted_identifiers, because when we create Java classes, we should not care about reserved SQL keywords.

Example: we can't create Release named class (or @Entity(name="Release") ) if we use MySQL database with JPA api.

andyjefferson:
+1 for auto-quoting database identifiers when necessary to avoid use of SQL keywords (*not* adding some Hibernate property), but then other persistence providers (e.g DataNucleus) have done that for years (before the JPA spec). 

Suggest that you change the title of this issue to something like "Auto-Quote database identifiers to avoid use of SQL keywords" ... the current title is way too generic and could mean many things to many people

gavinking:
The Hibernate team doesn't recommend the use of `hibernate.globally_quoted_identifiers` so I agree that JPA should not add anything like that.

beikov:
I guess what you are after is `hibernate.auto_quote_keyword`, but IMO this should be enabled by default in Hibernate. Not sure if the JPA spec should specify this behavior, but it's for sure a problem of Hibernate because it didn't enable this setting by default.

gavinking:
It's a hard call, and I'm not sure what the spec should day. 

It's pretty hard for the spec to demand that the implementations auto-quote every single keyword on every single database they support. (How would you even test for compliance with that demand?)

On the other hand, I agree that it's a desirable default behavior.

I wondered if there was some sort of reasonable compromise: that the implementations should quote all ANSI SQL keywords, or something like that, but that doesn't seem to work either, since ANSI defines many irrelevant keywords.

I'm not sure. Tricky one.

gavinking:
Alright, so the thing I forgot about is case-sensitivity. Quoting identifiers makes them case-sensitive, which can cause problems since, when the column name is inferred from a Java field name, JPA just doesn't know the case on the database.

Now, it might seem like, well, we should quote it *anyway*, since using a keyword as a table or column name is sure to fail *anyway*, so we can just at least do a best-effort.

But that's not actually true. Lots of databases will accept a keyword column or table name depending on how it occurs syntactically. E.g. you can easily have a column called `key` on (some? most?) databases.

So, for example, if I have a table with a column named `KEY` in uppercase, and a Java field `String key`, then that works today. But if we turn on auto-quoting, that working code actually will break!

So it turns out that this issue is a whole lot more subtle than it looks.

[...]

PR #861, opened 2025-12-22
@PersistenceModule

gavinking:
This change introduces:

- a way to declare a Java `module` as defining a persistence unit (or "persistence archive" in language we were using early on), that is, a package of entities and related code, along with XML mapping files, and 
- an annotation for specifying "static" configuration information which does not vary with the deployment, as an alternative to `persistence.xml`, and
- a way to segregate deployment-specific information into a separate `.properties` file.

See https://github.com/jakartaee/persistence/issues/701#issuecomment-3683334139.

_NOTE: This proposal is on ice until I can get feedback from someone else who is prepared to spend time thinking carefully about what a "persistence unit" or "persistence module" is going to mean in the future. Since I have not been able to gather such feedback yet, this feature is not currently planned for JPA 4._

### DIFF
--- a/api/src/main/java/jakarta/persistence/PersistenceModule.java
+++ b/api/src/main/java/jakarta/persistence/PersistenceModule.java
@@ -31,7 +31,8 @@ import static jakarta.persistence.ValidationMode.AUTO;
 
 /**
  * Declares that the annotated module defines a persistence unit
- * and specifies static configuration for the persistence unit.
+ * with the same name as the module itself and specifies static
+ * configuration for the persistence unit.
  * <p>For example, the following annotation declares a persistence
  * unit named {@code org.example.library} with resouce-local
  * transaction management:
@@ -53,18 +54,17 @@ import static jakarta.persistence.ValidationMode.AUTO;
  * <p>
  * Configuration more specific to a given deployment of the unit
  * may be provided in a standard Java properties file located at
- * {@code META-INF/org.example.persistenceModule.properties},
- * where {@code org.example.persistenceModule} is the name of the
- * unit, as specified by the {@link #name} attribute. Properties
- * settings specified in this file override setting specified by
- * this annotation.
+ * for example, {@code META-INF/org.example.library.properties},
+ * where {@code org.example.library} is the name of the unit
+ * (the module name). Properties settings specified in this file
+ * override setting specified by this annotation.
  * <p>
  * This annotation is an alternative to declaring the persistence
  * unit by means of the {@code <persistence-unit>} element of the
  * venerable {@code META-INF/persistence.xml} file.
  * <p>
  * An entity manager factory may be created for the persistence
- * unit by passing the {@linkplain #name} of the unit to
+ * unit by passing the unit/module name to
  * {@link Persistence#createEntityManagerFactory(String)}.
  * @see Persistence#createEntityManagerFactory(String)
  * @see Persistence#createEntityManagerFactory(String,java.util.Map)
@@ -74,11 +74,6 @@ import static jakarta.persistence.ValidationMode.AUTO;
 @Retention(RetentionPolicy.RUNTIME)
 @Target(ElementType.MODULE)
 public @interface PersistenceModule {
-    /**
-     * The name of the persistence unit.
-     * <p>Defaults to the module name.
-     */
-    String name() default "";
 
     /**
      * The transaction type of the persistence unit.
--- a/api/src/main/java/jakarta/persistence/PersistenceModule.java
+++ b/api/src/main/java/jakarta/persistence/PersistenceModule.java
@@ -37,6 +37,12 @@ import static jakarta.persistence.ValidationMode.AUTO;
  * unit named {@code org.example.library} with resouce-local
  * transaction management:
  * {@snippet :
+ * import jakarta.persistence.PersistenceModule;
+ * import jakarta.persistence.PersistenceProperty;
+ * import static jakarta.persistence.TransactionType.RESOURCE_LOCAL;
+ * import static jakarta.persistence.PersistenceConfiguration.*;
+ * import static jakarta.persistence.PersistenceConfiguration.SchemaManagementAction.VALIDATE;
+ *
  * @PersistenceModule(
  *     transactionType = RESOURCE_LOCAL,
  *     classes = {Book.class, Author.class, Publisher.class},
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/persistence/PersistenceModule.java
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * The schema management action to be performed against the database.
+     */
+    SchemaManagementAction schemaManagementDatabaseAction() default NONE;
+
+    /**
+     * The schema management action to be performed by generated DDL scripts.
+     */
+    SchemaManagementAction schemaManagementScriptsAction() default NONE;
```
gavinking, line 195: Not sure it really makes sense to have this here.

api/src/main/java/jakarta/persistence/PersistenceModule.java
```diff
@@ -0,0 +1,198 @@
...
+ * @PersistenceModule(
+ *     transactionType = RESOURCE_LOCAL,
+ *     classes = {Book.class, Author.class, Publisher.class},
+ *     excludeUnlistedClasses = true, // faster startup (no scanning)
+ *     nonJtaDatasource = "java:/comp/env/libraryData",
+ *     qualifiers = Library.class,
+ *     schemaManagementDatabaseAction = VALIDATE,
+ *     properties = {
+ *         @PersistenceProperty(name="jakarta.persistence.jdbc.batchSize", value="20"),
+ *         @PersistenceProperty(name="jakarta.persistence.jdbc.fetchSize", value="1000")
```
hantsy: Like the properties in other `@XXXXDefinition`, use a list of string that contain the property field/value is ok. 

```
properties = {"abc=xyz", ...}

```

Optionally, add a propertiesMap  as alternative. 

```java
propertiesMap = mapOf(...)
```

api/src/main/java/jakarta/persistence/PersistenceModule.java
```diff
@@ -0,0 +1,198 @@
...
+ *     }
+ * )
+ * module org.example.library { ... }
+ * }
```
hantsy, line 59: The new `PersistenceModule` depends on JPMS system?

api/src/main/java/jakarta/persistence/PersistenceModule.java
gavinking: @hantsy Well, there's various ways I could answer that question.

**First:** if we're inventing a minilanguage, why not go all the way and just say `properties` is a text block and you write:
```java
properties = """
             jakarta.persistence.jdbc.batchSize = 20
             jakarta.persistence.jdbc.fetchSize = 1000
             """;
```

That's even better, no?

**Second:** to me this question simply proves what I've been trying to say, which is that annotations are simply a pretty uncomfortable place to stick configuration properties! Even *you*, who were the one who wanted this feature, don't really feel comfortable with `@PersistenceProperty` (which has been there since JPA 1.0, from back when we didn't really know how to write Java).

**Third:** well, there's an actual good answer to your question and it's my fault for posting a bad code example. The answer is that you can write: 
```java
@PersistenceProperty(name=PersistenceConfiguration.JDBC_FETCH_SIZE, value="20")
```
I will update the code sample to reflect that usage.

gavinking: Anyway, please keep in mind that the real essence of this proposal is that your config properties are more likely to be specified in the `.properties` file, not in the Java code. The idea here is to decouple:

1. persistence unit declaration, from
2. config properties.

gavinking: Oh and:

> `propertiesMap = mapOf(...)`

You can't have a `Map` as an annotation member in Java.

api/src/main/java/jakarta/persistence/PersistenceModule.java
gavinking: According to this proposal, it depends on having a module descriptor, yes.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -78,6 +81,21 @@ of a Java SE application, it is necessary to explicitly list the managed
 persistence classes included in the persistence unit using the `class`
 element of the `persistence.xml` file. See <<a12305>>.
 
+==== Persistence Properties Files
```
OndroMih, line 84: A question: Where do these properties file reside? For example, I have a WAR with a JAR, which is a JPA module. Do I put the properties file into the JAR, into the WAR or keep it outside and specify it at deployment time?

OndroMih, line 84: Another question: To use the properties files, do I need to use the new PersistenceModule annotation? Or I can define a properties file also for PUs defined in persistence.xml?

Since the properties file name is tied with PU name, it should also make sense to use it with any PU, not only a PU in a persistence module, right? If yes, I think it makes sense to split this proposal about properties files into a separate PR.

OndroMih (commented):
I added a few questions to specific lines, @gavinking

spec/src/main/asciidoc/ch08-entity-packaging.adoc
gavinking: According to what is specified here, it goes in the `META-INF` directory of the JAR (i.e. of the persistence unit package).

Obviously I'm anticipating a future where these properties could also be specified in a location external to the persistence unit archive. But that's not in *this* proposal.

gavinking: > To use the properties files, do I need to use the new PersistenceModule annotation?

No, it says "a persistence unit", that is, *any* persistence unit.

> Since the properties file name is tied with PU name, it should also make sense to use it with any PU, not only a PU in a persistence module, right? 

Yes.

>  I think it makes sense to split this proposal about properties files into a separate PR.

As it is, it's not really especially useful without the annotation. If you're going to put properties in the `META-INF` directory, and you already have a `persistence.xml` there, you might as well just put them in `persistence.xml`. (Obviously it becomes more useful if they can also be sourced from somewhere else.)

gavinking:
Note that if we ever did this, the annotation should also allow specification of the info that is currently allowed in `<persistence-unit-defaults>`. See https://github.com/jakartaee/persistence/issues/237#issuecomment-4045345091.

api/src/main/java/jakarta/persistence/PersistenceModule.java
```diff
@@ -0,0 +1,204 @@
...
+    String[] jarFileUrls() default {};
+
+    /**
+     * A list of names of the mapping files the persistence
+     * provider must load to determine mappings for the entity
+     * classes. The mapping files must be in the standard XML
+     * mapping format, be uniquely named, and be resource-loadable
+     * from the application classpath.
+     */
+    String[] mappingFileNames() default {};
```
dmatej, line 145: Following https://github.com/jakartaee/persistence/issues/237#issuecomment-4045345091 - how we could add those defaults here? Would we follow with similar annotation for mapping files? Then, would make sense to separate those defaults somehow and use them directly here? Not sure about consequences ...

PR #862, opened 2025-12-25
more consistent treatment of properties + constants in Persistence

gavinking:
This pull request contains the following changes.

- The property `jakarta.persistence.schema-generation.create-database-schemas` is defined to be usable "in Java SE" as well is in Jakarta EE in the text of the spec (and that's useful) but it was not mentioned in the part of the spec which defines `persistence.xml` and so it was also missing from `PersistenceConfiguration`. I have added it where it was missing.
- The properties `database-product-name` and friends had already been promoted out of the schema generation section, but I forgot to add them to the section on `persistence.xml`. This is fixed here.
- The property `jakarta.persistence.bean.manager` was mentioned in one place in the text, but it was not in any of the lists of properties. I have added it to the list that includes the similar property `jakarta.persistence.validation.factory`. I have not added these to the list of `persistence.xml` properties, since apparently they're considered "integration-level", whatever that means.
- I have added some subsection headings to break this material up a bit better and make it easier to find stuff.
- I have proposed adding lists of configuration properties in inner interfaces of the `Persistence` class, as discussed in #863.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Persistence.java
+++ b/api/src/main/java/jakarta/persistence/Persistence.java
@@ -300,10 +300,10 @@ public final class Persistence {
     }
 
     /**
-     * Properties used to configure the interaction with JDBC.
+     * Properties used to connect to the database via JDBC in Java SE.
      * @since 4.0
      */
-    public interface JdbcProperties {
+    public interface ConnectionProperties {
         /**
          * Fully qualified name of the JDBC driver class.
          */
@@ -328,7 +328,13 @@ public final class Persistence {
          * An instance of {@code javax.sql.DataSource}.
          */
         String JDBC_DATASOURCE = "jakarta.persistence.dataSource";
+    }
 
+    /**
+     * Properties used to optimize the interaction with JDBC.
+     * @since 4.0
+     */
+    public interface JdbcProperties {
         /**
          * Override the default {@linkplain java.sql.Statement#setFetchSize JDBC fetch size}.
          * @since 4.0
@@ -342,20 +348,14 @@ public final class Persistence {
          * @since 4.0
          */
         String JDBC_BATCH_SIZE = "jakarta.persistence.jdbc.batchSize";
-    }
 
-    /**
-     * Properties specifying timeouts applying to interactions with the database.
-     * @since 4.0
-     */
-    public interface TimeoutProperties {
         /**
          * Default pessimistic lock timeout hint.
          */
         String LOCK_TIMEOUT = "jakarta.persistence.lock.timeout";
 
         /**
-         * Default query timeout hint.
+         * Default {@linkplain java.sql.Statement#setQueryTimeout query timeout} hint.
          */
         String QUERY_TIMEOUT = "jakarta.persistence.query.timeout";
     }
--- a/api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
+++ b/api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
@@ -74,7 +74,7 @@ import java.util.Objects;
  */
 public class PersistenceConfiguration
         implements Persistence.JdbcProperties,
-                   Persistence.TimeoutProperties,
+                   Persistence.ConnectionProperties,
                    Persistence.DatabaseProperties,
                    Persistence.SchemaManagementProperties,
                    Persistence.ValidationProperties,


gavinking:
@anija-anil would you given this a review, please?

api/src/main/java/jakarta/persistence/Persistence.java
MilaCodesIt, line 1: The use of interface for constant representation is never a good approach as the interface purpose is broken.

I would strongly suggest to change from interface to utility classes and actually add a new directory that can hold all the utility classes in the project

MilaCodesIt: Aside of this, using implements to read String properties is not the best approach, you can rely on core java properties to read the classes (Reflection) or in any other API that helps you read properties.

gavinking: > The use of interface for constant representation is never a good approach as the interface purpose is broken. I would strongly suggest to change from interface to utility classes

I'm not sure how using a utility class would be any different to an interface here. Both classes and interfaces define types, and so having a class as a bucket for sticking statics in is just as much of an "abuse" as using an interface for that purpose. And both interfaces and classes are widely used as containers for static things in modern Java, _because Java gives us no other option_.

>  and actually add a new directory that can hold all the utility classes in the project

It would certainly have been nice if we had defined the bootstrap APIs in a sub-package of `javax.persistence` all those years ago. But unfortunately we didn't. These properties are related to the `Persistence` and `PersistenceConfiguration` classes, and are used by clients of those classes, and so I would say it makes perfect sense to declare them within those classes.

> Aside of this, using implements to read String properties is not the best approach, you can rely on core java properties to read the classes (Reflection) or in any other API that helps you read properties.

I don't know what you mean by this.

MilaCodesIt: Yes.

Using interfaces is not the best approach as these config properties are static, there are better approaches to do so.

interfaces should be used for inheritance, and some other properties of programming paradigms.

using an interface will provide you with coupling and behavior that can be inherited... Static constants are not behavior. Thats the first point of using an utility class or an enum

njr-11 (commented):
I like the approach of having constants for all of the standard property names defined by the spec.  This will encourage applications not to hard-code them and is also a good reference from which anyone writing a persistence.xml can copy/paste the correct names.

In general, grouping the constants is also nice.  But I'm not sure it is worth it to remove & move the existing constants that were present in 3.2 in order to achieve that.  Applications would need to recompile to use the new constants. Or worse, applications that are trying to maintain shared source that works on both versions might switch to hard-code the names to maintain compatibility.  The JdbcProperties group seems like it is trying to be two things that are not the same: a group of properties by which the persistence provider can create a DataSource in Java SE/standalone usage vs. properties that are related to JDBC drivers such as fetch size and query timeout.  The latter have more general use in that they also apply to Jakarta EE scenarios with a DataSource in JNDI where fetch sizes, query timeout, lock timeouts, and so forth are also used.  If they are grouped, I would say it should be done differently.

Regarding the completeness and accuracy of all of the properties, I'll delegate the remainder of this review to Anija ( @anija-anil ) from our company who can cover all of that.

gavinking:
>  But I'm not sure it is worth it to remove & move the existing constants that were present in 3.2 in order to achieve that. Applications would need to recompile to use the new constants.

@njr-11 `PersistenceConfiguration` inherits the relevant properties. That's probably even *binary compatible* because I think constants get inlined by the compiler (might be wrong about that). But anyway it's certainly source-level compatible.

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -72,224 +72,13 @@
...
-         */
-        POPULATE
-    }
+public class PersistenceConfiguration
+        implements Persistence.JdbcProperties,
+                   Persistence.DatabaseProperties,
+                   Persistence.SchemaManagementProperties,
+                   Persistence.ValidationProperties,
+                   Persistence.CacheProperties,
+                   Persistence.BeanManagementProperties {
```
gavinking, line 82: @njr-11 👀

njr-11:
> > But I'm not sure it is worth it to remove & move the existing constants that were present in 3.2 in order to achieve that. Applications would need to recompile to use the new constants.
> 
> @njr-11 `PersistenceConfiguration` inherits the relevant properties. That's probably even _binary compatible_ because I think constants get inlined by the compiler (might be wrong about that). But anyway it's certainly source-level compatible.

That's good to see.  Somehow I missed that when reviewing.

gavinking:
>The JdbcProperties group seems like it is trying to be two things that are not the same: a group of properties by which the persistence provider can create a DataSource in Java SE/standalone usage vs. properties that are related to JDBC drivers such as fetch size and query timeout.

@njr-11 You mean that these two don't fit, is that right?

```java
        /**
         * Default pessimistic lock timeout hint.
         */
        String LOCK_TIMEOUT = "jakarta.persistence.lock.timeout";

        /**
         * Default query timeout hint.
         */
        String QUERY_TIMEOUT = "jakarta.persistence.query.timeout";
```

You got me &mdash; I tacked them on later because I didn't know where to put them.

gavinking:
> You mean that these two don't fit, is that right?

I added `TimeoutProperties` as a container for those two.

njr-11:
> > The JdbcProperties group seems like it is trying to be two things that are not the same: a group of properties by which the persistence provider can create a DataSource in Java SE/standalone usage vs. properties that are related to JDBC drivers such as fetch size and query timeout.
> 
> @njr-11 You mean that these two don't fit, is that right?
> 
> ```java
>         /**
>          * Default pessimistic lock timeout hint.
>          */
>         String LOCK_TIMEOUT = "jakarta.persistence.lock.timeout";
> 
>         /**
>          * Default query timeout hint.
>          */
>         String QUERY_TIMEOUT = "jakarta.persistence.query.timeout";
> ```
> 
> You got me — I tacked them on later because I didn't know where to put them.

Yes, those two, but also 

```java
        String JDBC_FETCH_SIZE = "jakarta.persistence.jdbc.fetchSize";
```

and 

```java
        String JDBC_BATCH_SIZE = "jakarta.persistence.jdbc.batchSize";
```

are dealing with how JDBC is _used_ to implement Jakarta Persistence.

I assume the other properties currently listed under `JdbcProperties` are there for obtaining a Connection in Java SE environments, either via a `DataSource` instance or possibly `DriverManager`.

One way to categorize them might be like the following, except that JavaSEConnectivityProperties is a lengthy, poor choice of a name,

UnitProperties - Jakarta EE only
JavaSEConnectivityProperties - Java SE only
JdbcProperties - batch size, fetch size, lock timeout, query timeout, which apply to both EE and SE

gavinking:
Hrm. I guess I think the most reasonable solution is a `BatchProperties` interface for just those two. I know the "fetch size" isn't usually considered a "batch size" but really it is, at least conceptually....

gavinking:
Well, OK, or, as you say `ConnectionProperties` and `JdbcProperties`.

gavinking:
There, I pushed your solution.

njr-11:
> There, I pushed your solution.

Thanks!  This will allow Jakarta EE users to completely ignore the connectivity properties group and Java SE users to completely ignore the unit properties group.

gavinking:
FTR the unit properties are actually defined specifically for the purpose of use in Java SE. See section 9.7 of the JPA 3.2 spec.

It's not very clear from the spec whether they can be passed to `createContainerEntityManagerFactory()`, but we (Hibernate) indeed allow it.

gavinking:
Any objections to just doing this? @sebersole?

I understand that @njr-11 is in favor of it now that his objections were addressed.

Anyone else?

gavinking:
I plan to merge this soon if there are no objections.

api/src/main/java/jakarta/persistence/Persistence.java
```diff
@@ -239,4 +239,320 @@ public boolean isLoaded(Object entity) {
...
+
+        /**
+         * Username for JDBC authentication.
+         */
+        String JDBC_USER = "jakarta.persistence.jdbc.user";
+
+        /**
+         * Password for JDBC authentication.
+         */
+        String JDBC_PASSWORD = "jakarta.persistence.jdbc.password";
```
njr-11, line 325: Reviewing these lists again, the only complication I can think of is if the user/password properties were ever allowed to be specified for a container managed persistence unit to indicate the user/password to connect to the data source with.  I don't know if Jakarta Persistence would ever allow for that usage pattern, but if it did intend to, then these to properties would belong in the common list rather than here.

sebersole: I get what you are saying, Hibernate allows this usage/interpretation.  But personally I still think looking at them as "connection properties" is completely valid.

gavinking: > Reviewing these lists again

🙏🙏

>  if the user/password properties were ever allowed to be specified for a container managed persistence unit to indicate the user/password to connect to the data source with

Do you mean that the persistence provider would call:

```java
public interface DataSource {
    Connection getConnection(String username, String password);
}
```

If so, that's a very interesting idea. 

I had to check the code, but indeed it looks like Hibernate does actually support exactly that, but only via the (deprecated) `hibernate.connection.username` and `hibernate.connection.password` properties, *not* via the standard Jakarta ones.

I think it's not common to use this, because it interferes with connection *pooling*.

And so I also don't think it would be a good idea to just change the semantics of these so that they suddenly started affecting the way the JPA provider obtains connection from the pool in Java EE.

sebersole:
Not sure how I feel about inner types for this, but overall +1 for grouping them.

gavinking:
> Not sure how I feel about inner types for this

@sebersole What would you do instead? I don't want to further crowd out the toplevel namespace. And, sure, we could put them in a subpackage, but they have a subtype in the root package.

api/src/main/java/jakarta/persistence/Persistence.java
njr-11: > I get what you are saying, Hibernate allows this usage/interpretation. But personally I still think looking at them as "connection properties" is completely valid.

Yes, a category named "connection properties" is appropriate for user/password properties.  However, a previous point that I brought up was to come up with a category that doesn't apply to usage with container managed persistence units, such that it would become clear to the user which properties to ignore.  In that regard, having user/password grouped with the Java SE-only properties does not make sense.  The category name should be adjusted and user/password moved to JDBC properties.  That is, if we decide that a container managed persistence unit should (either now or in the future) be able to use them.

> Do you mean that the persistence provider would call:
> 
> ```java
> public interface DataSource {
>     Connection getConnection(String username, String password);
> }
> ```

Yes, exactly.

> I think it's not common to use this, because it interferes with connection _pooling_.

I don't understand why that would be.  In our connection pool, the user name is part of the matching criteria and requests and it is easily distinguished which connections match and which do not.  I would expect any connection pool of JDBC connections to have that ability.

> And so I also don't think it would be a good idea to just change the semantics of these so that they suddenly started affecting the way the JPA provider obtains connection from the pool in Java EE.

This is an excellent point. Unless the current behavior requires an exception to be raised, preventing usage, when user/password are specified for container managed persistence units, we cannot switch the behavior without causing a breaking change in behavior.

Thinking about this some more, the Jakarta EE environment already provides an alternative solution when the DataSource bound in JNDI lacks a default user/password or when the user wishes to use a different one that is not the default: Instead of configuring the JNDI name of the DataSource, the user can configure the JNDI name of a resource reference to the DataSource where the resource reference specifies container managed authentication along with the user/password to use.

So I think we are fine here as long as user/password of the persistence unit are not currently being used for container managed.  I'll go ahead and approve.

njr-11 (approved):

api/src/main/java/jakarta/persistence/Persistence.java
gavinking: > I don't understand why that would be. In our connection pool, the user name is part of the matching criteria and requests and it is easily distinguished which connections match and which do not. I would expect any connection pool of JDBC connections to have that ability.

Yeah, well, I checked this, and all of the pooling JDBC drivers and all except one of the connection pool implementations I checked either simply throw from this method, or return an unpooled connection. So, sure, apparently there are *some* well-behaved pools, but it's definitely not universal.

gavinking: > So I think we are fine here as long as user/password of the persistence unit are not currently being used for container managed. I'll go ahead and approve.

@njr-11 if you think this is important, or you or @sebersole happen to know of users/customers who are using this, we can totally add some new properties for specifying the credentials to use when obtaining a connection from a `DataSource`. All I'm saying is let's not change the semantics of the ones that exist today.

sebersole: All I can tell you is that some users *used* to use this as we've had requests pertaining to it.  No idea how many, if any, still do.

That said, we can continue to just support this for our historical settings.  Its not important to me tbh.  Just commenting that they are still logically "connection properties".

gavinking:
Merging this. We can try it out in M1 and see how we like it and if we want to change anything.

Issue #864, opened 2025-12-27
declare DataSources in persistence.xml


gavinking:
Currently, and _incredibly_, the only Jakarta EE-standard way to declare a `DataSource` is using the `@DataSourceDefinition` annotation.

This annotation is not fit for purpose:

- It's declared `@Target(TYPE)` though it relates to no specific Java type
- It's compiled into the Java code, even though it specifies information that varies depending on the context in which the code is used (database credentials, server port and hostname, connection pool size, and so on)
- There's no well-defined way to override the information specified in the annotation with properties specified externally to the code

Therefore, in practice, applications get to choose from a zoo of completely vendor-specific approaches, e.g. WildFly `standalone.xml`, Quarkus `application.properties`, Hibernate `hibernate.properties`, and so on.

It certainly *can* make sense to separate the definition of datasources from the application program itself, but it's perhaps not completely necessary in all circumstances.

For cases where the `DataSource` definition doesn't need to be completely separate from the persistence unit archive, [I ask](https://github.com/jakartaee/persistence/issues/701#issuecomment-3693862175): _what could possibly be more natural than putting a datasource declaration in a file named `persistence.xml`?_

For example:

```xml
<persistence xmlns="https://jakarta.ee/xml/ns/persistence"
             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
             xsi:schemaLocation="https://jakarta.ee/xml/ns/persistence
                                 https://jakarta.ee/xml/ns/persistence/persistence_4_0.xsd"
             version="4.0">

    <persistence-unit name="Library" transaction-type="JTA">
        <description>My example persistence unit</description>
        <jta-data-source>jdbc/library</jta-data-source>
        <class>org.example.Book</class>
        <class>org.example.Author</class>
        <class>org.example.Publisher</class>
        <exclude-unlisted-classes/>
        <default-fetch-type>LAZY</default-fetch-type>
        <shared-cache-mode>ENABLE_SELECTIVE</shared-cache-mode>
    </persistence-unit>

    <jta-data-source>
        <name>jdbc/library</name>
        <description>My example persistence datasource</description>
        <url>jdbc:postgresql://localhost/library</url>
        <user>gavin</user>
        <password>p0ny</password>
        <max-pool-size>100</max-pool-size>
        <max-idle-time>10000</max-idle-time>
    </jta-data-source>
</persistence>
```

Or, in a slightly streamlined form:

```xml
[...]

PR #865, opened 2025-12-28
<data-source> in persistence.xml

gavinking:
This changeset realizes the idea proposed in #864, namely, that it should be possible to declare and configure a `DataSource` in the `persistence xml` file.

For example:

```xml
    <data-source name="jdbc/library">
        <description>My example persistence datasource</description>
        <class-name>org.postgresql.Driver</class-name>
        <url>jdbc:postgresql://localhost/library</url>
        <user>gavin</user>
        <password>p0ny</password>
        <isolation-level>TRANSACTION_READ_COMMITTED</isolation-level>
        <max-pool-size>100</max-pool-size>
        <max-idle-time>10000</max-idle-time>
    </data-source>
```

Or:

```xml
    <persistence-unit name="Library" transaction-type="JTA">
        <description>My example persistence unit</description>
        <data-source>
            <description>My example persistence datasource</description>
            <class-name>org.postgresql.Driver</class-name>
            <url>jdbc:postgresql://localhost/library</url>
            <user>gavin</user>
            <password>p0ny</password>
            <isolation-level>TRANSACTION_READ_COMMITTED</isolation-level>
            <max-pool-size>100</max-pool-size>
            <max-idle-time>10000</max-idle-time>
        </data-source>
        <class>org.example.Book</class>
        <class>org.example.Author</class>
        <class>org.example.Publisher</class>
        <exclude-unlisted-classes/>
        <shared-cache-mode>ENABLE_SELECTIVE</shared-cache-mode>
    </persistence-unit>
```


api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -617,4 +620,448 @@ public PersistenceConfiguration properties(Map<String,?> properties) {
...
+    public static class DataSourceConfiguration {
+
+        private String className;
+        private String description = "";
+        private String url = "";
+        private String user = "";
+        private String password = "";
+        private String databaseName = "";
+        private int portNumber = -1;
+        private String serverName = "localhost";
```
marko-bekhta: maybe:
```suggestion
        private String serverName = "";
```

my thoughts here are that since the precedence of `url` vs `servername+others` is vendor specific we can end up in a situation where the user is passing the `url` expecting it to be used, but if `servername+others` takes precedence they would be forced to do something like

```
new DataSourceConfiguration()
             .serverName(null)
             .url("jdbc:postgresql://localhost/documents")
```

marko-bekhta: and with that in mind ... maybe it's better to go with object types and have nulls instead of -1/0 ?

that way implementation will know for sure if the value was defined as such by the user or it is undefined.

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -617,4 +620,448 @@ public PersistenceConfiguration properties(Map<String,?> properties) {
...
+            return properties;
+        }
+
+        /**
+         * Specify the maximum time in seconds that this data source will
+         * wait while attempting to connect to a database. A value of 0
+         * specifies that the timeout is the default system timeout if
+         * there is one; otherwise, it specifies that there is no timeout.
+         * @param loginTimeout the login timeout in seconds
+         * @return this configuration
```
marko-bekhta: just checking if `0` is a standard or if we would be better off saying that the default is vendor specific, and let the user see the docs of their particular datasource type and figure it out?

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
```diff
@@ -617,4 +620,448 @@ public PersistenceConfiguration properties(Map<String,?> properties) {
...
+
+        /**
+         * Specify the number of seconds that a physical connection should
+         * remain unused in the connection pool before the connection is
+         * closed.
+         * A value of 0 indicates that there is no limit.
+         * @param maxIdleTime maximum idle time in seconds
+         * @return this configuration
+         */
+        public DataSourceConfiguration maxIdleTime(int maxIdleTime) {
```
marko-bekhta: btw speaking of various timeouts, should we have an overload that accepts `Duration` ?
I like when there's such option in the API since I don't need to go back to the javadoc to see in what units the timeout is set 🙂

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
```diff
@@ -413,4 +426,228 @@
...
+  <!-- **************************************************** -->
+
+  <xsd:complexType name="data-source-type">
+    <xsd:annotation>
+      <xsd:documentation>
+        Declares a container DataSource to be registered with JNDI. The
+        DataSource may be configured by setting the child elements for
+        commonly used DataSource properties. Additional standard and
+        vendor-specific properties may be specified using the properties
+        element.
```
marko-bekhta, line 450: what would happen if one passes any of the "predefined" properties through the properties element instead ?

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
```diff
@@ -413,4 +426,228 @@
...
+                   type="persistence:properties-type">
+        <xsd:annotation>
+          <xsd:documentation>
+
+            Used to specify vendor-specific properties and less commonly
+            used DataSource properties.
+
+            If a DataSource property is specified in the properties element
+            and the annotation element for the property is also specified,
+            the annotation element value takes precedence.
```
marko-bekhta, line 625: oh right, so we have some precedence expectations here, should we add the case from above for the predefined property ?

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
```diff
@@ -413,4 +426,228 @@
...
+          </xsd:documentation>
+        </xsd:annotation>
+      </xsd:element>
+      <xsd:element name="url" type="xsd:string" minOccurs="0">
+        <xsd:annotation>
+          <xsd:documentation>
+            A JDBC URL. If the url annotation element contains a DataSource
+            property that was also specified using the corresponding
+            annotation element, the precedence order is undefined and
+            implementation specific.
```
marko-bekhta, line 485: implementation vs vendor specific -- if I understand you correctly here, implementation is about the Jakarta Persistence implementation and vendor is about the database behind the datasource, so maybe it's worth to be more specific here so there's no possible confusion 🙂

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -106,20 +107,163 @@ The object/relational mapping information can take the form of:
...
 - specified by any combination of these means.
 
 See <<a12305>>.
 
 The root element of the `persistence.xml` file is the `persistence` element.
-The `persistence` element consists of one or more `persistence-unit` elements.
+It may contain:
+
+- any number of child `persistence-unit` elements, along with
+- any number of child `data-source` elements.
```
marko-bekhta, line 121: the way the `persistence` element is defined (a sequence) it would mean that it's any number of datasources followed by any number of persistence units (since sequence does not allow random order of elements)

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
```diff
@@ -413,4 +426,228 @@
...
+      <xsd:enumeration value="TRANSACTION_SERIALIZABLE"/>
+    </xsd:restriction>
+  </xsd:simpleType>
+
+  <!-- **************************************************** -->
+
+  <xsd:complexType name="data-source-type">
+    <xsd:annotation>
+      <xsd:documentation>
+        Declares a container DataSource to be registered with JNDI. The
```
marko-bekhta, line 446: I think this complex type is reused for top level and for the PU child element, and for the PU we later have

>  datasource may be declared as a child of the `persistence-unit` element, in
  which case it may or may not be bound to JNDI

soo should we be saying that `DataSource to be registered with JNDI` ?

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -106,20 +107,163 @@ The object/relational mapping information can take the form of:
...
+
+A `persistence-unit` element declares a named persistence unit.
+
+===== persistence-unit
 
 The `persistence-unit` element consists of the `name` and `transaction-type`
 attributes and the following sub-elements:
 `description`, `provider`,
-`jta-data-source`, `non-jta-data-source`,
+`jta-data-source`, `non-jta-data-source`, data-source
```
marko-bekhta:
```suggestion
`jta-data-source`, `non-jta-data-source`, `data-source`
```

or even 


```suggestion
`data-source`, `jta-data-source`, `non-jta-data-source`
```

(since it's a sequence and the order would matter..)

btw I noticed that there seem to be other elements in the xsd but not listed here : `qualifier`, `scope`

marko-bekhta (commented):
added a few questions inline 🙂

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
gavinking: I dunno. I was directly mirroring `@DataSourceDefinition`, so that there are no new semantics to define.

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
gavinking: The exact same thing that happens as if you used `@DataSrouceDefinition`.

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
gavinking: As I said: I am *reallly* trying hard not to diverge from existing semantics of things that already exist.

I am aware that the existing semantics probably kinda blow. But it would be worse to deviate from them.

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
gavinking: > if I understand you correctly here

It's not me you're trying to understand. I copy/pasted all that text verbatim.

api/src/main/java/jakarta/persistence/PersistenceConfiguration.java
gavinking: I guess that would be OK, but we would have to also do it for `loginTimeout`. So that would be two extra methods.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
gavinking: Yeah actually I dislike how everything in `persistence.xml` has a defined order, even though no such natural order exists. The order is not semantically meaningful nor relevant.

gavinking:
For reasons [explained in the issue](https://github.com/jakartaee/persistence/issues/864#issuecomment-3727504936),  I have removed the proposal for a `DataSourceConfiguration` object from this proposal.

gavinking:
@scottmarlow @starksm64 @bstansberry I know we have already discussed this via email, but is there anything you would like to add here, or should I just move forward with this?

@njr-11 Anything to say from the Liberty side?

njr-11:
> @njr-11 Anything to say from the Liberty side?

I'm not in favor of duplicating `<data-source>` into persistence.xml when we already have it in web.xml and application.xml.  I don't believe it adds value on the Jakarta EE app server side, and it introduces complexity there with how this merges or overrides definitions (and annotations) of the same name that are found in the existing locations.  I can also see it being a likely outcome that Jakarta EE will de-emphasize/deprecate `<data-source>` in favor of something better once we have Jakarta Config.

gavinking:
> I'm not in favor of duplicating `<data-source>` into `persistence.xml` when we already have it in `web.xml` and `application.xml`. I don't believe it adds value on the Jakarta EE app server side

If we consider only Jakarta EE in isolation then I agree.

The problem is that `web.xml` and `application.xml` aren't available in the Java SE standalone case, and that creates various kinds of friction for testing and other scenarios where people are using the same persistence unit both in EE and SE.

>  I can also see it being a likely outcome that Jakarta EE will de-emphasize/deprecate `<data-source>` in favor of something better once we have Jakarta Config.

That's entirely possible, but there's no guarantee that Jakarta Config&mdash;if and when it arrives&mdash;will actually address Java SE programs or be something that people will embrace for usage outside of the EE platform. (And it also implies an additional dependency.)

Issue #866, opened 2025-12-30
TypedQueryReference#getOptions


sebersole:
`TypedQueryReference#getOptions` returns `List<FindOption>`.  It is not inconceivable that "irrelevant" `FindOption` types may be contained there.  I think the Javadoc ought to mention that such options are simply ignored.

Or better yet, consider adding a `QueryOption` corollary.

gavinking:
> Or better yet, consider adding a `QueryOption`

Yeah so I've been wondering about that. It makes sense and it's definitely conceptually cleaner.

The only thing that gave me pause was: `Query` has all these special-purpose methods like `setLockMode()`, `setFlushMode()`, `setCacheRetrieveMode()`, `setCacheStoreMode`, `setTimeout()`, *some of which we only just added in 3.2*, which would all be made redundant by a `setOption()` method on `Query` :-(

(Of course, that's not a very good argument against doing this. It just makes me feel dumb.)

gavinking:
Trying to think about things which might be `QueryOption`s but not `FindOption`s:

- `QueryFlushMode`, as proposed by #787, if we ultimately decide to do that
- some kind of Jakarta Data-style `Page` object as an alternative to `setFirstResult()`/`setMaxResult()`

There are also a couple of vendor-specific options that make sense, from Hibernate, since that's what I know: 
 
- an option to control query result set caching
- an option to add a comment to the generated SQL

So, there's some stuff there I guess.

sebersole:
> an option to control query result set caching

This is definitely one I was hitting my head against implementing some of this.  

Another (though this is largely amorphic atm) is the difference, if any, between a query timeout and a lock timeout applied to the query.

gavinking:
> the difference, if any, between a query timeout and a lock timeout applied to the query

Ah. Yeah. Legacywise, we have:

- `jakarta.persistence.lock.timeout`
- `jakarta.persistence.query.timeout`

I'm not really sure they're meaningfully different.

I guess in Hibernate we map the first one to a `WAIT` clause in the query we second one we always mapped to `Statement.setQueryTimeout()`. So they're in some sense "different" in our implementation, but I'm not sure I know how they're really different in a cosmic sense.

gavinking:
[...]

PR #867, opened 2025-12-30
introduce QueryOption

gavinking:
See #866.

The value of this over just having `FindOption` is a bit marginal. But it's conceptually cleaner, and we might very well appreciate having it in the future.


gavinking:
We took the other path.

PR #868, opened 2025-12-30
clean up redundant overloads of find() and get() on EntityHandler

sebersole:
Commit message says it all, but -

* Moved legacy find(Class,Object,LockModeType) back from EntityHandler to EntityManager[1]
* Dropped get(Class,Object,LockModeType) from EntityHandler
* Added Javadoc @apiNote clarifying that get is the same as find except throwing an exception rather than returning null 

[1] We might also consider simply removing it as it is now redundant with the FindOption form.  For now, I just marked it deprecated.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -171,7 +171,7 @@ import jakarta.persistence.criteria.CriteriaQuery;
  * {@link LockModeType}. Some operations of this interface,
  * including the methods {@link #lock(Object, LockModeType)},
  * {@link #refresh(Object, LockModeType)}, and
- * {@link #find(Class, Object, LockModeType)}, accept an explicit
+ * {@link #find(Class, Object, FindOption...)}, accept an explicit
  * {@link LockModeType}, allowing the client to request a specific
  * type of lock.
  *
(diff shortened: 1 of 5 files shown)


gavinking:
> We might also consider simply removing

I would just remove it.

A strong reason to just remove is that if I write `entityAgent.find(Book.class, isbn, LockMode.PESSIMISTIC_READ)`, I will get a deprecation warning, and even if I tried to write `entityAgent.find(Book.class, isbn, (FindOption) LockMode.PESSIMISTIC_READ)` to squash it, I would get a different warning about a useless cast.

sebersole:
As you prefer :)  Gone.

gavinking:
@sebersole Can you fix `EntityManagerImpl` in the TCK, the build is failing.

sebersole:
> This is backward compatible

It is not though as a method has been removed.

gavinking:
> > This is backward compatible
> 
> It is not though as a method has been removed.

Any source code which was calling the old removed method will continue to compile, and have the same semantics at runtime, even though it will now be calling a different method.

It's not *binary* compatible, but it's source compatible.

sebersole:
I am beyond struggling to get all of this loaded into my IDE properly (Maven ftw).  I "blindly" made the changes I think were needed to `tck/` stuff, but :shrug:

gavinking:
LGTM.

gavinking:
@njr-11 any objection?

njr-11:
> It's not _binary_ compatible, but it's source compatible.

Did already we break binary compatibility anywhere else for Persistence 4.0 ?  If not, it would be unfortunate to do so for this method, where the usage remains valid.

Although I understand why you want to remove it.  If the deprecated method isn't removed, then a user would need to avoid the warning on the new method with something awkward like,

```
entityAgent.find(Book.class, isbn, new FindOption[] { LockMode.PESSIMISTIC_READ })
```

gavinking:
> Did already we break binary compatibility anywhere else for Persistence 4.0

Ummmmmmm that is a great question to which I do not know the answer.

gavinking:
Well, actually I do know the answer, we _did_, because:

> Deleted APIs marked deprecated for removal

Not sure if you think that counts.

gavinking:
Anyway, I don't view binary compatibility as a goal here, at least not for code which interacts with the `EntityHandler` APIs. I'm pretty sure that everyone is going to expect to recompile their programs when they upgrade to JPA4. (Entity classes naturally stay binary compatible.)

gavinking:
Going to go ahead and merge this.

PR #869, opened 2025-12-31
Remove declaration of EntityHandler as sealed

sebersole:
Remove declaration of `EntityHandler` as sealed.  As defined, providers cannot properly define a common implementation.

For example, in Hibernate we have a contract named `SharedSessionContract` which is the base of both our `Session` (`EntityManager`) and `StatelessSession` (`EntityAgent`).  Naturally, this thing would implement `EntityHandler` and define common implementations of various `EntityHandler` methods.  However, that is not possible currently because `EntityHandler` is sealed and only permits `EntityMaager` and `EntityAgent` as implementors.

This PR removes the sealing of `EntityHandler`.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityAgent.java
+++ b/api/src/main/java/jakarta/persistence/EntityAgent.java
@@ -56,7 +56,7 @@ import java.util.List;
  *
  * @see EntityManager
  */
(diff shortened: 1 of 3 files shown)


gavinking (approved):
Makes me sad but oh well 😢

PR #871, opened 2025-12-31
Remove use of FindOption in TypedQueryReference

### LINKED ISSUE #866: TypedQueryReference#getOptions

gavinking:
Fixes #866.

I think I like this solution better than #867.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Timeout.java
+++ b/api/src/main/java/jakarta/persistence/Timeout.java
@@ -65,4 +65,20 @@ public class Timeout implements FindOption, RefreshOption, LockOption {
     public int milliseconds() {
         return milliseconds;
     }
+
+    @Override
+    public String toString() {
+        return "Timeout[" + milliseconds + " ms]";
+    }
+
+    @Override
+    public boolean equals(Object other) {
+        return other instanceof Timeout that
+            && this.milliseconds == that.milliseconds;
+    }
+
+    @Override
+    public int hashCode() {
+        return milliseconds;
+    }
 }


gavinking:
@sebersole 👀

sebersole:
A little torn tbh, mainly because I despise "hints" :)  Thinking about Hibernate's "cacheable" or any of the other provider-specific things here.

But given that I don't love FindOption here either, I'm fine with it.

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -415,4 +415,12 @@ TypedQuery<X> setParameter(int position, Date value,
...
      */
     TypedQuery<X> setTimeout(Integer timeout);
+
+    /**
+     * Set the query timeout. This is a hint.
+     * @param timeout the timeout, or null to indicate no timeout
+     * @return the same query instance
+     * @since 4.0
+     */
+    TypedQuery<X> setTimeout(Timeout timeout);
```
sebersole, line 425: +1000

gavinking:
> A little torn tbh, mainly because I despise "hints" :) 

Yeah same. But this doesn't preclude us introducing a `QueryOption` at some future point if we have a better justification. That possibility doesn't really much affect this design here, because the whole `XxxOption` approach doesn't work with annotations (and this stuff is all about the annotations).

Anyway, it looks like we have a consensus here, since you guys both approved.

I'm going to merge this so Steve can continue with his implementation efforts.

PR #872, opened 2025-12-31
Invert query hierarchy

gavinking:
Neither #842 nor #858 has exactly set the world on fire, so, out of desperation, I decided to go where I told myself I wasn't going to go. This is by far the simplest solution to our problem, but it breaks perfectly legit code.

This change inverts the inheritance relationship between `Query` and `TypedQuery`. This has the following advantages:

1. `executeUpdate()` is gone from `TypedQuery`, which inherited it inappropriately.
2. I can safely mark the raw-type returning methods as deprecated without any risk of impacting `TypedQuery` and its clients.
3. I can add the beautiful new `ofType()` method to `Query` without it spilling over onto `TypedQuery`. ❤️
4. The design doesn't force me to introduce additional constructs; but, if we wanted to, at some point in the future, we could in principle introduce: `interface Statement extends TypedQuery<Void>` or whatever, so I don't think we're closing off any options there.

Of course, the change also has the massive downside that a `TypedQuery` is no longer assignable to `Query`, which is most certainly, without a shadow of a doubt, going to cause breakage for some people. 

- I don't believe there are any idioms *typical* of JPA code which involve assigning `TypedQuery` to `Query`, but certainly it's the sort of change that will impact frameworks and generic code. 
- I noticed that it broke a few bits of code in the TCK. Now, that code in the TCK is particularly bad code, that just randomly assigned a `TypedQuery` to `Query` for no good reason at all, and I sure *hope* that's not what typical JPA code looks like, but still. I guess I don't much mind breaking code like that, which has been setting there with compiler warnings for *years*.
- It's going to impact (i.e. break) Hibernate's `Query` hierarchy, but fortunately I believe that the impact it has is on methods we actually already deprecated in H6 precisely because of the problems with the previous design in JPA. If we think this approach is worth serious consideration, we will have to look a bit more closely at the impact on Hibernate and Hibernate users, but I think it might be OK.

This approach actually works out elegantly. The question is: how angry are users going to be about the breakage it causes?

### DIFF
--- a/tck/common/src/main/java/ee/jakarta/tck/persistence/common/pluggability/altprovider/implementation/EntityManagerImpl.java
+++ b/tck/common/src/main/java/ee/jakarta/tck/persistence/common/pluggability/altprovider/implementation/EntityManagerImpl.java
@@ -92,7 +92,7 @@ public class EntityManagerImpl implements jakarta.persistence.EntityManager {
 
 	@Override
 	public <T> TypedQuery<T> createNamedQuery(String arg0, Class<T> arg1) {
-		QueryImpl query = new QueryImpl();
+		TypedQueryImpl query = new TypedQueryImpl();
 		query.name = arg0;
 		query.queryOnClass = arg1;
 		return (TypedQuery<T>) query;
@@ -112,7 +112,7 @@ public class EntityManagerImpl implements jakarta.persistence.EntityManager {
 
 	@Override
     public <T> TypedQuery<T> createNativeQuery(String arg0, Class<T> arg1) {
-		QueryImpl query = new QueryImpl();
+		TypedQueryImpl query = new TypedQueryImpl();
 		query.nativeSQL = arg0;
 		query.queryOnClass = arg1;
 		return query;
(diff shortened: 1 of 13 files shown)


beikov:
This doesn't look that bad and I think I like it. Breakage will be unavoidable anyway, but this seems like the most source compatible for reasonable source code :)

sebersole:
My $0.02... if we are going to break it, just break it properly and define SelectionQuery and MutationQuery

gavinking:
> My $0.02... if we are going to break it, just break it properly and define SelectionQuery and MutationQuery

Am I right in interpreting that as a vote for https://github.com/jakartaee/persistence/pull/842?

gavinking:
Note that #842, unlike this option, doesn't break things. (Though it does deprecate a lot of things.)

sebersole:
> Note that #842, unlike this option, doesn't break things. (Though it does deprecate a lot of things.)

Yet ;)

gavinking:
> Yet ;)

Yeah. I mean, clearly all those `@Deprecated(since = "4.0", forRemoval = true)` tags are pointing in a certain direction.

gavinking:
>  if we are going to break it, just break it properly 

I think you're right though. If we're going to do *anything* here, given that there's no option with zero impact on existing code, we should do the "most correct" thing. And I guess that's #842 or something very very like it.

gavinking:
After several backchannel discussions with various people, the sense I took away was: "if we're going to do this, do #842".

I'm therefore closing this PR.

gavinking:
[FTR, the huge black mark against this option was the breakage involved.]

PR #883, opened 2026-01-07
Add TCK tests for EntityManagerFactory#addListener

sebersole:
To be clear, I have no idea how to run these tests so this PR may be wonky.  

Also, I tried to stick to the wonderful existing test infrastructure as best I understood :)

### DIFF
--- a/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/callback/added/Client.java
+++ b/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/callback/added/Client.java
@@ -1,0 +1,361 @@
+/*
+ * Copyright (c) 2007, 2023 Oracle and/or its affiliates. All rights reserved.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0, which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * This Source Code may also be made available under the following Secondary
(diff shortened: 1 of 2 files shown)


gavinking:
Thanks Steve. We will have to sit down soon and figure out how to run the TCK against your branch. But I am snowed under right now with exams and stuff.

tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/callback/added/Client.java
```diff
@@ -0,0 +1,361 @@
...
+
+	public Client() {
+		super();
+	}
+
+	public JavaArchive createDeployment() throws Exception {
+		String pkgNameWithoutSuffix = Client.class.getPackageName();
+		String pkgName = pkgNameWithoutSuffix + ".";
+		String[] classes = { pkgName + "LineItem", pkgName + "Order", pkgName + "Product" };
+		String[] xmlFiles = { ORM_XML };
```
marko-bekhta:
```suggestion
		String[] classes = { Product.class.getName() };
		String[] xmlFiles = { };
```

hey @sebersole 👋🏻 🙂 
I've tried running this test with Hibernate ORM and noticed that we have a few extra entity classes listed that aren't in the package + looks like we don't need to provide extra info through `orm.xml`. With these updates ^ it seems to be passing.

marko-bekhta (approved):

PR #884, opened 2026-01-08
fix the definition of SharedCacheMode

gavinking:
make clear that it determines whether and entity *might* be cached


gavinking:
Nathan has approved, so I'm going to just merge this straightforward textual improvement.

Thanks @njr-11.

Issue #888, opened 2026-01-08
Clarify behavior of EMF#getResultSetMappings when passed Object[] or Tuple


sebersole:
Consider a simple example:

```java
@SqlResultSetMapping( name = "id",
        columns = @ColumnResult( name="id", type = Integer.class )
)
```

What should happen with these 2 calls given that mapping...

```java
factory.getResultSetMappings( Object[].class );
factory.getResultSetMappings( Tuple.class );
```

I can justify a number of interpretations, but I think it is something we should be explicit about since I think, to a degree, these are special cases.

gavinking:
Interesting. What do *you* think the semantics should be?

A couple of thoughts:

- I believe I'm supposing that the right behavior here is inferable from section 3.14.13 (Result Classes for SQL Queries) of the latest spec draft.
- If I'm not mistaken, according to the legacy spec, `Tuple` is simply not supported as a result class anywhere except for `createTupleQuery()` for criteria. Obviously we (Hibernate) support it in other contexts. And now that we have added `TupleMapping`, we do need to support it for native queries too.

So there is work to do here.

sebersole:
> I believe I'm supposing that the right behavior here is inferable from section 3.14.13 (Result Classes for SQL Queries) of the latest spec draft.

That's fair - I had not looked at the spec doc.  I'll followup there to see if it clarifies.

> If I'm not mistaken, according to the legacy spec, Tuple is simply not supported as a result class anywhere except for createTupleQuery() for criteria. Obviously we (Hibernate) support it in other contexts. And now that we have added TupleMapping, we do need to support it for native queries too.

Regardless of what the previous versions of the spec did or did not support, as you point out there is now a `TupleMapping` which implies it should be possible from this method to me.

To me the point about `Object[]` and `Tuple` are really the same - like `Object` they *could* really apply to any ResultSetMapping.  If we want to treat them differently, that's fine - but we should be clear about that.  See below..

> Interesting. What do you think the semantics should be?

I think a natural conclusion would be that `Object[]` applies to any ResultSetMapping with multiple results and would therefore skip mappings (like my simple example) which define just one result.  

[...]

Issue #887, opened 2026-01-08
Remove declaration of ResultSetMapping as sealed


sebersole:
Hibernate, and I am assuming other providers, have historical ways to define more rich ResultSet mappings.

Because ResultSetMapping was introduced as sealed and all of its impls as records, it is impossible to model those richer models using this API; which is a shame imo. Here is a PR to remove the seal and allow such usage.

A further step would be to make the impls NOT records. This would allow even better implementation of this idea by implementors. I held off on that for the time being until we had consensus on this idea in general.

PR #889, opened 2026-01-08
tuples and aliases

gavinking:
A rough sketch of a possible way to deal with #888.

cc @sebersole

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
+++ b/api/src/main/java/jakarta/persistence/EntityManagerFactory.java
@@ -531,11 +531,11 @@ public interface EntityManagerFactory extends AutoCloseable {
      * A map keyed by {@linkplain SqlResultSetMapping#name name}, containing
      * a typed {@link ResultSetMapping} for every result set mapping defined
      * {@linkplain SqlResultSetMapping in annotations} whose inferred result
-     * type is assignable to the given Java type.
+     * type is exactly the given Java type.
      * @param resultType any Java type, including {@code Object.class}
      *                   meaning all result set mappings
      * @return a map keyed by query name
-     * @param <R> the specified upper bound on the query result types
+     * @param <R> the specified query result type
      *
      * @since 4.0
      */
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -3930,8 +3930,8 @@ of `EntityManager` for execution.
 An instance of `ResultSetMapping` representing a result set mapping
 declared using the `@SqlResultSetMapping` annotation may be obtained
 from the `EntityManagerFactory` by calling `getResultSetMappings()`,
-passing a Java class to which the query results are assignable
-according to the rules specified above in <<sql-result-set-mapping>>.
+passing the Java class if the query results as determined by the
+rules specified above in <<sql-result-set-mapping>>.
 
 ==== Result Classes for SQL Queries [[result-class-native-sql]]
 


sebersole:
For certain Hibernate features, it would be nice to define a proper alias across all of the `TupleElement` impls.

E.g. `ColumnMapping` currently just uses the column-name as the alias.

Consider a Hibernate feature leveraging alias with with constructor result mappings - we allow users to map these in a way such that the aliases can be used to define bean property names for use in cases where the constructed class does not have an appropriate constructor.  Roughly -

```java
class DropDownItem {
    private Integer key;
    private String text;

    // getters/setters...
}

var key = column("key", "id", Integer.class);
var text = column("text", "name", String.class);
var dto = constructor(DropDownItem.class, key, text);
```

Of course, this might be kind of moot depending on the decision for https://github.com/jakartaee/persistence/issues/887

gavinking:
@sebersole Ok, fine, I can pull `withAlias()` up. Gimme 3 mins.

gavinking:
@sebersole There, done, is that what you mean?

api/src/main/java/jakarta/persistence/sql/MappingElement.java
```diff
@@ -30,4 +30,9 @@
  */
 public sealed interface MappingElement<T> extends TupleElement<T>
         permits EntityMapping, ColumnMapping, ConstructorMapping  {
+    /**
+     * Specify an alias for this element of the result set.
+     * @param alias The alias
+     */
+    MappingElement<T> withAlias(String alias);
```
gavinking, line 37: 👀

sebersole: Functionally yes, this is what I mean.   

Practically, not sure I'd require the second instantiation here but I can understand wanting to avoid an explosion of factory methods on `ResultSetMapping` and that you really want to keep this sealed with record impls.

gavinking:
Mmmmm @sebersole should we be targeting this one for M1?

sebersole:
Its not critical for M1 imo.

gavinking:
> Its not critical for M1 imo.

OK then, not for M1.

gavinking:
OK, I believe I have now cleared up the ambiguity raised in #888, along with another ambiguity in the spec with respect to what types are allowed for `@ColumnResult`.

I've also added support for explicit aliases in programmatic result mappings, which AFAIU we all agree is useful.

Therefore I think this one is ready to go.

gavinking:
> Therefore I think this one is ready to go.

Well, ups, not anymore; looks like I need to resolve conflicts.

gavinking:
Rebased as #966 to deal with conflicts.

api/src/main/java/jakarta/persistence/sql/ColumnMapping.java
```diff
@@ -68,6 +77,6 @@ public static ColumnMapping<Object> of(String columnName) {
      * @param <T> The type of the resulting scalar value
      */
     public static <T> ColumnMapping<T> of(String columnName, Class<T> type) {
-        return new ColumnMapping<>(columnName, type);
+        return new ColumnMapping<>(columnName, type, columnName);
     }
 }
```
sebersole, line 82: I assume your expectation here is to just directly construct the `ColumnMapping` to apply an alias as opposed to having a factory method?

gavinking: Yeah I would use:

        column("name").withAlias("alias")

Or:

    column("name", int.class).withAlias("alias")

gavinking: If you prefer, we could rename `withAlias` to `as`.

sebersole: I like withAlias better here.

PR #892, opened 2026-01-12
remove redundant refresh(Object entity, LockModeType lockMode)

gavinking:
I just realized that since we're doing #868, we should for consistency also kill of a redundant overload of `refresh()`.


gavinking:
cc @sebersole

gavinking:
I am going to just merge this, since it's just a followup on #868, which is sorta incomplete without this.

PR #893, opened 2026-01-12
Combine API/TCK/Spec under the same parent pom / Initial release Jenkinsfile

marko-bekhta:
I've started with release job per "module" but that was too cumbersome, so I ended up grouping all the modules under one parent (that should also simplify the project import 🙂 🤞🏻)

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -23,7 +23,9 @@ jobs:
 
     strategy:
       matrix:
-        java_version: [ 21, 25 ]
+        java-version:
+          - { java_version_numeric: 21, extra_args: "-Denforcer.skip" }
+          - { java_version_numeric: 25, extra_args: "" }
 
     steps:
     - name: Checkout for build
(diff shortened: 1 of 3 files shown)


gavinking:
Excellent! Let me know when this is ready to merge. 🙏🙏🙏

marko-bekhta:
ok 🙂 
I thiiiiiiiiiiiiink this should do it ... 

For me the project in Intelij was getting messed up once you reopened with the parent pom, and the way I got it to a normal import was to : 
1. close it 
2. remove from the recent project list 
3. remove all `targets` and `.idea` files (I went with `find . -name "target" -type d -exec rm -rf {} +` from a project root) 
4. open as a new project from the parent pom, 
5. run repair IDE (from IDE actions) to get the project fully re-scanned and correctly imported. 

For the release job, we'll need to create a new one on https://ci.eclipse.org/jpa/ but there's no rush, we can do it any time before the release.

Additional things updated in the PR:
- build requires JDK 25 (so we would get consistent javadoc rendering and just consistent builds 😃)
- I tried to adjust the file license headers to let the build plugin run over all the modules, but overall the license headers a bit of a mess since API/spec and TCK have different ones

if you'll have the time we can try to set it up some time this or next week 🤞🏻

gavinking:
Thanks @marko-bekhta!

Some comments:

0. I did, as you say, have to close/nuke/reopen the project, but it worked.
1. I keep getting the following messages:

    ```
    [WARNING] The requested profile "postgresql" could not be activated because it does not exist.
    [WARNING] The requested profile "hibernate" could not be activated because it does not exist.
    ```
    it's weird, because Is still see those profiles existing in IntelliJ.
2. The spec is now built with the name `jakarta-persistence-spec-4.0.pdf` instead of `jakarta-persistence-spec-4.0-SNAPSHOT.pdf` which I think it had before. (This isn't a huge problem, but it's not perfect.)

marko-bekhta:
> I keep getting the following messages:

hmm, I've tried running things from the terminal and haven't noticed it in the log... what command were you trying to execute ? (I was running something like `mvn -pl :persistence-tck-spec-tests test -P "hibernate,postgresql" .......`

 
> The spec is now built with the name jakarta-persistence-spec-4.0.pdf instead of jakarta-persistence-spec-4.0-SNAPSHOT.pdf which I think it had before. (This isn't a huge problem, but it's not perfect.)

yeah ... it was there originally but I wasn't sure if it would make much difference if we drop it 🫣, this pr: 

* https://github.com/jakartaee/persistence/pull/899

should get it back to the original format.

gavinking:
> hmm, I've tried running things from the terminal and haven't noticed it in the log... what command were you trying to execute 

It's when I follow @scottmarlow's instructions on how to run the TCK: namely, select those profiles in the Maven View in IntelliJ, and then try to run a build from within IJ.

marko-bekhta:
hmm so I've been trying to get that warning/error ... but couldn't reproduce, as for me Intelij just fails to run anything From the failing logs I see some issue with JUnit and I wonder if it's not a version  mismatch vs Hibernate ORM or if Intelij is just acting up as usual 😖... (note that I rarely run tests from InteliJ as most of the time that just doesn't work 😔)

 I'd suggest trying to run "repair IDE" 
 
image

 and go through all the steps including clearing the cache.. maybe something got mixed from the previous import ... but can't really test on my side 😕

PR #894, opened 2026-01-13
add two crosslinks to ch2

gavinking:
Added two missing links.


gavinking:
I'm just going to go ahead and merge this one, since it's so trivial.

PR #896, opened 2026-01-13
add canonical constructors to ResultSetMapping stuff with null checks

gavinking:
I forgot to do validation in the record constructors.

### DIFF
--- a/api/src/main/java/jakarta/persistence/sql/EntityMapping.java
+++ b/api/src/main/java/jakarta/persistence/sql/EntityMapping.java
@@ -27,8 +27,8 @@ import static java.util.Objects.requireNonNull;
  * @param lockMode The lock mode acquired by the SQL query
  * @param discriminatorColumn The name of the column holding the
  *        {@linkplain jakarta.persistence.DiscriminatorColumn
- *        discriminator}; an empty string or a {@code null} value
- *        indicates that there is no discriminator column.
+ *        discriminator}; a {@code null} value indicates that
+ *        there is no discriminator column.
  * @param fields Mappings for fields or properties of the entity
  * @param <T> The entity type
  *
@@ -43,6 +43,9 @@ public record EntityMapping<T>
     public EntityMapping {
         requireNonNull(entityClass, "entityClass is required");
         requireNonNull(lockMode, "lockMode is required");
+        if (discriminatorColumn.isBlank()) {
+            throw new IllegalArgumentException("discriminatorColumn may not be blank");
+        }
         requireNonNull(fields, "fields are required");
         for (var field : fields) {
             requireNonNull(field, "field is required");
@@ -81,8 +84,8 @@ public record EntityMapping<T>
      * @param entityClass The entity class
      * @param discriminatorColumn The name of the column holding the
      *        {@linkplain jakarta.persistence.DiscriminatorColumn
-     *        discriminator}; an empty string or a {@code null} value
-     *        indicates that there is no discriminator column.
+     *        discriminator}; a {@code null} value indicates that
+     *        there is no discriminator column.
      * @param fields Mappings for fields or properties of the entity
      * @param <T> The entity type
      */
@@ -97,8 +100,8 @@ public record EntityMapping<T>
      * @param lockMode The lock mode acquired by the SQL query
      * @param discriminatorColumn The name of the column holding the
      *        {@linkplain jakarta.persistence.DiscriminatorColumn
-     *        discriminator}; an empty string or a {@code null} value
-     *        indicates that there is no discriminator column.
+     *        discriminator}; a {@code null} value indicates that
+     *        there is no discriminator column.
      * @param fields Mappings for fields or properties of the entity
      * @param <T> The entity type
      */


api/src/main/java/jakarta/persistence/sql/EntityMapping.java
```diff
@@ -62,7 +73,7 @@ public String getAlias() {
      */
     @SafeVarargs
     public static <T> EntityMapping<T> of(Class<T> entityClass, MemberMapping<T>... fields) {
-        return new EntityMapping<>(entityClass, LockModeType.NONE, "", fields);
+        return new EntityMapping<>(entityClass, LockModeType.NONE, null, fields);
```
njr-11, line 79: The Javadoc still says that "an empty string indicates that there is no discriminator column"
```suggestion
        return new EntityMapping<>(entityClass, LockModeType.NONE, "", fields);
```

It would be fine to switch to `null` instead of empty string, but corresponding updates are needed in at least 3 places in Javadoc if we do that.

api/src/main/java/jakarta/persistence/sql/EntityMapping.java
```diff
@@ -38,6 +40,15 @@ public record EntityMapping<T>
         (Class<T> entityClass, LockModeType lockMode, String discriminatorColumn, MemberMapping<?>[] fields)
         implements MappingElement<T>, ResultSetMapping<T> {
 
+    public EntityMapping {
+        requireNonNull(entityClass, "entityClass is required");
+        requireNonNull(lockMode, "lockMode is required");
+        requireNonNull(fields, "fields are required");
```
njr-11, line 49: If we go with empty string (per Javadoc) rather than `null`, then,
```suggestion
        requireNonNull(discriminatorColumn, "use empty string instead of null for discriminatorColumn to indicate none");
        requireNonNull(fields, "fields are required");
```

njr-11 (commented):
Mostly looks good, but added one comment about an inconsistency in usage of `null` (in code) vs empty string (in doc) for discriminator column.

api/src/main/java/jakarta/persistence/sql/EntityMapping.java
pzygielo: [suggested] new l. 46 doesn't look right...

njr-11: It's because the Javadoc says that empty string, not `null`, is what is used to indicate no discriminator column.  I suppose a better message to the user could be written. I'll update that.  The other option of course is change the Javadoc to use `null` instead of empty string.

pzygielo: Nope. It's because `fields` is checked (and again in l. 47) but the comment is about `discriminatorColumn`.

pzygielo: The recent edit still keeps `fields` under the null-test:
```
requireNonNull(***fields***, "use empty string...
```
where IMO it should be 
```
requireNonNull(***discriminatorColumn***, "use empty string...
```
for suggestion to make more sense.

njr-11: Thanks - I didn't notice that part.  It is updated now.

gavinking: Thanks for the catch. I had made it an empty string to match the annotation, but now I think that's silly. `null` is more correct.

api/src/main/java/jakarta/persistence/sql/EntityMapping.java
```diff
@@ -25,8 +27,8 @@
  * @param lockMode The lock mode acquired by the SQL query
  * @param discriminatorColumn The name of the column holding the
  *        {@linkplain jakarta.persistence.DiscriminatorColumn
- *        discriminator}; an empty string indicates that there
- *        is no discriminator column.
+ *        discriminator}; an empty string or a {@code null} value
+ *        indicates that there is no discriminator column.
```
gavinking: WDYT, @njr-11?

njr-11: I'd rather choose one or the other than have 2 ways of representing no discriminator column.  Code that invokes `discriminatorColumn()` would probably rather only check for `null` or empty string, but not both.  That said, what you have isn't wrong and is more a matter of opinion, so it's  good enough for an approval.

gavinking: OK well that's fine by me, but if we're going to say it has to be null then I need to reject an empty string in the constructor.

gavinking: @njr-11 There, done.

gavinking:
Done.

api/src/main/java/jakarta/persistence/sql/EntityMapping.java
```diff
@@ -38,6 +40,18 @@ public record EntityMapping<T>
         (Class<T> entityClass, LockModeType lockMode, String discriminatorColumn, MemberMapping<?>[] fields)
         implements MappingElement<T>, ResultSetMapping<T> {
 
+    public EntityMapping {
+        requireNonNull(entityClass, "entityClass is required");
+        requireNonNull(lockMode, "lockMode is required");
+        if (discriminatorColumn.isBlank()) {
```
pzygielo, line 46: What if `discriminatorColumn` is `null`? What will `.isBlank()` do?

pzygielo: A: I suppose `null` as `discriminatorColumn` will be rejected by this C-tor with `NPex`.

gavinking: Ah shit. Good catch.

gavinking: Fixed by #901. Thanks!

PR #899, opened 2026-01-14
Include the version qualifier in the rendered spec file names

marko-bekhta:
tested with snapshot/m1/(final i.e. 4.0.0) and all seem to work ok.


gavinking:
Thanks man, much appreciated!

PR #906, opened 2026-01-17
make explicit that the type conversion must be inferrable from type args

gavinking:
this was previously underspecified


marko-bekhta (approved):
nice 👍🏻

gavinking:
Thanks, @marko-bekhta. I'm going to merge this as the last one for M1.

Issue #716, opened 2025-04-16
common supertype of CriteriaDelete and CriteriaUpdate


gavinking:
`CriteriaDelete` and `CriteriaUpdate` share a number of operations and are conceptually similar. We should consider adding a supertype of these. The main problem is: what would we call it?

beikov:
I think `CriteriaMutation` or `CriteriaModification` are good names, because "mutation" or "modification" both imply changing something existing, whereas an "insert" is adding something new. That would leave the door open for adding something for insert later if needed.
A common type for delete, update and insert could be `CriteriaWrite`.

gavinking:
> I think `CriteriaMutation` or `CriteriaModification` are good names

Yes I think those names are fine.

gavinking:
`CriteriaChange` is also a possibility, it seems to me.

sebersole:
`CriteriaMutation` - +1 for me

sebersole:
> A common type for delete, update and insert could be CriteriaWrite.

Or `CriteriaMutation` and `CriteriaRestrictedMutation`

hantsy:
@gavinking Hibernate itself uses `SelectionQuery` and `MutationQuery` in the Query APIs, following these naming conventions is good for me.

hantsy:
@beikov `CriteriaRead` and `CriteriaWrite` are also clear. But only increasing the version with a specific LOCK in a selection query is read or write?

beikov:
Not sure what you're asking here. Explicit locking is obviously only relevant for select/read queries, since write queries will exclusive lock somehow in the database anyway.

gavinking:
Damn, I had forgotten about this one.

And with all the work that went into the `Query` stuff, it totally make sense for this to be an integral part of that.

And it would let us remove an overload from `EntityHandler`. Now is the right time to do that.

So I will put together a change. I might even try to sneak it into M1. (As the last thing, I promise!)

gavinking:
Done.

PR #909, opened 2026-01-18
introduce CriteriaStatement interface

gavinking:
see #716


gavinking:
Merging this so we can get M1 out.

Issue #910, opened 2026-01-19
Add @NamedStatement and @NamedNativeStatement


sebersole:
I think this is a natural continuation from the work done to split TypedQuery and Statement.  Previous versions really do not allow `@NamedQuery` and `@NamedNativeQuery` to declare mutations - it is implied by the spec and by the API itself.

This is espcially important when we consider the newly added method -

```java
interface EntityManagerFactory {
    ...
    Map<String, StatementReference> getNamedStatements();
}
```

At the moment, that method is only useful after named statements have been added (see below).  A named query (annotation) can never be a statement.

I also suggest we add a method -

```java
interface EntityManagerFactory {
    ...
    StatementReference addNamedStatement(String name, Statement statement);
}
```

Yes, there is already this method but it does not return a Reference -

```java
interface EntityManagerFactory {
    ...
    void addNamedQuery(String name, Query query);
}
```

gavinking:
Agreed on all those things.

gavinking:
Done.

PR #913, opened 2026-01-20
update spec and XSD after #910

gavinking:
cc @sebersole


gavinking:
After discussions with @sebersole, I'm merging this to have it in M1.

PR #914, opened 2026-01-20
Add docinfo to include canonical link/keywords/description

marko-bekhta:
I remembered that I had to do this for the validation spec.. so here's a patch for persistence as well.
(If you have a better description/keywords -- let me know 🙂)


gavinking:
Don't really know what this does but I trust you 😁

PR #920, opened 2026-01-21
introduce StatementOrTypedQuery and withResultSetMapping()

### LINKED ISSUE #919: StatementOrTypedQuery
I woke up today worrying about the fact that in the new design for the `Query` hierarchy  I can write silly things like:

```java
session.createStatement("...").ofType(Book.class).getResultList()
```

and wondering if the redesign was even worth it if such nonsensical things are allowed. 

At about the same time, it occurred to me that if we have a `withEntityGraph()`, we should also have a `withResultSetMapping()` for native queries. 

The solution which presented itself was to introduce a new subtype of `Query`, just to put those "typecasting" methods `ofType()`, `asStatement()`, `withEntityGraph()`, and `withResultSetMapping()`. 

But what to call this thing? Well, the best option I could come up with was `QueryOrStatement`, which at first I thought I was going to hate. But when I tried it out in the code I didn't hate it *at all*, and I actually found it pretty elegant, clarifying the differences between some of the overloaded methods on `EntityHandler` and how they are used, as well as, clearly, making the whole thing a bit more type safe.

I have implemented these ideas in a new PR I'm about to send, which sadly won't make it into M1.

I think this polishes off all the work on the `Query` hierarchy rather nicely.

gavinking:
fixes #919


gavinking:
I've also removed the `EntityGraph` mutator from `TypedQuery`, as requested by @sebersole, so that the graph for a `TypedQuery` is now immutable, and the query itself has a fixed interpretation upfront.

api/src/main/java/jakarta/persistence/EntityGraph.java
```diff
@@ -51,7 +51,8 @@
  * An entity graph passed as the first argument to
  * {@link EntityManager#find(EntityGraph, Object, FindOption...)},
  * {@link EntityManager#get(EntityGraph, Object, FindOption...)},
- * or {@link TypedQuery#setEntityGraph(EntityGraph)} is
+ * {@link EntityHandler#createQuery(String, EntityGraph)}, or
```
sebersole, line 54: Awesome!

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -40,7 +40,7 @@
  *
  * @see Query
  * @see Parameter
- * @see Query#ofType(Class)
+ * @see StatementOrTypedQuery#ofType(Class)
```
sebersole, line 43:
```suggestion
 * @see StatementOrTypedQuery#ofType(Class)
 * @see EntityHandler#createQuery(CriteriaQuery)
 * @see EntityHandler#createQuery(CriteriaSelect)
 * @see EntityHandler#createQuery(String,Class)
 * @see EntityHandler#createQuery(String,EntityGraph)
 * @see EntityHandler#createNamedQuery(String,Class)
 * @see EntityHandler#createQuery(TypedQueryReference)
 * @see EntityHandler#createNativeQuery(String,Class)
 * @see EntityHandler#createNativeQuery(String,ResultSetMapping)
```

sebersole (commented):
Suggested a minor improvement.  Might add similar `@see` tags to `Statement` as well.

gavinking:
Thanks, @sebersole, I've committed your suggestion, and now I'm going to merge this one.

PR #931, opened 2026-01-26
Avoid depending on the java.logging module

Sanne:
There's a single class in the JPA API which is producing some diagnostic logging, and this causes the module to require the 'java.logging' module.

Yet many runtimes will be plugging in a different logging system; [JEP 264](https://openjdk.org/jeps/264) was introduced in Java 9 and seems perfect for this case: it allows libraries to rely on a very minimal logging system which will delegate to the logging system chosen by the runtime.

By default it delegates to `java.util.logging` IFF the 'java.logging' module is present - so this should have no impact on people who want to use `java.util.logging`.  The upside of merging this is that it would allow minimalist systems to avoid needing this module, as in many platforms we use different logger implementations.

### DIFF
--- a/api/src/main/java/jakarta/persistence/spi/PersistenceProviderResolverHolder.java
+++ b/api/src/main/java/jakarta/persistence/spi/PersistenceProviderResolverHolder.java
@@ -26,8 +26,8 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
-import java.util.logging.Level;
-import java.util.logging.Logger;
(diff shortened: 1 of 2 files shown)


yrodiere:
FWIW there's debate... https://quarkusio.zulipchat.com/#narrow/channel/187038-dev/topic/Platform.20Logging/with/570055512

gavinking:
@Sanne You need to sign the contributor agreement.

Sanne:
Looks like I had created the commit with my usual email, which isn't recognized by the Eclipse ECA tool.

I've pushed again, should be fine now.

gavinking:
Thanks.

PR #932, opened 2026-01-27
persistence unit names

gavinking:
reflect outcome of discussion with @njr-11 on last Jakarta Data call


gavinking:
@njr-11 Please take a look over this.

gavinking:
Both Nathan and Scott have approved, so I assume this one is fine. Merging.

Issue #806, opened 2025-12-11
annotations for static query options


gavinking:
In #780 we added `@StaticQuery`, `@StaticNativeQuery`, `@ReadQueryOptions`, `@WriteQueryOptions`.

Everyone was very happy with the `@ReadQueryOptions` and `@WriteQueryOptions` annotations, and @sebersole especially expressed enthusiasm about it.

At risk of making something good worse, I'm going to propose an alternative design that is a bit more consistent with the direction we've gone with annotation design in Jakarta Data.

Essentially the point is this: over time, Java APIs have moved away from using complex structured annotations toward simpler annotations with at most one member. CDI was I guess the first spec to go down that path, but it's not the only one. `@StaticQuery`, `@StaticNativeQuery` are like that now; they only have a `value` member. (Happy, @otaviojava?)

But `@ReadQueryOptions` in particular is more of the old style annotation from an earlier period. That's OK, most of the annotations in JPA have quite a lot of members. And it does arguably make options more discoverable.

However, we should at least *consider* an alternative, which would be to separate this stuff into smaller annotations:

- `@Caching(store=BYPASS, retrieve=BYPASS)`
- `@QueryTimeout(1000)`
- `@Lock(level=PESSIMISTIC_READ)`
- `@QueryHint(name="com.example.persistence.hint", value="👻")`
- `@NamedEntityGraph(name=Customer_.GRAPH_WITH_ORDERS)`

That would be 3 new annotations instead of 2 because:

-  we could reuse the existing `@NamedEntityGraph` annotation, allowing it to specify a graph by reference or in the usual way, and
- we can reuse `@QueryHint` even more easily, just by giving it `@Target(METHOD)` and making it repeatable.

I dunno if this design is better or worse. I think it's a bit nicer for specifying timeouts and query hints, but not hugely better for the others.

Anyway, I would like to hear feedback from others.

There's no particular urgency on this one.

gavinking:
A different possible change: even if we don't do the thing proposed above, we might want to rename these annotations `@QueryOptions` and `@StatementOptions`. After M1 I mean.

sebersole:
In general I do really prefer more granular approaches.  

But here I'm just not seeing it.  What does the aggregation of this information look like?  My biggest uncertainty is the fact that the named-query annotations are repeatable - how do you "link" these to the query they describe?  Matched names?  (barf).

```java
@Entity
@NamedQuery(name="byLastName", query="...")
[...]

PR #933, opened 2026-01-27
align naming of QueryOptions/StatementOptions with new Query/Statement distinction

gavinking:
See #806.


gavinking:
I believe this one should just be merged. Any objections??

njr-11 (commented):
I'm not sure these names are an improvement over what we already had. I suppose it depends on what we see as the more dominant use case for them.

If we were only considering Jakarta Data, the term Statement doesn't have a particular meaning. Some users might think of java.sql.Statement, which is a common to queries that perform find operations as well as deletes and updates and even insert. In Jakarta Data, `UPDATE` and `DELETE` are performed by `@Query`, `@StaticQuery`, and `@StaticNativeQuery` which fit nicely with the existing `@WriteQueryOptions` (or `@ReadQueryOptions` if instead a `SELECT` query). The position from Jakarta Data should be that existing names are a very good fit, and a change to `@StatementOptions` and `@TypedQueryOptions` do not match up with existing terminology in the spec, lessening the usability.

Considering Jakarta Persistence, the alignment with new API classes there makes more sense.

I'm probably biased, but I would opt for the better alignment with Jakarta Data here.

gavinking:
Hrmph OK I had not looked at it like that.

I don't have a useful response right now, let me think it over.

gavinking:
#984 proposes a simpler way out of this conundrum.

sebersole:
This one is obsolete now given https://github.com/jakartaee/persistence/pull/984 was merged, no?

gavinking:
Yeah, sure, closing.

PR #936, opened 2026-01-30
specialized expression types for easier JPA Criteria

gavinking:
@beikov had already done most of the hard work in #522.

This is a light reworking of that PR.

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/NumericExpression.java
+++ b/api/src/main/java/jakarta/persistence/criteria/NumericExpression.java
@@ -27,7 +27,8 @@ import java.math.BigInteger;
  *
  * @since 4.0
  */
-public interface NumericExpression<N extends Number & Comparable<N>> extends ComparableExpression<N> {
+public interface NumericExpression<N extends Number & Comparable<N>>
+        extends ComparableExpression<N> {
 
     // comparisons
 
--- a/api/src/main/java/jakarta/persistence/criteria/ComparableExpression.java
+++ b/api/src/main/java/jakarta/persistence/criteria/ComparableExpression.java
@@ -24,7 +24,8 @@ package jakarta.persistence.criteria;
  *
  * @since 4.0
  */
-public interface ComparableExpression<C extends Comparable<? super C>> extends Expression<C> {
+public interface ComparableExpression<C extends Comparable<? super C>>
+        extends Expression<C> {
 
     // comparisons
 
--- a/api/src/main/java/jakarta/persistence/criteria/TemporalExpression.java
+++ b/api/src/main/java/jakarta/persistence/criteria/TemporalExpression.java
@@ -23,7 +23,8 @@ import java.time.temporal.Temporal;
  *
  * @since 4.0
  */
-public interface TemporalExpression<T extends Temporal & Comparable<? super T>> extends ComparableExpression<T> {
+public interface TemporalExpression<T extends Temporal & Comparable<? super T>>
+		extends ComparableExpression<T> {
 
 	/**
 	 * Create an expression that returns the value of a


gavinking:
Still need to figure out what to do about `BooleanExpression` vs `Predicate`. We didn't bother with a `BooleanExpression` in Data 1.1. I'm OK with leaving it in here, but if are going to have a `BooleanExpression` type, we don't need `Predicate` anymore and we can deprecate it.

api/src/main/java/jakarta/persistence/criteria/NumericExpression.java
```diff
@@ -0,0 +1,388 @@
...
+    /**
+     * Create an expression that returns the sign of this
+     * expression, that is, {@code 1} if this expression is
+     * positive, {@code -1} if this expression is negative,
+     * or {@code 0} if this expression is exactly zero.
+     * @return sign
+     * @see CriteriaBuilder#sign(Expression)
+     */
+    NumericExpression<Integer> sign();
+
```
beikov, line 118: Seems you're missing the `mod` function here. Or was the omission on purpose?

```suggestion

    /**
     * Create an expression that returns the modulus
     * (remainder under integer division) of this
     * expression divided by the argument.
     * @param y expression
     * @return modulus
     * @see CriteriaBuilder#mod(Expression, Expression)
     */
    NumberExpression<Integer> mod(Expression<Integer> y);

    /**
     * Create an expression that returns the modulus
     * (remainder under integer division) of this
     * expression divided by the argument.
     * @param y value
     * @return modulus
     * @see CriteriaBuilder#mod(Expression, Integer)
     */
    NumberExpression<Integer> mod(Integer y);

```

gavinking:
I have tried replacing `Predicate` with `BooleanExpression` or `Expression<Boolean>` in locations where that would not break backward compatibility. It's not perfect, we still have `Predicate` in some return types, but it's not _terrible_ either.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
```diff
@@ -72,7 +72,7 @@ public interface AbstractQuery<T> extends CommonAbstractCriteria {
      * @param restrictions  zero or more restriction predicates
      * @return the modified query
      */
-    AbstractQuery<T> where(Predicate... restrictions);
+    AbstractQuery<T> where(BooleanExpression... restrictions);
```
jwgmeligmeyling, line 75: Will the JPQL specification also allow boolean expressions here? I.e. `where TRUE` instead of `where TRUE = TRUE`. Or do users have to be cautious that boolean expressions provided here, actually have to be predicates still? In the case of the latter I wouldn't really see the point of this type change.

gavinking: > Will the JPQL specification also allow boolean expressions here? 

I think that probably makes sense, yes.

gavinking: Anyway, that is issue https://github.com/jakartaee/query/issues/110

gavinking:
I do believe this is ready to go. In the end I did not see any way to really untangle `Predicate` without major breakage to existing programs. But it's fine, we can live with that.

Also I have aligned naming of numeric operations with what we have in Jakarta Data.

api/src/main/java/jakarta/persistence/criteria/NumericExpression.java
gavinking: It's intentional. Integer division is only defined for `int`/`Integer` operands, not for other numeric types.

gavinking:
> I do believe this is ready to go. 

Spoke too soon. I have made the following additional changes:

1. Removed the unnecessary `XxxxPath` types which represented terminal paths. The API is nicer if we just use `XxxxExpression` for this.
2. Aligned naming with Jakarta Data, so `TextExpression` instead of `StringExpression`.
2. Aligned operations of `TextExpression` with Jakarta Data.
3. Simplified naming of `XxxxAttribute` types, and made `NumericAttribute` a `ComparableAttribute`.
4. Fixed a couple of generic method signatures.
5. Changed `CollectionExpression` to `PluralExpression` so it can represent `Map`s.

gavinking:
**Now** I think this is ready to go. If there are further adjustments to be made, they can be done in separate pull requests. Need to (finally) get this merged so that people can experiment on the implementation side.

gavinking:
Going to pull the trigger on this in 3 ... 2 ... 1 ....

gavinking:
Merged. If anyone has further suggestions, send a PR or open an issue.

Issue #937, opened 2026-02-02
Add @Discoverable for allowing flexible managed-class discovery


sebersole:
Add a `@Discoverable` meta-annotation to allow more flexible discovery of managed-classes.

We'd add this to `@Entity`, `@MappedSuperclass`, etc.

But it also allows to discover classes with provider-specific annotations as well.

----
Original :

The main purpose of `PersistenceUnitInfo#getAllManagedClassNames` is to shift responsibility for "scanning" to containers rather than the providers.  One part of that which is very difficult to swallow is discovery of classes containing custom annotations.  Hibernate does this as I assume other providers do as well.

I think a simple solution here would be for us (Hibernate) to tell users to place such annotations on package-info and have the container simply tell us about all discovered `package-info.class` files as part of `PersistenceUnitInfo.getAllManagedClassNames()`.  Other options (providers telling containers about "extra annotations" to look for, e.g.) are just too difficult to spec out and implement.

sebersole:
An alternative solution we've been discussing on Hibernate channels is a meta-annotation, say `@Discoverable`.  E.g.

```java
@Discoverable
public @interface Entity {...}

@Discoverable
public @interface MappedSuperClass {...}

...
```

which could be used by providers for custom annotations as well

```java
@Discoverable
public @interface HibernateSpecificAnnotation { ... }
```

gavinking:
A fantastic side effect of this is that we can finally allow named queries to be declared by module and package descriptors.

gavinking:
Done.

PR #940, opened 2026-02-03
introduce @Discoverable

gavinking:
see #937

### DIFF
--- a/api/src/main/java/jakarta/persistence/SequenceGenerator.java
+++ b/api/src/main/java/jakarta/persistence/SequenceGenerator.java
@@ -18,6 +18,8 @@
 
 package jakarta.persistence;
 
+import jakarta.persistence.spi.Discoverable;
+
 import java.lang.annotation.Target;
 import java.lang.annotation.Retention;
 import static java.lang.annotation.ElementType.METHOD;
@@ -60,6 +62,7 @@ import java.lang.annotation.Repeatable;
 @Repeatable(SequenceGenerators.class)
 @Target({TYPE, METHOD, FIELD, PACKAGE})
 @Retention(RUNTIME)
+@Discoverable
 public @interface SequenceGenerator {
 
     /** 
--- a/api/src/main/java/jakarta/persistence/TableGenerator.java
+++ b/api/src/main/java/jakarta/persistence/TableGenerator.java
@@ -18,6 +18,8 @@
 
 package jakarta.persistence;
 
+import jakarta.persistence.spi.Discoverable;
+
 import java.lang.annotation.Target;
 import java.lang.annotation.Retention;
 import static java.lang.annotation.ElementType.FIELD;
@@ -95,6 +97,7 @@ import java.lang.annotation.Repeatable;
 @Repeatable(TableGenerators.class)
 @Target({TYPE, METHOD, FIELD, PACKAGE})
 @Retention(RUNTIME)
+@Discoverable
 public @interface TableGenerator {
 
     /** 


gavinking:
@sebersole @scottmarlow @njr-11 @anija-anil  Please review.

This is as per discussions we've had on Zulip.

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
```diff
@@ -164,17 +164,26 @@ public interface PersistenceUnitInfo {
...
+     * named statement, named stored procedure query, or
+     * SQL result set mapping. Classes bearing custom
+     * discoverable annotation types are also discoverable.
      * @return the list of names of all classes belonging
      * to the persistence unit
+     * @see Discoverable
      * @since 4.0
      */
-    List<String> getAllManagedClassNames();
+    List<String> getAllClassNames();
```
scottmarlow, line 188: I'm not sure that it needs to be stated but are there any concerns about the amount of storage needed to contain the returned List and whether the reference to the list should be cleared (by the provider) after the Persistence Unit deploys (if not sooner)?  The EE container should probably also clear its copy of the List as well after the Persistence Unit deploys.

sebersole: For what it is worth, Hibernate at least does not keep references to this beyond bootstrap.

sebersole (approved):

scottmarlow (approved):

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
gavinking: I assume that the provider is just going to iterate over the list and then discard it. I don't see a problem here.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -253,18 +253,22 @@ persistence classes to be included within a persistence unit:
...
-<<named-queries-statements>> must also be implicitly or explicitly denoted
+Furthermore, other discoverable types, as defined below in <<discoverable>>,
+including types declaring named queries and statements, as defined in
+<<named-queries-statements>>, must also be implicitly or explicitly denoted
 as belonging to a persistence unit.
 
+[NOTE]
+Terminology Note: In this context, the word "class" refers to compiled Java
+`.class` files, and implicitly includes both Java types and Java module and
+package descriptors.
```
gavinking, line 264: @sebersole 👀

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
```diff
@@ -156,25 +156,36 @@ public interface PersistenceUnitInfo {
     List<String> getManagedClassNames();
 
     /**
-     * Returns the names of all classes belonging to the
-     * persistence unit, including classes:
+     * Returns the names of all classes (compiled Java types,
+     * module descriptors, and package descriptors) belonging
+     * to the persistence unit, including classes:
```
gavinking, line 161: @sebersole 👀

api/src/main/java/jakarta/persistence/spi/Discoverable.java
```diff
@@ -0,0 +1,42 @@
...
+import static java.lang.annotation.ElementType.ANNOTATION_TYPE;
+import static java.lang.annotation.RetentionPolicy.CLASS;
+
+/**
+ * Specifies that the meta-annotated annotation type identifies
+ * classes which are automatically discovered by the Jakarta EE
+ * container and passed on to the persistence provider via
+ * {@link PersistenceUnitInfo#getAllClassNames()}. Such classes
+ * might include compiled Java types, module descriptors, and
+ * package descriptors.
```
gavinking, line 30: @sebersole 👀

spec/src/main/asciidoc/ch08-entity-packaging.adoc
sebersole: In fact it always does ;)

gavinking: haha

api/src/main/java/jakarta/persistence/spi/Discoverable.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p>
+ * A persistence provider may define its own custom discoverable
+ * annotation types. Classes annotated with custom discoverable
+ * annotations must be included in the list of class names passed
+ * by the container to the provider.
+ *
+ * @since 4.0
+ */
+@Retention(CLASS)
```
sebersole: Shouldn't this be RUNTIME?

gavinking: If you prefer. I asked you and you didn't give me s straight answer :)

If `@Discoverable` is *only* meant for tools like jandex then `CLASS` is appropriate, I believe.

But if you see it being useful in other ways then yeah we should promote it to `RUNTIME`.

sebersole: `CLASS` might work for us because, as you say, we will use Jandex for discovery and it handles `CLASS` retention here fine (I believe).  But other providers probably use another mechanism, so :shrug: .

I think `RUNTIME` would be safest - but I have no idea of your thought process behind picking `CLASS` in the first place so might be missing something.  I just rarely seen that retention used, so I figured you had your reasons.

gavinking: Fine, `RUNTIME` it is.

I've been waiting decades of my life to write an annotation which was anything other than `RUNTIME`, and I wrongly thought my moment had finally arrived.

gavinking: Done 😭

gavinking:
Alright, so I think this one is good to go...

Issue #599, opened 2024-03-19
duplicate persistent field names in hierarchy


gavinking:
I just noticed that JPA does not say what happens in the following circumstance:

```java
@Entity Super { @Id long id; String name; }
```
```java
@Entity Sub extends Super { String name; }
```

Nor in this case:

```java
@Entity Super { @Id long id; @Column("super_name") String name; }
```
```java
@Entity Sub extends Super { @Column("sub_name") String name; }
```

I would say that these are disallowed, or at least "not required/portable" and strongly discouraged. But we don't actually say that.

PR #945, opened 2026-02-05
duplicate persistent field names in hierarchy

gavinking:
Update spec to address cases raised in #599.

### DIFF
--- a/spec/src/main/asciidoc/ch02-entities.adoc
+++ b/spec/src/main/asciidoc/ch02-entities.adoc
@@ -2665,33 +2665,32 @@ public class Employee extends Person {
 }
 ----
 
-==== Duplicate Column Mappings in Inheritance Hierarchies
+==== Duplicate Column Mappings
 
-With an important exception noted below:
+Two persistent attributes (declared by the same entity class or
+by distinct entity classes) are permitted to map the same column
+of a given table if and only if:
 
-- All persistent attributes declared by a given entity class must
-  map to distinct columns. This restriction is automatically
-  satisfied by default, due to the restriction specified in above
-  in <<dupe-attributes>>. It can be violated if two `@Column` or
-  `@JoinColumn` annotations specify the same column name. It is
-  the responsibility of the application developer to avoid such
-  mappings.
+- the persistent attributes have exactly the same type, and
+- either the two attributes are declared in distinct branches of
+  the entity class hierarchy, so that no entity class features
+  both attributes, or one of the persistent attributes is declared
+  with `insertable=false, updatable=false`, as defined later in
+  <<metadata-annotations>>.
 
-- In addition, when the single table inheritance mapping strategy
-  defined in <<single-table>> is used to map an entity class
-  inheritance hierarchy, all persistent attributes defined across
-  the whole hierarchy must map to distinct columns. It is the
-  responsibility of the application developer to satisfy this
-  restriction by explicitly specifying distinct column mappings
-  for persistent attributes with identical names.
+It is the responsibility of the application developer to satisfy
+this restriction. In an entity class hierarchy mapped via the
+single table strategy or table per concrete class strategy, the
+developer might need to explicitly specify distinct column mappings
+for persistent attributes with identical names.
 
-- Similarly, when the table-per-concrete class inheritance mapping
-  strategy defined in <<table-per-concrete-class>> is used to map
-  an entity class inheritance hierarchy, all persistent attributes
-  declared or inherited by a given entity class must map to distinct
-  columns. It is the responsibility of the application developer to
-  satisfy this restriction by explicitly specifying distinct column
-  mappings for persistent attributes with identical names.
+The behavior is undefined and unportable if:
+
+- any two distinct persistent attributes with distinct types map
+  the same column of a given table, or
+- two distinct persistent attributes declared or inherited by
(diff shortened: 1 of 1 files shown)


gavinking:
@sebersole @mbellade does look right?

sebersole (commented):
I think the semantics you define are logical.  I do think the section could benefit from some sample code to help illustrate what is and is not supported.

gavinking:
OK, will do.

gavinking:
@sebersole I have added some examples. Is this what you were looking for?

spec/src/main/asciidoc/ch02-entities.adoc
beikov, line 2592: Not sure if this is clear, but that is not the behavior of Hibernate ORM right now. Since you write above that the behavior is not portable, is there a reason to keep this definition?
In Hibernate ORM we ignore a `FIELD` access persistent attribute of a subclass if there is a `FIELD` access persistent attribute in a superclass:

```
@MappedSuperclass
public class Base {
  @Id Long id;
  String name;
}
@Entity
public class Sub extends Base {
  String name;
}
```

The field `Sub#name` will not be considered, only `Base#name`.

gavinking: > that is not the behavior of Hibernate ORM right now

Well, it seems to me that this is a bug. Do we say anywhere that it's not allowed?

> Since you write above that the behavior is not portable, is there a reason to keep this definition?

Not sure what you mean by this.

> In Hibernate ORM we ignore a `FIELD` access persistent attribute of a subclass if there is a `FIELD` access persistent attribute in a superclass:

Well, that doesn't appear to respect the semantics of the Java language.

gavinking: > Not sure what you mean by this.

Oh, now I see. Indeed I might be contradicting myself. Hold on.

gavinking: Oh no, everything is fine I believe.

If we're going to say "you can't have two distinct attributes with the same name", then of course you need to define what counts as "distinct". Otherwise that sentence is ambiguous.

gavinking: >> In Hibernate ORM we ignore a FIELD access persistent attribute of a subclass if there is a FIELD access persistent attribute in a superclass:
>
> Well, it seems to me that this is a bug.

To be clear: the bug here is in just ignoring it. It would be fine to throw.

beikov: >Well, it seems to me that this is a bug. Do we say anywhere that it's not allowed?

Not sure if we're documenting it somewhere, but I had to go back to ORM 5.3 to check the behavior because of a bug report and that's why I know Hibernate ORM behaves this way for a long time.

>If we're going to say "you can't have two distinct attributes with the same name", then of course you need to define what counts as "distinct". Otherwise that sentence is ambiguous.

Ok, that was my confusion I guess. When reading the sub-section I thought it was saying that we must support such same named fields as distinct attributes, but I now understand this is just definition.

>To be clear: the bug here is in just ignoring it. It would be fine to throw.

The first section says the behavior is not portable, so I guess whatever we do is fine?

gavinking: >  I guess whatever we do is fine?

Whatever we do is *compliant with the spec*. But we should still throw.

beikov: The example (allowed) you added here does not fully work on Hibernate ORM today, because we assume that we can use the persistent attribute name as key in a map. Talking about this:

```java
@Entity
@Inheritance(strategy = SINGLE_TABLE)
public class Person {
    @Id
    private Long id;
}

@Entity
public class Employee extends Person {
    @Column(name="EMPLOYEE_CODE", length=8)
    private String code;
}

@Entity
public class Customer extends Person {
    @Column(name="CUSTOMER_CODE")
    private int code;
}
```

gavinking: > because we assume that we can use the persistent attribute name as key in a map.

Well we should _definitely_ fix that.

beikov: Oops, sorry. I misread. I thought `Customer` extended `Employee`, such that `Customer` has two fields and persistent attributes with the name `code`. Do we agree that this is not portable?

```java
@MappedSuperclass
public class Base {
  @Id Long id;
  String name;
}
@Entity
public class Sub extends Base {
  String name;
}
```

On the other hand, the following example is valid in Hibernate ORM right now:

```java
@Entity
@Inheritance(strategy = SINGLE_TABLE)
public class Person {
    @Id
    private Long id;
}

@Entity
public class Employee extends Person {
    @Column(length=8)
    private String code;
}

@Entity
public class Customer extends Person {
    // Disallowed: maps to the same CODE column as Employee.code
    private int code;
}
```

gavinking: Yes, and yes.

gavinking:
I have generalized and simplified the restriction on duplicate column mappings.

gavinking:
Alright, I guess this clarification has had enough time to stew now, and three different people have reviewed. Time to merge it.

PR #947, opened 2026-02-06
[#636, #934, #935] Add a way to customize SQL DDL for Index

### LINKED ISSUE #636: Add a way to define custom SQL DDL for Index
The spec should allow user to define his own, custom DDL for Index definition to support usage of DB specific features, ie MySQL `CREATE [UNIQUE | FULLTEXT | SPATIAL] INDEX ...` or Oracle DB `CREATE [BITMAP | MULTIVALUE | VECTOR ] INDEX ...` We already have `@Column.columnDefinition`, so we could add something like `@Index.indexDefinition`

### LINKED ISSUE #934: Add @Index kind and type members
Some databases have a "kind" and a "type" configuration option for indexes, which can't be handled with the `options` member that was introduced in JPA 3.2. See the documentation links:

* MySQL has `UNIQUE`, `FULLTEXT`, `SPATIAL` kinds and `BTREE` and `HASH` types: https://dev.mysql.com/doc/refman/8.4/en/create-index.html
* MariaDB has `UNIQUE`, `FULLTEXT`, `SPATIAL`, `VECTOR` kinds and `BTREE`, `RTREE` and `HASH` types: https://mariadb.com/docs/server/reference/sql-statements/data-definition/create/create-index
* H2 has `UNIQUE`, `SPATIAL` kinds: https://www.h2database.com/html/commands.html#create_index
* PostgreSQL has index methods, which are like types: https://www.postgresql.org/docs/current/sql-createindex.html
* SQL Server has `UNIQUE`, `CLUSTERED` and `NON CLUSTERED` kinds: https://learn.microsoft.com/en-us/sql/t-sql/statements/create-index-transact-sql?view=sql-server-ver17
* Oracle has `UNIQUE`, `BITMAP`, `MULTIVALUE` kinds: https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/CREATE-INDEX.html and a special `HYBRID VECTOR` kind: https://docs.oracle.com/en/database/oracle/oracle-database/26/sqlrf/create-hybrid-vector-index.html

I hope this justifies adding this to `Index`:

```java
    /**
     * (Optional) A SQL fragment representing the index kind,
     * usually inserted before the keyword {@code index} in the generated DDL
     * which creates this index.
     *
     * @since 4.0
     */
    String kind() default "";

    /**
     * (Optional) A SQL fragment representing the index type.
     * When given and the database supports it, is inserted into the generated DDL
     * which creates this index.
     *
     * @since 4.0
     */
    String type() default "";
```

The reason I'm asking for this is that I would like to be able to specify a vector index on PostgreSQL which requires me to specify the `using ivfflat` clause that I can't do right now.

### LINKED ISSUE #935: Relax the @Index columnList BNF requirement
Some databases have additional options that can be provided for a column e.g. `nulls first`/`nulls last` or index access operations (e.g. on PostgreSQL), which currently can't be specified.
I suggest that we change the `columnList` BNF to:

```
column_list ::= index_column [, index_column]*
index_column ::= column_name (\b[^,]*)?
```

From what I can tell, usually only the column name is interesting to JPA providers for validation, though I'm not sure if that's really useful. Maybe we just allow free text so that people can also use expressions instead of just column names?

The reason I'm asking for this is that I would like to be able to specify a vector index on PostgreSQL which requires me to specify the `vector_cosine_ops` for an embedding column.

beikov:
Fixes #636. Fixes #934. Fixes #935.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Index.java
+++ b/api/src/main/java/jakarta/persistence/Index.java
@@ -23,9 +23,9 @@ import java.lang.annotation.Target;
 /**
  * Used in schema generation to specify creation of an index.
  *
- * <p>The syntax of the {@code columnList} element is SQL
- * dialect specific, but usually a comma separated list of
- * column names, with optional ordering specification
+ * <p>The syntax of the {@code columnList} element depends on
+ * the SQL dialect, but is usually a comma-separated list of
+ * column names, with an optional ordering specification,
  * {@code ASC} or {@code DESC}.
  *
  * <p>Note that it is not necessary to specify an index for a
@@ -64,8 +64,8 @@ public @interface Index {
 
     /**
      * (Optional) A SQL fragment representing the index kind,
-     * usually inserted before the keyword {@code index} in the generated DDL
-     * which creates this index.
+     * usually inserted before the keyword {@code index} in 
+     * the generated DDL which creates this index.
      *
      * @since 4.0
      */
@@ -73,7 +73,7 @@ public @interface Index {
 
     /**
      * (Optional) A SQL fragment representing the index type.
-     * When given and the database supports it, is inserted into the generated DDL
+     * When specified, it is inserted into the generated DDL 
      * which creates this index.
      *
      * @since 4.0
--- a/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
+++ b/spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
@@ -2352,10 +2352,9 @@ public @interface Index {
 }
 ----
 
-The syntax of the `columnList` element is SQL
-dialect specific, but usually a comma separated list of
-column names, with optional ordering specification
-`ASC` or `DESC`.
+The syntax of the `columnList` element depends on the SQL dialect,
+but is usually a comma-separated list of column names, with an
+optional ordering specification, `ASC` or `DESC`.
 
 <<a14877>> lists the annotation elements that may be specified
 for the `Index` annotation.
@@ -2388,7 +2387,7 @@ included in the index.
 
 |String
 |type
-|(Optional) A SQL fragment representing the index type. When given and the database supports it, is inserted into the generated DDL which creates this index.
+|(Optional) A SQL fragment representing the index type. When specified, it is inserted into the generated DDL which creates this index.
 |Nothing appended.
 
 |String


api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -27,7 +28,7 @@
  * 
  * {@snippet :
  *     column_list ::= index_column [, index_column]*
- *     index_column ::= column_name [ASC | DESC]
+ *     index_column ::= column_name (\b[^,]*)?
  * }
```
gavinking: I would just delete this BNF entirely. We don't need it.

beikov: How about this?

```suggestion
<p>The syntax of the {@code columnList} element is SQL
 * dialect specific, but usually a comma separated list of
 * column names, with optional ordering specification
 * {@code ASC} or {@code DESC}.
```

gavinking: Perfect.

api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -22,16 +23,10 @@
...
- *     column_list ::= index_column [, index_column]*
- *     index_column ::= column_name [ASC | DESC]
- * }
- * 
- * <p>If neither {@code ASC} nor {@code DESC} is not specified,
- * {@code ASC}, that is, ascending order, is assumed.
+ * <p>The syntax of the {@code columnList} element is SQL
+ * dialect specific, but usually a comma separated list of
+ * column names, with optional ordering specification
+ * {@code ASC} or {@code DESC}.
```
gavinking:
```suggestion
 * <p>The syntax of the {@code columnList} element depends on
 * the SQL dialect, but is usually a comma-separated list of
 * column names, with an optional ordering specification,
 * {@code ASC} or {@code DESC}.
```

api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -67,6 +62,24 @@
      */
     boolean unique() default false;
 
+    /**
+     * (Optional) A SQL fragment representing the index kind,
+     * usually inserted before the keyword {@code index} in the generated DDL
+     * which creates this index.
```
gavinking:
```suggestion
     * (Optional) A SQL fragment representing the index kind,
     * usually inserted before the keyword {@code index} in 
     * the generated DDL which creates this index.
```

api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -67,6 +62,24 @@
...
+     *
+     * @since 4.0
+     */
+    String kind() default "";
+
+    /**
+     * (Optional) A SQL fragment representing the index type.
+     * When given and the database supports it, is inserted into the generated DDL
+     * which creates this index.
```
gavinking, line 77:
```suggestion
     * (Optional) A SQL fragment representing the index type.
     * When specified, it is inserted into the generated DDL 
     * which creates this index.
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -2387,6 +2381,16 @@ included in the index.
...
 |false
 
+|String
+|kind
+|(Optional) A SQL fragment representing the index kind, usually inserted before the keyword `index` in the generated DDL which creates this index.
+|Nothing appended.
+
+|String
+|type
+|(Optional) A SQL fragment representing the index type. When given and the database supports it, is inserted into the generated DDL which creates this index.
```
gavinking:
```suggestion
|(Optional) A SQL fragment representing the index type. When specified, it is inserted into the generated DDL which creates this index.
```

spec/src/main/asciidoc/ch11-metadata-for-or-mapping.adoc
```diff
@@ -2346,22 +2346,16 @@ public @interface Index {
...
-----
-
-The persistence provider must observe the specified ordering of the columns.
-
-If `ASC` or `DESC` is not specified, `ASC`
-(ascending order) is assumed.
+The syntax of the `columnList` element is SQL
+dialect specific, but usually a comma separated list of
+column names, with optional ordering specification
+`ASC` or `DESC`.
```
gavinking:
```suggestion
The syntax of the `columnList` element depends on the SQL dialect,
but is usually a comma-separated list of column names, with an
optional ordering specification, `ASC` or `DESC`.
```

gavinking (approved):
Thanks @beikov. I believe this is ready, and I will merge it later today, assuming no-one else speaks up.

PR #951, opened 2026-02-19
Address some TCK issues

marko-bekhta:
Hey @gavinking  🙂 👋🏻 

I've tried looking closer into some of the issues with the TCK...
- added the stored procedures as "before step" in the corresponding test. It's ugly and goes through `em.createNativeQuery` to create the procedures ... but at least we won't lose those definitions ?
- added a `drop-create` as a default on all the tests and in the few where the tests were setting the schema-generation action explicitly added an override to ignore the "default"
- some tests had issues where the same table name was used by multiple entities within the same PU leading to various issues. I don't think that the intention of those tests were to enforce the behaviour under such circumstances .. os I just went ahead and applied more "unique" names there.
- a couple tests were trying to assign the same Uni-Directional `OneToMany` to multiple "owners" and I think we actually require a unique constraint there ... so I made a couple changes to the tests doing that.

### DIFF
--- a/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/id/Client.java
+++ b/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/annotations/id/Client.java
@@ -19,9 +19,11 @@ package ee.jakarta.tck.persistence.core.annotations.id;
 import java.lang.System.Logger;
 import java.math.BigDecimal;
 import java.math.BigInteger;
+import java.math.RoundingMode;
 
 import org.jboss.shrinkwrap.api.spec.JavaArchive;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -215,7 +217,7 @@ public class Client extends PMClientBase {
 
 		try {
 			getEntityTransaction().begin();
-			final BigDecimal id = new BigDecimal(new BigInteger("1"));
+			final BigDecimal id = new BigDecimal(new BigInteger("1")).setScale(2, RoundingMode.UP);
 
 			FieldBigDecimalId expected = new FieldBigDecimalId(id, id);
 
@@ -227,7 +229,7 @@ public class Client extends PMClientBase {
 			getEntityTransaction().begin();
 			FieldBigDecimalId actual = getEntityManager().find(FieldBigDecimalId.class, id);
 			if (actual != null) {
-				if (actual.getBigDecimal().equals(id)) {
+				if (actual.getBigDecimal().setScale(2, RoundingMode.UP).equals(id)) {
 					logger.log(Logger.Level.TRACE, "Received expected result:" + actual.getBigDecimal());
 					pass = true;
 				} else {
@@ -544,7 +546,7 @@ public class Client extends PMClientBase {
 
 		try {
 			getEntityTransaction().begin();
-			final BigDecimal id = new BigDecimal(new BigInteger("1"));
+            final BigDecimal id = new BigDecimal(new BigInteger("1")).setScale(2, RoundingMode.UP);
 
 			PropertyBigDecimalId expected = new PropertyBigDecimalId(id, id);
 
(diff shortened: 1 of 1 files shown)


gavinking:
> added the stored procedures as "before step" in the corresponding test.  It's ugly and goes through em.createNativeQuery to create the procedures 

Can't you use `SchemaManager.exportMappedObjects()` together with a script source?

gavinking:
> a couple tests were trying to assign the same Uni-Directional OneToMany to multiple "owners" and I think we actually require a unique constraint there ... so I made a couple changes to the tests doing that.

WDYM by this? Which tests?

tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/relationship/annotations/Client.java
```diff
@@ -781,17 +781,18 @@ public void annotationMappingTest7() throws Exception {
...
 		int foundCol4 = 0;
 		int foundCol5 = 0;
 		final Integer[] expectedCol1 = new Integer[] { 1, 2, 3, 4 };
 		final Integer[] expectedCol2 = new Integer[] { 5, 6 };
-		final Integer[] expectedCol3 = new Integer[] { 3, 8 };
-		final Integer[] expectedCol4 = new Integer[] { 4, 7 };
-		final Integer[] expectedCol5 = new Integer[] { 1, 2, 3, 4 };
+		final Integer[] expectedCol3 = new Integer[] { 7, 8 };
+		final Integer[] expectedCol4 = new Integer[] { 9 };
+		final Integer[] expectedCol5 = new Integer[] { 10 };
```
marko-bekhta, line 795: > > a couple tests were trying to assign the same Uni-Directional OneToMany to multiple "owners" and I think we actually require a unique constraint there ... so I made a couple changes to the tests doing that.

> WDYM by this? Which tests?

it's this type of tests ^ (there's a couple like it)

we have on the person side:
```java
	/* Uni-Directional Single-Valued OnePersonsToManyReviews */
	@OneToMany(cascade = CascadeType.ALL)
	@JoinTable(name = "PERSON_ANNUALREVIEW", joinColumns = @JoinColumn(name = "Person_PERSONID", referencedColumnName = "PERSONID"), inverseJoinColumns = @JoinColumn(name = "annualReviews_AID", referencedColumnName = "AID"))
	public Collection<AnnualReview> getAnnualReviews() {
		return annualReviews;
	}
```

but then the test was trying to assign e.g. Review `3` to person `1` and `3` at the same time...

gavinking: Oh well that would be a many-to-many, not a one-to-many.

So yeah, that's clearly wrong.

marko-bekhta:
> Can't you use SchemaManager.exportMappedObjects() together with a script source?

ah I think that's Hibernate extension 🙈 ... let me see if I can do something with `SchemaManager#populate()` instead...

gavinking:
Sorry, my bad, I meant `export()`. 

`populate()` would probably work but it's a bit of an abuse.

gavinking:
> Sorry, my bad, I meant `export()`.

No, I meant `create()`.

tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/StoredProcedureQuery/Client1.java
```diff
@@ -68,13 +75,30 @@ public void setup() throws Exception {
...
             dataBaseName = System.getProperty("jdbc.db");
         } catch (Exception e) {
             logger.log(Logger.Level.ERROR, "Exception: ", e);
             throw new Exception("Setup failed:", e);
         }
     }
 
+    @Override
+    protected Map<String, Object> extraPersistenceUnitProperties() {
+        return Map.of(Persistence.SchemaManagementProperties.SCHEMAGEN_LOAD_SCRIPT_SOURCE,
```
gavinking: So this is kind of an abuse because the "load script" is supposed to be used for loading data.

You're supposed to use `SCHEMAGEN_CREATE_SCRIPT_SOURCE` for this.

marko-bekhta: right .... 👍🏻
let me give it another try  🙈 🫣 🙂

gavinking:
Is this ready to go now, @marko-bekhta ?

marko-bekhta:
yes, I think so. Thanks! 🙇🏻 🙂

Issue #948, opened 2026-02-08
standard BatchSize option


gavinking:
All implementations support some sort of batch loading or batch fetching. I therefore propose that we add:

```java
/**
 * Specifies a batch size, that is, how many entities should be
 * fetched in each request to the database. This option is always
 * a hint, and might be ignored by the persistence provider.
 *
 * @since 4.0
 */
public record BatchSize(int batchSize) implements FindOption, FetchOption {
}
```

Perhaps we should have added this in JPA 3.2, but we didn't.

gavinking:
If we're going to add this, it might also make sense to have a standard way to specify the batch fetch size globally and/or per-`EntityManager`.

gavinking:
Added.

gavinking:
> it might also make sense to have a standard way to specify the batch fetch size globally and/or per-`EntityManager`

I did not do this part of it. If anyone wants this, please speak up.

PR #955, opened 2026-02-20
introduce BatchSize option

gavinking:
See #948.


gavinking:
@sebersole WDYT?

gavinking:
OK, so let's just do this:

- it's a hint, so imposes no real burden on implementors
- and it has a zero blast radius
- nevertheless, it's a quite common "option"

Issue #787, opened 2025-11-28
query-specific flush modes


gavinking:
While working on #780, I again ran into the mismatch between `FlushModeType` and its usage patterns with respect to queries.

Let me back up a bit. Historically, that is, before JPA existed, and still today, Hibernate had/has four `FlushMode`s: 

- `NEVER`
- `COMMIT`
- `AUTO`
- `ALWAYS`

The settings `AUTO`, `COMMIT`, and `NEVER` make sense at the session level, in order to control the cost of dirty checking.

On the other hand, `ALWAYS` and `NEVER` make sense at the query level, to either:

- force a flush before a native query, or
- skip dirty checking before a given query.

When we wrote the JPA spec, Mike didn't think that `NEVER` or `ALWAYS` were necessary, and I didn't object, and so JPA wound up with two flush modes: `COMMIT` and `AUTO`. Neither of these make a whole lot of sense as arguments to `Query.setFlushMode()`, though you can at least use:

```java
query.setFlushMode(COMMIT)
```

to suppress flushing before a given query.

And, if you need to force a flush before execution of a native query, you can just call `flush()`.

But I would like to be able to specify a flush mode for, e.g. a `@NamedNativeQuery` or a `@NativeReaQuery` as part of the annotation, and unfortunately the `FlushModeType` just doesn't work for that.

We recently added a new `QueryFlushMode` enum to Hibernate:

```java
/**
 * Enumerates the possible flush modes for execution of a
 * {@link org.hibernate.query.Query}. An explicitly-specified
 * {@linkplain Query#setQueryFlushMode(QueryFlushMode)
 * query-level flush mode} overrides the current
 * {@linkplain org.hibernate.Session#getHibernateFlushMode()
 * flush mode of the session}.
 *
 * ...
 */
public enum QueryFlushMode {
	/**
	 * Flush before executing the query.
	 */
	FLUSH,
	/**
	 * Do not flush before executing the query.
	 */
	NO_FLUSH,
	/**
	 * Let the owning {@linkplain org.hibernate.Session session}
	 * decide whether to flush, depending on its current
	 * {@link org.hibernate.FlushMode}.
	 */
	DEFAULT
}
```

I think it would be nice to add something very similar to the Persistence spec, after we reach a resolution on #780.

Note that this is a lower-priority issue than other things we're working on right now, so I'm just parking it here.

gavinking:
[...]

PR #956, opened 2026-02-20
introduce QueryFlushMode and FlushModeType.EXPLICIT

gavinking:
See #787.

### DIFF
--- a/api/src/main/java/jakarta/persistence/FlushModeType.java
+++ b/api/src/main/java/jakarta/persistence/FlushModeType.java
@@ -19,33 +19,36 @@ package jakarta.persistence;
 
 /**
  * Enumerates flush modes recognized by the {@link EntityManager}.
+ * The {@linkplain EntityManager#getFlushMode() flush mode of a
+ * persistence context} controls synchronization of modifications
+ * to managed entities with the database.
  *
- * <p>When a Jakarta Persistence or native SQL query is executed
- * within a transaction via an instance of {@link Query} obtained
- * from an {@link EntityManager} joined to the transaction, the
- * effective flush mode is determined by the current
- * {@linkplain Query#getFlushMode flush mode of the query object},
- * which defaults to the current
- * {@linkplain EntityManager#getFlushMode flush mode of the
- * persistence context}.
+ * <p>When a transaction is active and the persistence context is
+ * joined to the transaction, the persistence provider runtime is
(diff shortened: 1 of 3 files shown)


gavinking:
Alright then, I say let's do this. 

My strongest motivation for wanting this is to be able to control flushing with stateful Jakarta Data repositories ... cc @njr-11.

gavinking:
Ooooops, I just realized I messed up the spec of `QueryFlushMode.FLUSH`. It's supposed to *force* a `FLUSH`. I didn't really intend for it to be equivalent to `AUTO`. I have fixed that now.

gavinking:
I have done some more work on the specification of this stuff.

I believe this is now ready to merge.

gavinking:
> This will be nice to have on `@ReadQueryOptions` and `@WriteQueryOptions` although I worry that having 2 very similarly named enumerations `FlushModeType` and the new `QueryFlushMode` for `Query` methods get/setFlushMode and getSetQueryFlushMode will be confusing to users.

Well that's why I'm deprecating the older method, which was already confusing, since `COMMIT` meant "don't flush" which is super unintuitive.

> I was trying to think of some alternative approaches such as expanding the existing `FlushModeType` enum, but that opens up the possibility of some other values being supplied to `@ReadQueryOptions` and `@WriteQueryOptions`. I was also trying to think of whether there was any way for `@ReadQueryOptions` and `@WriteQueryOptions` to accept a boolean instead, but it doesn't look like there is any really clean solution to this.

I maybe would have preferred the `boolean` too, but it runs into the same problem you pointed out with cache modes: there's no "null" meaning "do whatever the EM specifies".

api/src/main/java/jakarta/persistence/FlushModeType.java
```diff
@@ -56,31 +58,44 @@
  * @since 1.0
```
sebersole, line 62: Perhaps add a note here mentioning `QueryFlushMode`?

```suggestion
 * @apiNote For query execution, the flush mode defined on the EntityManager may be overridden by a {@linkplain QueryFlushMode} declared on the query.
 * 
 * @since 1.0
```

gavinking: @sebersole I have expended the Javadoc of both enums to address this.

gavinking:
After some Javdoc improvements, I'm merging this one. Thanks for the reviews!

Issue #332, opened 2021-11-15
Define default AccessType for XML configuration


dazey3:
Section 2.3.1 (Default Access Type) goes into great detail to define the default access type for mapping annotations. However, it does not define a default access type strategy for XML descriptor configurations.
This causes issues for persistence providers as they are open to default to whatever strategy they want for XML configurations and overall reduces portability.

Possible fixes:

1) We pick one. Define either AccessType.FIELD or AccessType.PROPERTY as the default type for XML configuration.

2) Specifically state that `access` attribute is required in orm.xml schema.

Current orm.xml schema fragment:
```xml
    <xsd:complexType name="entity">
        <xsd:sequence>
            ...
            <xsd:attribute name="class" type="xsd:string" use="required"/>
            <xsd:attribute name="access" type="orm:access-type"/>
```
This seems pretty clean, but creates issues with backwards compatibility. If the spec can't define a default access type (because there is no annotation placement to use), this would force users to make that determination.

3) Just say that the behavior is provider dependent and address portability. This has precedent as the spec already addresses portability at several places. This would be the easiest solution as well if no one agrees we should make any changes.

dazey3:
One other thing to note is that Section 2.3.2 does contain the following, but I don't think it is very clear what it means:

```
It is an error if a default access type cannot be determined and an access type is not explicitly specified by means of annotations or the XML descriptor.
```

Who is in error? Is this saying that # 2 is correct; that an error should be thrown by the provider for XML configurations because `'a default type cannot be determined'` and `'an access tyoe is not explicitly specified'`? Perhaps this could be cleaned up to make its point clearer

lukasj:
2. is not an option - access in an orm.xml can be defined on multiple places - there can be a `persistence-unit-default`, a global setting or per-entity/embeddable/... setting - which one should be the one required to be set? What impact would such setting have on the overriding rules?

I think that portable behavior in this case is covered by the last paragraph of 2.3.1:

_It is an error if a default access type cannot be determined and an access type is not explicitly specified by means of annotations or the XML descriptor. The behavior of applications that mix the placement of annotations on fields and properties within an entity hierarchy without explicitly specifying the Access annotation is undefined._

which, IMHO, clearly says, that the only correct (=portable) behavior in case of missing access type strategy is to fail - access type is not explicitly specified by means of annotations or the XML descriptor => it cannot be determined.

dazey3:
@lukasj 

[...]

PR #957, opened 2026-02-20
define a default <access> for XML mapped entities

gavinking:
see #332


gavinking:
@sebersole How do we feel about this?

sebersole:
In general, I think FIELD makes sense.  

However, this is a change in behavior for Hibernate users.  So from that perspective I'm obviously a little concerned.

gavinking:
> So from that perspective I'm obviously a little concerned.

Well we don't have to do this. I guess it's your call.

PR #958, opened 2026-02-20
Add spec + examples of @MapsId with @IdClass

### LINKED ISSUE #526: Allow use of @MapsId with @IdClass
After a close read of the spec, I was quite surprised to realize that the spec doesn't bless the following very natural thing:

```java
@Entity
@IdClass(DerivedId.class)
class Child {
    @Id
    Long parentId;
    
    @Id 
    String childId;

    @ManyToOne
    @MapsId(Child_.PARENT_ID)
    @JoinColumn(name="parentId")
    Parent parent;
}
```

```java
class DerivedId {
    Long parentId;
    String childId;
}
```

In fact, the spec implies that `@MapsId` can't be used with `@IdClass`, which seems like a pretty annoying limitation, forcing me to rename the `parentId` property of `DerivedId` to `parent`.

gavinking:
see #526

### DIFF
--- a/api/src/main/java/jakarta/persistence/MapsId.java
+++ b/api/src/main/java/jakarta/persistence/MapsId.java
@@ -67,7 +67,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  *     @EmbeddedId
  *     PaycheckId id;
  *
- *     @MapsId("empId")  // maps the empId attribute of DependentId
+ *     @MapsId("empId")  // maps the empId attribute of PaycheckId
  *     @ManyToOne
  *     Employee emp;
  *
@@ -80,7 +80,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * {@snippet :
  * class PaycheckId {
  *     int period;
- *     long empId;  // agrees with empId field of Dependent
+ *     long empId;  // agrees with empId field of Paycheck
  * }
  *
  * @Entity
--- a/api/src/main/java/jakarta/persistence/MapsId.java
+++ b/api/src/main/java/jakarta/persistence/MapsId.java
@@ -37,10 +37,13 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * <p>In this example, the parent entity has simple primary key:
  * {@snippet :
  * @Entity
- * public class Employee {
+ * class Employee {
+ *
  *     @Id
- *     long empId;
+ *     long id;
+ *
  *     String name;
+ *
  *     ...
  * }
  * }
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/persistence/MapsId.java
```diff
@@ -49,29 +56,58 @@
  * declare its composite primary key:
  * {@snippet :
  * @Embeddable
- * public class DependentId {
+ * class DependentId {
  *     String name;
```
njr-11: The example will be easier to understand if he have if the dependent class is named to be something that has a ManyToOne relationship with an `Employee`.  Here is one idea,

```suggestion
 * class PaycheckId {
 *     LocalDate dateOfPayment;
```

api/src/main/java/jakarta/persistence/MapsId.java
```diff
@@ -49,29 +56,58 @@
...
  * @Entity
- * public class Dependent {
+ * class Dependent {
+ *
  *     @EmbeddedId
  *     DependentId id;
+ *
+ *     @MapsId("empId")  // maps the empId attribute of DependentId
+ *     @ManyToOne
+ *     Employee emp;
```
njr-11: Corresponds to prior suggestion,
```suggestion
 * class Paycheck {
 *
 *     @EmbeddedId
 *     PaycheckId id;
 *
 *     @MapsId("empId")  // maps the empId attribute of PaycheckId
 *     @ManyToOne
 *     Employee emp;
```

api/src/main/java/jakarta/persistence/MapsId.java
```diff
@@ -49,29 +56,58 @@
...
+ *
+ *     @Id
+ *     String name;
+ *
+ *     @Id
+ *     long empId;  // agrees with primary key type of Employee
+ *
+ *     @MapsId("empId")  // maps the empId attribute
  *     @ManyToOne
  *     Employee emp;
```
njr-11:
```suggestion
 * class PaycheckId {
 *     LocalDate dateOfPayment;
 *     long empId;  // agrees with empId field of Paycheck
 * }
 *
 * @Entity
 * @IdClass(PaycheckId.class)
 * class Paycheck {
 *
 *     @Id
 *     LocalDate dateOfPayment;
 *
 *     @Id
 *     long empId;  // agrees with primary key type of Employee
 *
 *     @MapsId("empId")  // maps the empId attribute
 *     @ManyToOne
 *     Employee emp;
```

api/src/main/java/jakarta/persistence/MapsId.java
gavinking: I think that's a good change, but we will need to make it to all the existing examples at the same time.

gavinking:
@njr-11 done, WDYT?

njr-11 (approved):

gavinking:
@sebersole WDYT of this?

sebersole (approved):

gavinking:
This was already approved by Nathan and Steve, and I just further discussed it with @sebersole and we agree that this one should be merged, if only because `@MapsId` is kinda incomplete without it.

PR #959, opened 2026-02-21
introduce FetchMapping

### LINKED ISSUE #928: New MemberMapping
I know @gavinking is not super keen on unsealing `ResultSetMapping` - #887.  

As mentioned there, there are a few things in Hibernate's legacy dynamic ResultSet mapping API for native SQL that we cannot express in `ResultSetMapping`.  The biggest one is the notion of mapping a join fetch (to-one).  

As an alternative to #887, how about adding the ability to map fetched to-one associations -

```java
public sealed interface MemberMapping<T>
        permits FieldMapping, EmbeddedMapping, ToOneMapping {
}

public record ToOneMapping<T>
        (Class<C> containerType, String alias,
        String name, Class<T> toOneType,
        ColumnMapping[] fkColumns,
        ...) implements MemberMapping<T> {
}
```

E.g.

```java
var sqlString = """
        select ...,
            b.title,
            p.name,
            b.publisher_fk,
            b.id,
            p.id
        from books b
            join publishers p on b.publisher_fk = p.id
        """;

var publisher = toOne(Book_.publisher,
        "p",
        members(
                field("name", ... ),
                ...
        )
);
var bookMapping = entity(Book.class,
        "b",
        members(
                field("title" ),
                publisher,
                ...
        )
);

em.createNativeQuery( sqlString, bookMapping );
```

gavinking:
Fixes #928.

Looking for feedback.


sebersole:
So just don't support fetching element-collection?

gavinking:
> So just don't support fetching element-collection?

Correct. To add support for `@ElementCollection` we would need to introduce a new supertype of `EntityMapping`, `ColumnMapping`, and maybe `EmbeddableMapping`. I have experimented with that locally and ... well, it works.

PR #962, opened 2026-02-22
correct handling of inheritance in EntityMapping constructor methods

gavinking:
Entities in result set mappings can have:

1. attributes inheritance from a suprtype (this also applies to embeddables)
2. attributes declared by entity subtypes, which need mapping, because queries are polymorphic

This change allows that, without loss of typesafety in the `entity()` constructor method.

### DIFF
--- a/api/src/main/java/jakarta/persistence/sql/EmbeddedMapping.java
+++ b/api/src/main/java/jakarta/persistence/sql/EmbeddedMapping.java
@@ -34,10 +34,10 @@ import static java.util.Objects.requireNonNull;
  * @since 4.0
  */
 public record EmbeddedMapping<C,T>
-        (Class<C> container, Class<T> embeddableClass, String name, MemberMapping<?>[] fields)
+        (Class<? super C> container, Class<T> embeddableClass, String name, MemberMapping<T>[] fields)
         implements MemberMapping<C> {
 
-    public EmbeddedMapping(Class<C> container, Class<T> embeddableClass, String name, MemberMapping<?>[] fields) {
+    public EmbeddedMapping(Class<? super C> container, Class<T> embeddableClass, String name, MemberMapping<T>[] fields) {
         requireNonNull(container, "container is required");
         requireNonNull(embeddableClass, "embeddableClass is required");
         requireNonNull(name, "name is required");
@@ -52,7 +52,7 @@ public record EmbeddedMapping<C,T>
     }
 
     @Override
-    public MemberMapping<?>[] fields() {
+    public MemberMapping<T>[] fields() {
         return fields.clone();
     }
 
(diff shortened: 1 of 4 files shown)


api/src/main/java/jakarta/persistence/sql/ResultSetMapping.java
```diff
@@ -142,18 +142,33 @@ static <T> EntityMapping<T> entity(Class<T> entityClass, MemberMapping<T>... fie
...
+     * Construct a mapping for an entity class.
+     *
+     * @param entityClass The Java class of the entity
+     * @param lockMode The lock mode acquired by SQL query
+     * @param fields Mappings for fields or properties of the entity
+     *
+     * @see jakarta.persistence.EntityResult
+     */
+    @SafeVarargs
+    static <T> EntityMapping<T> entity(Class<T> entityClass, LockModeType lockMode, MemberMapping<T>... fields) {
```
marko-bekhta, line 155: hmm so we offer the `MemberMapping<? extends T>... fields` option if there's a discriminator column,  what if the table per class is used as inheritance strategy ? should we just go with `MemberMapping<? extends T>... fields` for all `entity(...)` cases ?

api/src/main/java/jakarta/persistence/sql/EmbeddedMapping.java
```diff
@@ -52,7 +52,7 @@ public EmbeddedMapping(Class<C> container, Class<T> embeddableClass, String name
     }
 
     @Override
-    public MemberMapping<?>[] fields() {
+    public MemberMapping<T>[] fields() {
         return fields.clone();
```
marko-bekhta, line 56: not related to the changes in this PR, I was just wondering how "expensive" this clone could get in the app ?

maybe we should go with the `List<MemberMapping<T>>` and something like
```java
this.fields = List.copyOf(fields); / this.fields = List.of(fields);
```
instead ? (so that we don't clone on get)

api/src/main/java/jakarta/persistence/sql/ResultSetMapping.java
gavinking: There are no "mapping strategies" for native SQL queries. For any polymorphic query there's always a discriminator column in the SQL result set. There has to be, how else are you going to distinguish what class a row belongs to.

api/src/main/java/jakarta/persistence/sql/EmbeddedMapping.java
gavinking: I highly doubt that cloning a very small array is expensive.

And. the problem with your suggestion is that `List.of()` returns an object which is not type safe at all. Half its methods throw `UnsupportedOperationException`. This is a really super-sucky mistake in the design of the Java collections API that for some reason has never been fixed.

api/src/main/java/jakarta/persistence/sql/ResultSetMapping.java
marko-bekhta: ah right, SQL !!! yes makes sense. 
I should probably turn off the 💻 🙂🤦🏻‍♂️(or think more before saying something 🙂)

api/src/main/java/jakarta/persistence/sql/EmbeddedMapping.java
marko-bekhta: got it 🙂 👍🏻

gavinking:
Going to merge this since it's more of a bug fix than anything.

gavinking:
Thanks for the review @marko-bekhta

PR #963, opened 2026-02-23
One more batch of TCK updates

marko-bekhta:
I thought I got all of them in the previous PR .... apparently I was wrong 😔 

- There was one more test relying on one-to-many mapping but treating it as if many-to-many when preparing test data (adjusted the test to not reuse the values in onetomany collections )
- One more test was using the same table names for multiple entities with different mapping (this time it was the `MapKeyEnumerated`), so just used different table names there
- apparently there are some "simple" tests on various Jakarta Persistence enums ... like check that `Something.values()` has a certain amount of values in there ... or if `valueOf` works ... I find those tests a bit redundant... but since they were there already, I just updated a couple cases where enums got a new constant so that the test would pass again... (but honestly I'd be in favor of just dropping such tests from the TCK, there's no provider involvement in them)
- I've disabled two tests where the expectation was that the query must fail ... since I could not see a reason for it to fail ...


tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/entityManager2/Client1.java
marko-bekhta: I'm not sure why we'd expect this query to fail (either on create or execute ..)
`DoesNotExist` is an entity added to the "current PU" so .. seems like it should just pass (which btw it does)

gavinking:
> apparently there are some "simple" tests on various Jakarta Persistence enums ... like check that Something.values() has a certain amount of values in there ... or if valueOf works ... I find those tests a bit redundant...  there's no provider involvement in them

Yes, *please* delete any such tests. We don't need tests for Java language features.

tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/core/entityManager2/Client1.java
gavinking: Perhaps the original idea was that the table `DOESNOTEXIST` does not exist in the database.

But because we're using schema export it does now.

marko-bekhta:
> Yes, please delete any such tests. We don't need tests for Java language features.

🙂 ok these are removed

> Perhaps the original idea was that the table DOESNOTEXIST does not exist in the database.

tried to update those two tests for update/delete queries I've disabled initially to let the update fail because of the length constraint on the column and with delete attempting to remove an entity that is still referenced from the other one. With those changes we now get the runtime exceptions again and the tests pass 🤞🏻 🙂

gavinking:
So am I pulling the trigger on this now?

marko-bekhta:
I think .. you can yes 🙂 

from my local run I'm only getting a few failures for validation tests (we need to make changes in Hibernate ORM) and one other test related to `java.sql.Timestamp` and friends, which seems to be a problem on the ORM side too

gavinking:
Thanks man.

PR #965, opened 2026-02-24
Use entityHandler instead of session in query javadoc examples

marko-bekhta:
I noticed that a few examples were using `session`, while I like that 😃, I wondered whether we should stick to em/entityhandler and so on for these ?

feel free to just close it if you think it's fine as it currently is 🙂

### DIFF
--- a/api/src/main/java/jakarta/persistence/Query.java
+++ b/api/src/main/java/jakarta/persistence/Query.java
@@ -479,7 +479,7 @@ public interface Query {
      * }
      * {@snippet :
      * var books =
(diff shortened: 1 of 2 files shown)


gavinking:
Haha, so `session` is definitely wrong there, oops!

But I would prefer to go with `entityAgent` or `entityManager` rather than `entityHandler`, because `EntityHandler` is a rather abstract type.

marko-bekhta:
changed it to `em` to match the other examples and to generate less changes 🙂

gavinking:
Thanks Marko!

PR #966, opened 2026-02-25
tuples and aliases

gavinking:
Rebase of #889.


gavinking:
OK, as per discussion on #889, I believe we have agreement on this one, and leaving it unmerged has started to generate conflicts with other work. Merging now.

PR #967, opened 2026-02-25
modernize impl of Persistence + helper class

gavinking:


gavinking:
This is just a cleanup to implementation details, to hopefully make the code a bit more understandable. Merging.

PR #968, opened 2026-02-26
simpler result mappings for @StaticNativeQuery

### LINKED ISSUE #961: simpler result mappings for @StaticNativeQuery
Currently, we say you can do this:

```java
@StaticNativeQuery("select * from books where isbn = ?")
@SqlResultSetMapping(entities =
        @EntityResult(entityClass = Book.class,
                fields = {@FieldResult(name = "isbn", column = "isbn_13"),
                          @FieldResult(name = "title", column = "title_en")}))
Book getBookWithIsbn(String isbn);
```

I think that's just unnecessary ceremony, when we can easily just let you write this:

```java
@StaticNativeQuery("select * from books where isbn = ?")
@EntityResult(entityClass = Book.class,
        fields = {@FieldResult(name = "isbn", column = "isbn_13"),
                  @FieldResult(name = "title", column = "title_en")})
Book getBookWithIsbn(String isbn);
```

All we need to do is promote the `XxxResult` annotations to `METHOD` and make them `@Repeatable`.

gavinking:
- roll back some changes to @SqlResultSetMapping
- allow @XxxxResult at the method level for a @StaticNativeQuery

see #961

### DIFF
--- a/api/src/main/java/jakarta/persistence/ConstructorResult.java
+++ b/api/src/main/java/jakarta/persistence/ConstructorResult.java
@@ -17,8 +17,12 @@
 
 package jakarta.persistence;
 
+import java.lang.annotation.Annotation;
+import java.lang.annotation.Repeatable;
 import java.lang.annotation.Target;
 import java.lang.annotation.Retention;
+
+import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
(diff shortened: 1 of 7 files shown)


gavinking:
@njr-11 Please review this, this mostly impacts Jakarta Data.

api/src/main/java/jakarta/persistence/ConstructorResult.java
njr-11: This isn't part of your changes, but I noticed a contradiction in the Javadoc for `columns()` and `entities()` regarding order.  Both of them currently say their corresponding constructor arguments must be ordered after the other.  I'm not sure which order you wanted or if it really matters, but the following would resolve it by putting column arguments first, then entity arguments,

```suggestion
    * <p>Constructor parameters mapping directly to column
    * results must occur before parameters mapping to entities.
```

njr-11, line 134: Alternatively, the following could be switched to resolve the conflict to result in the opposite order,
```suggestion
    * <p>Constructor parameters mapping to entity results must
    * occur before parameters mapping to columns.
```
**IMPORTANT** - choose between this or the previous suggestion. Accepting both would cause the conflict to remain.

njr-11 (approved):
This is a nice simplification.  While reviewing I noticed an inconsistency between two statements in Javadoc that are not part of your changes but which I noticed in nearby code right before.  I included suggestions in both places (we will need to pick one between the two how to resolve).  I also approved the PR in case it's better to handle the unrelated comment as a separate issue.

api/src/main/java/jakarta/persistence/ConstructorResult.java
gavinking: Thanks, nice catch. I applied this change.

gavinking:
I'm going to go ahead with this one and include it in M2, because it seems to me a significantly simpler solution on both the spec side, and for the user, than what we had previously.

If there are any problems lurking, we'll find out about them when we come to implement this and write the TCK tests.

PR #969, opened 2026-02-26
spec that changes to owned collections + associations trigger @XxxUpdate

### LINKED ISSUE #167: Entity listeners are not fired when a collection has been updated
I'm having a `Product` entity, which has a `manufacturers` attribute:

````
@ElementCollection(targetClass = java.lang.String.class, fetch = FetchType.EAGER)
--
@CollectionTable(name = "manufacturers", joinColumns = @JoinColumn(name = "man_id"), indexes = { @Index(columnList = "man_id") })
@Column(name = "manufacturer")
private Collection<String> manufacturers;
````

I also have a `ProductListener` with two `@PreUpdate` and `@PrePersist` methods. However none of these methods are triggered when I update the collection of manufacturers. Either we need to trigger the event when the collection is updated, or add some new annotations like `@PreCollectionUpdate` and `@PreCollectionPersist` (hibernate style).

### PR DESCRIPTION
**Pull request by gavinking**
see #167

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -1817,6 +1817,25 @@ and are not tied to the lifecycle of managed entities associated with a
 persistence context. These callbacks are not triggered when entity data is
 modified by execution of a `Query`.
 
+In this context, the _entity data_ belonging to an entity instance is
+considered to include:
+
+- every basic or embedded attribute of the entity instance, recursively,
+- every `@ElementCollection` belonging to the entity instance or to an
+  embeddable object belonging to the instance,
+- every `@ManyToOne` association field or property of the entity instance
+  or of an embeddable object belonging to the instance (but not the state
+  of the associated target entity),
+- every _owned_ `@OneToOne` association field or property of the entity
+  instance or of an embeddable object belonging to the instance, and
+- every _owned_ `@OneToMany` or `@ManyToMany` association belonging to the
+  entity instance or to an embeddable object belonging to the instance.
+
+Therefore, changes to the persistent state of owned associations result in
+invocation of the `@PreUpdate` and `@PostUpdate` callbacks. Changes which
+modify only unowned associations never result in `@PreUpdate` or `@PostUpdate`
+callbacks.
+
 [NOTE]
 ====
 It is implementation-dependent as to whether `@PreUpdate` and `@PostUpdate`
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by gavinking** [comment]
@sebersole We've talked a lot about this one. Your call whether we do this at last.

**Comment by sebersole** [comment]
Hibernate has a mush richer API for collection change events.  But this seems like a simple thing to do, for us, I guess from our collection event framework.  +0 for me

File: spec/src/main/asciidoc/ch03-entity-operations.adoc [hunk 1]
```diff
@@ -1817,6 +1817,25 @@ and are not tied to the lifecycle of managed entities associated with a
 persistence context. These callbacks are not triggered when entity data is
 modified by execution of a `Query`.
 
+In this context, the _entity data_ belonging to an entity instance is
+considered to include:
+
+- every basic or embedded attribute of the entity instance, recursively,
+- every `@ElementCollection` belonging to the entity instance or to an
+  embeddable object belonging to the instance,
```
yrodiere, line 1825: I don't want to be that guy, but you seem to be aiming for an extra-precise definition here, so:

This is not just the collection, but also the attributes of elements of that collection, in the case of an `@ElementCollection` of `@Embeddable`s. Right?

So maybe:

```suggestion
- every basic or embedded attribute of the entity instance, recursively,
- every `@ElementCollection` belonging to the entity instance or to an
  embeddable object belonging to the instance,
- every basic or embedded attribute within such an `@ElementCollection`
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1817,6 +1817,25 @@ and are not tied to the lifecycle of managed entities associated with a
...
+  of the associated target entity),
+- every _owned_ `@OneToOne` association field or property of the entity
+  instance or of an embeddable object belonging to the instance, and
+- every _owned_ `@OneToMany` or `@ManyToMany` association belonging to the
+  entity instance or to an embeddable object belonging to the instance.
+
+Therefore, changes to the persistent state of owned associations result in
+invocation of the `@PreUpdate` and `@PostUpdate` callbacks. Changes which
+modify only unowned associations never result in `@PreUpdate` or `@PostUpdate`
+callbacks.
```
yrodiere, line 1840: I assume that's covered elsewhere in the spec, but should you also stress that this also implies that changing a collection leads to a version bump for the corresponding entity?

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: OK, I have made that change.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: It is indeed already covered in the section on Optimistic Locking. Not quite in the laborious detail I'm stating it here, but I think the intention is clear enough.

gavinking:
I'm going to merge this so that it makes it into the m2 release.

PR #973, opened 2026-02-27
add mappedById to @ManyToOne and @OneToOne

### LINKED ISSUE #527: add mappedBy to @ManyToOne and @OneToOne
The couple of weeks I've been thinking about the following sort of code, which is already possible in JPA:

```java
@Entity
class Child {
     @Id 
     Long id;

     Long parentId;

     @ManyToOne 
     @Column(name="parentId", 
             insertable=false, updatable=false)
     Parent parent;

     ...
}
```

Here the `parentId` field and the `parent` association are mapped to the same column.

It seems to me that this sort of thing is not uncommon, but unfortunately that `@Column` annotation is quite clumsy.

We could make it quite a lot more elegant by allowing the following:

```java
@Entity
class Child {
     @Id 
     Long id;

     Long parentId;

     @ManyToOne(mappedBy=Child.PARENT_ID)
     Parent parent;

     ...
}
```

This is quite analogous to the use of `mappedBy` in `@OneToMany` and `@ManyToMany`, with the difference that here `mappedBy` points to a different field of the current entity.

An argument against this features is that it overlaps with `@MapsId`, and that's true, I can use it this way:

```java
@Entity
@IdClass(ChildId.class)
class Child {
     @Id 
     Long id;

     @Id
     Long parentId;

     @ManyToOne(mappedBy=Child.PARENT_ID)
     Parent parent;

     ...
}
```

or this way:

```java
@Entity
class Child {
     @Id 
     Long id;

     @OneToOne(mappedBy=Child.ID)
     Parent parent;

     ...
}
```

to do things that are already possible with `@MapsId`.

On the other hand, I find `@MapsId` to be quite confusing in general, and sort-of backwards compared to my intuition. So I actually _prefer_ the above code to the equivalent thing with `@MapsId`.

Reactions?

gavinking:
For consideration. Competes with #958.

see #527

### DIFF
--- a/api/src/main/java/jakarta/persistence/ManyToMany.java
+++ b/api/src/main/java/jakarta/persistence/ManyToMany.java
@@ -140,9 +140,16 @@ public @interface ManyToMany {
      */
     FetchType fetch() default FetchType.LAZY;
 
-    /** 
-     * The field that owns the relationship. Required unless 
-     * the relationship is unidirectional.
+    /**
+     * (Optional) The field or property that owns the relationship.
+     * This element is only specified on the inverse (non-owning)
+     * side of the association. Modifications to the owning side of
+     * a relationship determine the updates made to the relationship
+     * in the database. If the inverse side of an association is
+     * modified without a corresponding modification to the owning
+     * side, the behavior is undefined. The persistence provider is
+     * permitted to ignore any modification made only to the inverse
+     * side of a bidirectional association.
      */
     String mappedBy() default "";
 }
--- a/api/src/main/java/jakarta/persistence/OneToOne.java
+++ b/api/src/main/java/jakarta/persistence/OneToOne.java
@@ -213,7 +213,7 @@ public @interface OneToOne {
      *           foreignKey = @ForeignKey(name="child_parent_fk"))
      *     Long id;
      *
-     *     @ManyToOne(mappedById = Child_.ID)
+     *     @OneToOne(mappedById = Child_.ID)
      *     Parent parent;
      * }
      * }
--- a/api/src/main/java/jakarta/persistence/OneToOne.java
+++ b/api/src/main/java/jakarta/persistence/OneToOne.java
@@ -183,7 +183,6 @@ public @interface OneToOne {
      */
     String mappedBy() default "";
 
-
     /**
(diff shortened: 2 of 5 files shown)


gavinking:
Thoughts, @sebersole? @njr-11?

api/src/main/java/jakarta/persistence/OneToOne.java
```diff
@@ -183,12 +183,41 @@
...
+     * <p>
+     * {@snippet :
+     * @Entity
+     * class Child {
+     *     @Id
+     *     @JoinColumn(name = "parent_id",
+     *           foreignKey = @ForeignKey(name="child_parent_fk"))
+     *     Long id;
+     *
+     *     @ManyToOne(mappedById = Child_.ID)
```
njr-11: because this example is on the `OneToOne` annotation,
```suggestion
     *     @OneToOne(mappedById = Child_.ID)
```

api/src/main/java/jakarta/persistence/ManyToMany.java
```diff
@@ -140,9 +140,16 @@
      */
     FetchType fetch() default FetchType.LAZY;
 
-    /** 
-     * The field that owns the relationship. Required unless 
-     * the relationship is unidirectional.
+    /**
```
sebersole, line 143: mappedBy serves 2 purposes and I think it makes sense to discuss both of them here.  This is a great discussion about the runtime implication in terms of ownership.  

But it also defines semantics of the foreign-key - look "there" for it.  And in the case of one-to-one actually defines the key side versus the target side of the foreign-key.  That's perhaps obvious to us by the term "inverse" but is that generally understood?  Not sure.

Same comment for mapedBy on `@OneToMany` and `@OneToOne`.

sebersole (commented):
I like it.  Left a comment about changes to the javadoc for mappedBy for consideration, but otherwise +1

api/src/main/java/jakarta/persistence/ManyToMany.java
gavinking: Yup, OK.

gavinking:
@sebersole and I discussed this today, and decided to put it on ice for now. We might reconsider it for 4.1 or whatever.

PR #976, opened 2026-03-05
make @NamedEntityGraphXxxx inner annotations of @NamedEntityGraph

### LINKED ISSUE #975: Make @NamedEntityGraphAttributeNode shorter
`@NamedEntityGraphAttributeNode` is a mouthful, especially if it's going to be sprinkled across attributes of an entity class.

Could we perhaps shorten it?

One idea would be to turn it into `@NamedEntityGraph.AttributeNode`, so that users can, if they so choose, add an `import static` and use `@AttributeNode` instead.

gavinking:
as suggested by @yrodiere


gavinking:
@sebersole @njr-11 @anija-anil any objections to this change?

njr-11 (approved):
Looks good.

anija-anil (approved):

sebersole:
Not necessarily against it, but this was introduced in JPA 2.1 - seems odd to change it now for no real user benefit...

+0

gavinking:
> this was introduced in JPA 2.1 

@sebersole No, these are the new annotations added in 4.0, not the old ones.

sebersole:
You're right.  Sorry, my brain was primed by NamedAttributeNiode being the first file.  

Still +0 from me

dmatej (approved):

gavinking:
Lots of approvals on this one, so let's just do it.

PR #979, opened 2026-03-09
Draft a section on Jakarta Validation constraints influence on DDL

### LINKED ISSUE #813: @NotNull, @NotEmpty, @NotBlank from Bean Validation
In #614 we said that `@Nonnull` should imply `optional=false`.

I guess I had sorta assumed that we already said something similar about the Bean Validation annotations, since back in the day, the interop with the persistence spec was one of the main reasons for creating that spec. (It grew out of Hibernate Validator.)

But in fact, it looks like we don't say much about these annotations.

1. Hibernate takes `@NotNull`, `@NotEmpty`, and `@NotBlank` as implying `optional=false`.

2. It also considers `@Min`, `@Max`, `@Size`, and `@Digits` when generating DDL column lengths.

It's probably worth mentioning at least something about this in the spec and javadoc.

marko-bekhta:
I've started on this new section and realized that I don't have a solution for validation mode that I really like ... 
hence this is a draft and mostly to discuss who do we want to proceed 🙂

### DIFF
--- a/api/src/main/java/jakarta/persistence/Basic.java
+++ b/api/src/main/java/jakarta/persistence/Basic.java
@@ -97,7 +97,9 @@ public @interface Basic {
      * property may be null.
      *
      * <p>If the annotated field or property is of primitive type,
-     * or if it is also annotated {@code @jakarta.annotation.Nonnull},
+     * or if it is also annotated {@code @jakarta.annotation.Nonnull}
+     * or with one of the Jakarta Validation constraints influencing
+     * nullability (e.g. {@code @jakarta.validation.constraints.NotNull}),
      * then {@code optional=false} is implied, and the value of this
      * annotation member is ignored.
      *
--- a/api/src/main/java/jakarta/persistence/Column.java
+++ b/api/src/main/java/jakarta/persistence/Column.java
@@ -81,6 +81,12 @@ public @interface Column {
 
     /**
      * (Optional) Whether the database column is nullable.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * nullability (e.g. {@code @jakarta.validation.constraints.NotNull}),
+     * then {@code nullable=false} is implied, and the value of this
+     * annotation member is ignored.
      */
     boolean nullable() default true;
 
@@ -130,6 +136,12 @@ public @interface Column {
      * <p>
      * Applies only to columns whose type is parameterized by length,
      * for example, {@code varchar} or {@code varbinary} types.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * length (e.g. {@code @jakarta.validation.constraints.Size}),
+     * then {@code length=Size.max()} is implied, and the value of this
+     * annotation member is ignored.
      */
     int length() default 255;
 
@@ -141,6 +153,12 @@ public @interface Column {
      * <p>
      * The default value {@code 0} indicates that a provider-determined
      * precision should be inferred.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * precision (e.g. {@code @jakarta.validation.constraints.Digits}),
+     * then {@code precision=Digits.integer()+Digits.fraction()} is implied,
+     * and the value of this annotation member is ignored.
      */
     int precision() default 0;
 
@@ -152,6 +170,12 @@ public @interface Column {
      * <p>
      * The default value {@code 0} indicates that a provider-determined
      * scale should be inferred.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * scale (e.g. {@code @jakarta.validation.constraints.Digits}),
+     * then {@code scale=Digits.fraction()} is implied,
+     * and the value of this annotation member is ignored.
      */
     int scale() default 0;
 
--- a/api/src/main/java/jakarta/persistence/JoinColumn.java
+++ b/api/src/main/java/jakarta/persistence/JoinColumn.java
@@ -128,6 +128,12 @@ public @interface JoinColumn {
 
     /**
      * (Optional) Whether the foreign key column is nullable.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * nullability (e.g. {@code @jakarta.validation.constraints.NotNull}),
+     * then {@code nullable=false} is implied, and the value of this
+     * annotation member is ignored.
      */
     boolean nullable() default true;
 
--- a/api/src/main/java/jakarta/persistence/OrderColumn.java
+++ b/api/src/main/java/jakarta/persistence/OrderColumn.java
@@ -84,6 +84,12 @@ public @interface OrderColumn {
 
     /**
      * (Optional) Whether the database column is nullable.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * nullability (e.g. {@code @jakarta.validation.constraints.NotNull}),
+     * then {@code nullable=false} is implied, and the value of this
+     * annotation member is ignored.
      */
     boolean nullable() default true;
 
--- a/api/src/main/java/jakarta/persistence/ValidationMode.java
+++ b/api/src/main/java/jakarta/persistence/ValidationMode.java
@@ -37,17 +37,21 @@ package jakarta.persistence;
 public enum ValidationMode {
    
     /**
-     * If a Bean Validation provider is present in the environment,
+     * If a Jakarta Validation provider is present in the environment,
      * the persistence provider must perform the automatic validation
-     * of entities. If no Bean Validation provider is present in the
+     * of entities. If no Jakarta Validation provider is present in the
      * environment, no lifecycle event validation takes place.
      * This is the default behavior.
+     * <p>
+     * To ensure deterministic behavior and prevent silent validation omissions,
+     * applications should explicitly set the {@code validation-mode}
+     * to {@link ValidationMode#CALLBACK} rather than relying on the default.
      */
     AUTO,
 
     /**
      * The persistence provider must perform the lifecycle event
-     * validation. It is an error if there is no Bean Validation
+     * validation. It is an error if there is no Jakarta Validation
      * provider present in the environment.
      */
     CALLBACK,
--- a/spec/src/main/asciidoc/appA-revision-history.adoc
+++ b/spec/src/main/asciidoc/appA-revision-history.adoc
@@ -110,7 +110,7 @@ events: pre-persist, pre-merge, pre-remove, pre-delete.
 Added `@Discoverable` annotation
 
 Introduced the requirement to derive additional schema metadata from
-Jakarta Validation constraints.
+Jakarta Validation constraints
 
 ==== Minor Changes and Fixes
 
--- a/spec/src/main/asciidoc/appA-revision-history.adoc
+++ b/spec/src/main/asciidoc/appA-revision-history.adoc
@@ -109,6 +109,9 @@ events: pre-persist, pre-merge, pre-remove, pre-delete.
 
 Added `@Discoverable` annotation
 
+Introduced the requirement to derive additional schema metadata from
+Jakarta Validation constraints.
+
 ==== Minor Changes and Fixes
 
 `jakarta.annotation.Nonnull` now implies `optional=false`
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -2213,21 +2213,85 @@ it to be applied to that entity or mapped superclass and its subclasses.
 In the case of multiple callback methods for
 a single lifecycle event, the invocation order rules described in <<a2251>> apply.
 
-=== Jakarta Validation [[a2366]]
+=== Jakarta Validation integration [[a2366]]
 
-Jakarta Persistence integrates with Jakarta Validation <<a19498>>
-and specifies additional validation points for managed types
-(entities, mapped superclasses, and embeddable classes) at the
-entity lifecycle events described in <<a2202>>.
+Jakarta Persistence defines integration points with Jakarta Validation <<a19498>>.
+When the Jakarta Persistence provider is <<a2394,aware of the Jakarta Validation provider's presence>>,
+and <<a2374,the integration is enabled>>, the persistence provider must:
 
-When the Jakarta Persistence provider is
-<<a2394,aware of the Jakarta Validation provider's presence>>
-and <<a2374,enables automatic validation>>, it must automatically
-enforce the Jakarta Validation constraints <<a19498>> defined on
-managed objects. This process occurs internally within the persistence
-lifecycle (<<automatic-validation-lifecycle-events>>), ensuring
-validation without requiring additional application-side logic
-beyond the constraints' definitions.
+* <<automatic-validation-lifecycle-events,perform automatic validation upon lifecycle events>>:
+  The provider must automatically validate managed types (entities, mapped superclasses, and
+  embeddable classes) at the entity lifecycle events described in <<a2202>>. This process occurs
+  internally within the persistence lifecycle, ensuring validation without requiring explicit
+  application logic beyond the constraint definitions.
+
+* <<schema-generation-from-jakarta-validation-constraints,Derive schema metadata from Jakarta Validation constraints>>:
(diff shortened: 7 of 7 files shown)


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2155,21 +2155,86 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+`PersistenceProvider.createContainerEntityManagerFactory`, as defined in
+<<a13443>> and <<a12802>>.
+Expected values for this property are `auto`, `callback`, `ddl` and `none`.
+
+// TODO: How do we deal with DDL+CALLBACK ?
+//  - Do we break the API/XSD and make validation-mode a collection ?
+//  - Do we introduce another option CALLBACK_AND_DDL? or maybe call it `FULL` ?
+//  - Do we just say that CALLBACK == AUTO (but force Validator provider presence ?)
+//    i.e. we say that CALLBACK enables validation on events + influence the DDL ?
+//  - Something else ?
```
marko-bekhta: if we go with the definitions for the modes as
- ddl -- constraints influence the DDL, fail if validation provider is absent.
- callback -- trigger validation on life cycle events, fail if validation provider is absent.
- none -- don't enable any validation integrations.
- auto -- do ddl and callback if validation provider is available, none -- otherwise

then the initial idea with allowing users to pass multiple values to be able to get DDL+callback+fail-if-provider-not-there doesn't play nice with the rest, since it breaks a few methods in JPA as we'd need to start returning some collection instead of a single value + it breaks the XSD as we'd want to have those values in some list... 

The ideas I had so far on how to approach this: 
- Do we break the API/XSD and make `validation-mode` a collection ?
- Do we introduce another option CALLBACK_AND_DDL? or maybe call it `FULL` ?
- Do we just say that CALLBACK == AUTO (but force Validator provider presence ?)  i.e. we say that CALLBACK enables validation on events + influence the DDL ?

gavinking: The _simplest_ solution would be to have a total order for the modes (`NONE`, `DDL`, `CALLBACK`) So that either `CALLBACK` includes `DDL` or vice-versa.

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2290,24 +2325,130 @@ from a `jakarta.validation.ValidatorFactory` instance by calling
...
+This constraint is treated as a way to specify
+a check constraint on a numeric column.
+Influence on metadata:::
+None.
+Influence on DDL:::
+Adds a check constraint to the column, e.g.
+`CHECK (column_name \<= value)` if `inclusive`
+is `true` (default), or `CHECK (column_name < value)`
+otherwise.
+
```
njr-11, line 2488: Looking around at other constraints in Jakarta Validation, it seems like `@Positive`, `@PositiveOrZero`, `@Negative`, and `@NegativeOrZero` would also apply,

```suggestion

`jakarta.validation.constraints.Negative`::
This constraint is handled the same as `jakarta.validation.constraints.Max(-1)` for integral types and `jakarta.validation.constraints.DecimalMax(value = BigDecimal.ZERO, inclusive = false)` for decimal types.

`jakarta.validation.constraints.NegativeOrZero`::
This constraint is handled the same as `jakarta.validation.constraints.Max(0)` for integral types and `jakarta.validation.constraints.DecimalMax(value = BigDecimal.ZERO)` for decimal types.

`jakarta.validation.constraints.Positive`::
This constraint is handled the same as `jakarta.validation.constraints.Min(1)` for integral types and `jakarta.validation.constraints.DecimalMin(value = BigDecimal.ZERO, inclusive = false)` for decimal types.

`jakarta.validation.constraints.PositiveOrZero`::
This constraint is handled the same as `jakarta.validation.constraints.Min(0)` for integral types and `jakarta.validation.constraints.DecimalMin(value = BigDecimal.ZERO)` for decimal types.

```

api/src/main/java/jakarta/persistence/ValidationMode.java
```diff
@@ -25,21 +25,37 @@
...
+     * If a Jakarta Validation provider is present in the environment,
+     * the persistence provider must:
+     * <ul>
+     *     <li>perform the automatic validation of entities, and</li>
+     *     <li>must utilize Jakarta Validation constraint metadata
+     *     for the purpose of DDL schema generation.</li>
+     * </ul>
+     * If no Jakarta Validation provider is present in the
+     * environment, no lifecycle event validation takes place nor
+     * anything is derived from Jakarta Validation constraints regarding
```
njr-11:
```suggestion
     * is anything derived from Jakarta Validation constraints regarding
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: mm so I gave this some thought and ended up renaming the new  `DDL` option to `SCHEMA` and we'd have:
- `NONE` -- no integration
- `CALLBACK` -- only Jakarta Validation triggered upon lifecycle events.
- `SCHEMA` -- both automatic validation and schema checks 
- `AUTO` -- same as schema but don't fail if Jakarta Validation provider is not there.

My thinking is that with callback we are checking at the app level and call the validation provider, then with schema it's like a next layer, if we passed the validation checks by validation provider we "check at the DB level". WDYT?

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: thanks! added these constraints as well. 🙂 👍🏻

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: That sounds OK to me. Is `AUTO` really useful though?

marko-bekhta: mm I'd say yes, I thinking with `auto` being the default option, it's enough for the user to just add dependencies to get things working without any extra config... if we just have `none` `callback` `schema` then `none` seems like a "safe" default as there can be users without jakarta validation in their app and we wouldn't want to crash the app... 
And we already had it so there's that 🙂

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2290,24 +2319,172 @@ from a `jakarta.validation.ValidatorFactory` instance by calling
...
+Certain Jakarta Validation constraints can override
+the persistent attribute metadata and influence the
+schema generation.
+
+The following list describes how constraints affect
+the persistent attributes they are applied to:
+
+`jakarta.validation.constraints.NotNull`::
+This constraint is treated as a way to specify that
+the persistent attribute is not nullable.
```
```suggestion
the column mapped by the attribute is not nullable.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2290,24 +2319,172 @@ from a `jakarta.validation.ValidatorFactory` instance by calling
...
+Influence on metadata:::
+Overrides the `nullable` attribute and sets it to `false`.
+Influence on DDL:::
+Adds a `NOT NULL` constraint to the column.
+
+`jakarta.validation.constraints.NotBlank`::
+Similar to `jakarta.validation.constraints.NotNull`
+this constraint is treated as a way to specify that
+the persistent attribute is not nullable.
```
```suggestion
the column mapped by the attribute is not nullable.
```

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2290,24 +2319,172 @@ from a `jakarta.validation.ValidatorFactory` instance by calling
...
+attribute to `fraction`.
+Influence on DDL:::
+Defines the numeric type precision and scale,
+e.g. `DECIMAL(precision, scale)`.
+
+`jakarta.validation.constraints.Size`::
+If this constraint specifies the maximum value,
+e.g. `@Size(max = {someValue})`, then it is treated
+as a way to specify the length of the persistent
+attribute, i.e. `length={someValue}`.
```
```suggestion
as a way to specify the length of the column mapped 
by the attribute, i.e. `length={someValue}`.
```

gavinking (changes requested):
Need to clearly distinguish "persistent attribute" from the column it maps. DDL constraints apply to columns.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: > I thinking with auto being the default option, it's enough for the user to just add dependencies to get things working without any extra config... 

But it also implies that their code silently malfunctions if they later accidentally *omit* the dependency.

api/src/main/java/jakarta/persistence/Column.java
```diff
@@ -81,6 +81,12 @@
 
     /**
      * (Optional) Whether the database column is nullable.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing the
```
gavinking:
```suggestion
     * with one of the Jakarta Validation constraints influencing
```

api/src/main/java/jakarta/persistence/Basic.java
```diff
@@ -97,7 +97,9 @@
      * property may be null.
      *
      * <p>If the annotated field or property is of primitive type,
-     * or if it is also annotated {@code @jakarta.annotation.Nonnull},
+     * or if it is also annotated {@code @jakarta.annotation.Nonnull}
+     * or with one of the Jakarta Validation constraints influencing the
```
gavinking:
```suggestion
     * or with one of the Jakarta Validation constraints influencing
```

api/src/main/java/jakarta/persistence/Column.java
```diff
@@ -130,6 +136,12 @@
      * <p>
      * Applies only to columns whose type is parameterized by length,
      * for example, {@code varchar} or {@code varbinary} types.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * the length (e.g. {@code @jakarta.validation.constraints.Size}),
```
gavinking:
```suggestion
     * length (e.g. {@code @jakarta.validation.constraints.Size}),
```

api/src/main/java/jakarta/persistence/Column.java
```diff
@@ -141,6 +153,12 @@
      * <p>
      * The default value {@code 0} indicates that a provider-determined
      * precision should be inferred.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * the precision (e.g. {@code @jakarta.validation.constraints.Digits}),
```
gavinking:
```suggestion
     * precision (e.g. {@code @jakarta.validation.constraints.Digits}),
```

api/src/main/java/jakarta/persistence/Column.java
```diff
@@ -152,6 +170,12 @@
      * <p>
      * The default value {@code 0} indicates that a provider-determined
      * scale should be inferred.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing
+     * the scale (e.g. {@code @jakarta.validation.constraints.Digits}),
```
gavinking:
```suggestion
     * scale (e.g. {@code @jakarta.validation.constraints.Digits}),
```

api/src/main/java/jakarta/persistence/JoinColumn.java
```diff
@@ -128,6 +128,12 @@
 
     /**
      * (Optional) Whether the foreign key column is nullable.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing the
```
gavinking:
```suggestion
     * with one of the Jakarta Validation constraints influencing
```

api/src/main/java/jakarta/persistence/OrderColumn.java
```diff
@@ -84,6 +84,12 @@
 
     /**
      * (Optional) Whether the database column is nullable.
+     * <p>
+     * If the annotated field or property is also annotated
+     * with one of the Jakarta Validation constraints influencing the
```
gavinking:
```suggestion
     * with one of the Jakarta Validation constraints influencing
```

api/src/main/java/jakarta/persistence/ValidationMode.java
```diff
@@ -25,24 +25,54 @@
...
-     * the persistence provider must perform the automatic validation
-     * of entities. If no Bean Validation provider is present in the
-     * environment, no lifecycle event validation takes place.
+     * If a Jakarta Validation provider is present in the environment,
+     * the persistence provider must:
+     * <ul>
+     *     <li>perform the automatic validation of entities upon
+     *     lifecycle events, and</li>
+     *     <li>utilize Jakarta Validation constraint metadata
+     *     for the purpose of DDL schema generation.</li>
```
gavinking:
```suggestion
     * <li>perform the automatic validation of entities upon
     *     lifecycle events, and
     * <li>utilize Jakarta Validation constraint metadata
     *     for the purpose of DDL schema generation.
```

Believe it or not, no `</li>` tag is the standard in both HTML5 and in Oracle's standard conventions for Javadoc. (Also it's less noisy.)

api/src/main/java/jakarta/persistence/ValidationMode.java
```diff
@@ -25,24 +25,54 @@
...
     /**
-     * The persistence provider must perform the lifecycle event
-     * validation. It is an error if there is no Bean Validation
-     * provider present in the environment.
+     * The persistence provider must
+     * <ul>
+     *     <li>perform the automatic validation of entities upon
+     *     lifecycle events, and</li>
+     *     <li>utilize Jakarta Validation constraint metadata
+     *     for the purpose of DDL schema generation.</li>
```
gavinking:
```suggestion
     * <li>perform the automatic validation of entities upon
     *     lifecycle events, and
     * <li>utilize Jakarta Validation constraint metadata
     *     for the purpose of DDL schema generation.
```

api/src/main/java/jakarta/persistence/ValidationMode.java
```diff
@@ -25,24 +25,54 @@
...
 
     /**
-     * The persistence provider must not perform lifecycle event
-     * validation.
+     * The persistence provider must <b>not</b>
+     * <ul>
+     *     <li>perform the automatic validation of entities upon
+     *     lifecycle events, and</li>
+     *     <li>utilize Jakarta Validation constraint metadata
+     *     for the purpose of DDL schema generation.</li>
```
gavinking:
```suggestion
     * <li>perform the automatic validation of entities upon
     *     lifecycle events, and
     * <li>utilize Jakarta Validation constraint metadata
     *     for the purpose of DDL schema generation.
```

gavinking (changes requested):

api/src/main/java/jakarta/persistence/ValidationMode.java
marko-bekhta: huh 🙂 good to know 🙂 

well...  those lists are gone now, as I've removed the `SCHEMA` option

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2213,21 +2213,85 @@ it to be applied to that entity or mapped superclass and its subclasses.
...
+
+This specification defines the following _validation modes_:
+
+* `callback` -- the persistence provider must perform the automatic validation
+  of entities as described in <<a2380>>. If no Jakarta Validation provider is
+  available, the persistence provider must throw a `PersistenceException`.
+* `none` -- the persistence provider must not perform any automatic validation.
+* `auto` -- if a Jakarta Validation provider is obtainable -- persistent provider
+  must perform the automatic validation of entities. Otherwise, the
+  behavior is the same as with `none`.
```
```suggestion
* `CALLBACK` -- the persistence provider must perform the automatic validation
  of entities as described in <<a2380>>. If no Jakarta Validation provider is
  available, the persistence provider must throw a `PersistenceException`.
* `NONE` -- the persistence provider must not perform any automatic validation.
* `AUTO` -- if a Jakarta Validation provider is obtainable -- persistent provider
  must perform the automatic validation of entities. Otherwise, the
  behavior is the same as with `none`.
```

spec/src/main/asciidoc/appA-revision-history.adoc
```diff
@@ -109,6 +109,9 @@ events: pre-persist, pre-merge, pre-remove, pre-delete.
 
 Added `@Discoverable` annotation
 
+Introduced the requirement to derive additional schema metadata from
+Jakarta Validation constraints.
```
gavinking:
```suggestion
Introduced the requirement to derive additional schema metadata from
Jakarta Validation constraints
```

gavinking (approved):
Thanks so much @marko-bekhta !

PR #992, opened 2026-04-06
Rename for consistency with Jakarta Data

### LINKED ISSUE #991: `@StaticQuery`/`@StaticNativeQuery` naming taking into account `Restriction`, `Order`, etc.
As mentioned in https://github.com/jakartaee/data/pull/1409#issuecomment-4125062574, the name of `@StaticQuery` can be surprising, because methods annotated with it are not necessarily about static queries -- with Jakarta Data, they can incorporate a dynamic part through a parameter of type `Restriction`, `Order`, etc.

I'm creating an issue here at the request of @njr-11 , so we can brainstorm.
But let's keep in mind that @gavinking already warned me: we will need to provide an actual PR too. I can volunteer for that.

We could try to rename them to something better. The problem being to find that better name. There is already a lot of naming suggestions in https://github.com/jakartaee/persistence/issues/780 , but apparently none caught up.

What do we know about `@StaticQuery`?

1. It allows declaring query specifications which will be registered in the Jakarta Persistence metamodel. Even if Jakarta Data is not around.
2. It can be also used in Jakarta Data repositories, where it means (exactly?) the same thing as `@Query`.
3. It can be used in combination with `@ReadQueryOptions` and `@WriteQueryOptions` -- same as `@Query`, AFAICS.

Let me know if other aspects need to be taken into account, I'll add them here.

Will try to make constructive suggestions below...

njr-11:
First step toward resolving #991 

We know that we want to rename
`StaticNativeQuery` to `NativeQuery`
`StaticQuery` to `JakartaQuery` ?
`StaticQueryOptions` to just `QueryOptions`.

This PR covers items 1 and 3, while using `JakartaQuery` as a placeholder for item 2 (it accepts Jakarta Persistence Query Language defined in the Jakarta Query specification) which can later be switched to something else once there is agreement on a name.  Hopefully the temporary step will not be objectionable.  This will allow Jakarta Data to fully update its references to items 1 and 3, with only changes to the name for item 2 pending.


njr-11:
Resolve merge conflicts and switch from the placeholder name `PersistenceQuery` to `JakartaQuery` which is now being considered in #991

gavinking:
@njr-11 could you fix the compilation failures in the TCK, please

njr-11:
> @njr-11 could you fix the compilation failures in the TCK, please

Thanks for pointing out the failures.  I missed some references to the classes that appeared after rebasing that didn't show up as merge conflicts.  I added commit to address them and the pre-integration checks are successful now.

PR #995, opened 2026-04-14
introduce mappedById

gavinking:
see #527

cc @sebersole


gavinking:
Ugh, this is a partial dupe of #973.

PR #998, opened 2026-04-16
improve association-defining annotation javadoc

gavinking:
Improves the documentation of the (very important) relationship-defining annotations.

### DIFF
--- a/api/src/main/java/jakarta/persistence/ManyToMany.java
+++ b/api/src/main/java/jakarta/persistence/ManyToMany.java
@@ -144,8 +144,9 @@ public @interface ManyToMany {
     FetchType fetch() default FetchType.LAZY;
 
     /** 
-     * The field that owns the relationship. Required unless 
-     * the relationship is unidirectional.
+     * The field of property of the target entity that owns
+     * the relationship. This element is only specified on
+     * the inverse (non-owning) side of the association.
      *
      * <p>The static metamodel of the target entity may be
      * used to obtain a reference to the owning side, for
--- a/api/src/main/java/jakarta/persistence/ManyToMany.java
+++ b/api/src/main/java/jakarta/persistence/ManyToMany.java
@@ -145,8 +145,9 @@ public @interface ManyToMany {
 
     /** 
      * The field of property of the target entity that owns
-     * the relationship. This element is only specified on
-     * the inverse (non-owning) side of the association.
+     * the relationship and specifies its mapping to the
+     * database. This element is only specified on the
+     * inverse (non-owning) side of the association.
      *
      * <p>The static metamodel of the target entity may be
      * used to obtain a reference to the owning side, for
--- a/api/src/main/java/jakarta/persistence/ManyToMany.java
+++ b/api/src/main/java/jakarta/persistence/ManyToMany.java
@@ -26,18 +26,58 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 /**
  * Specifies a many-valued association with many-to-many multiplicity,
  * mapping to an intermediate table called the <em>join table</em>.
+ * If the target entity class cannot be inferred from the type arguments
+ * of the declared type of the annotated field or property, then
+ * {@link #targetEntity} must be explicitly specified.
  *
- * <p>Every many-to-many association has two sides, the owning side
- * and the non-owning, or inverse, side. The join table is specified
- * on the owning side. If the association is bidirectional, either
- * side may be designated as the owning side, and the non-owning side
- * must use the {@link #mappedBy} element of the {@code ManyToMany}
- * annotation to specify the relationship field or property of the
- * owning side.
+ * <p>The optional {@link JoinTable} annotation specifies the mapped join
+ * table.
+ * {@snippet :
+ * @ManyToMany
+ * @JoinTable(name = "CUST_PHONE",
+ *            joinColumns = @JoinColumn(name = "CUST_ID",
+ *                                      referencedColumnName = "ID"),
+ *            inverseJoinColumns = @JoinColumn(name = "PHONE_ID",
+ *                                             referencedColumnName = "ID"))
+ * Set<PhoneNumber> phones;
+ * }
  *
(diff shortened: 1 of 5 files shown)


api/src/main/java/jakarta/persistence/ManyToMany.java
```diff
@@ -26,18 +26,58 @@
...
+ * owning side of an association determine the updates made to the
+ * relationship in the database. If the inverse side of an association
+ * is modified without a corresponding modification to the owning side,
+ * the behavior is undefined. The persistence provider is permitted to
+ * ignore any modification made only to the inverse side of a
+ * bidirectional association.
+ *
+ * <p>The inverse side of a bidirectional {@code ManyToMany} association
+ * must be a field or property also annotated {@code @ManyToMany} of the
+ * {@linkplain #targetEntity target entity}, and the non-owning side must
```
njr-11: Consistency with "inverse" terminology already used earlier in the same sentence,
```suggestion
 * {@linkplain #targetEntity target entity}, and the inverse side must
```

api/src/main/java/jakarta/persistence/OneToOne.java
```diff
@@ -30,77 +30,128 @@
...
+ * owning side of an association determine the updates made to the
+ * relationship in the database. If the inverse side of an association
+ * is modified without a corresponding modification to the owning side,
+ * the behavior is undefined. The persistence provider is permitted to
+ * ignore any modification made only to the inverse side of a
+ * bidirectional association.
+ *
+ * <p>The inverse side of a bidirectional {@code OneToOne} association
+ * must be a field or property also annotated {@code @OneToOne} of the
+ * {@linkplain #targetEntity target entity}, and the non-owning side
```
njr-11: Use consistent term within sentence
```suggestion
 * {@linkplain #targetEntity target entity}, and the inverse side
```

njr-11 (approved):
These look like good improvements to the documentation.
I noted a couple of places where there was some minor inconsistency in terms within the same sentence.  Approving in advance because those are trivial and optional.

api/src/main/java/jakarta/persistence/ManyToMany.java
```diff
@@ -47,70 +91,27 @@
...
 @Retention(RUNTIME)
 public @interface ManyToMany {
 
     /**
      * (Optional) The entity class that is the target of the
-     * association. Optional only if the collection-valued
-     * relationship property is defined using Java generics.
-     * Must be specified otherwise.
+     * association. Required if the annotated field or
+     * property is declared using a raw collection type.
```
sebersole: Or, technically, any non-precise typing (an interface of the mapped entity e.g.).

api/src/main/java/jakarta/persistence/ManyToOne.java
```diff
@@ -127,8 +180,10 @@
     FetchType fetch() default FetchType.DEFAULT;
 
     /** 
-     * (Optional) Whether the association is optional. If set to
-     * {@code false} then a non-null relationship must always exist.
+     * (Optional) Whether the association is optional. When
+     * {@code optional=false}, the relationship must always exist.
+     * The annotated field or property must not be set to a null
+     * value when persistence provider reads the state of the entity.
```
sebersole:
```suggestion
     * value when persistence provider reads or writes the state of the entity.
```

api/src/main/java/jakarta/persistence/OneToMany.java
```diff
@@ -24,20 +24,77 @@
...
+ * one-to-many multiplicity. If the target entity class cannot be inferred
+ * from the type arguments of the declared type of the annotated field or
+ * property, then {@link #targetEntity} must be explicitly specified.
  *
  * <p>A {@code OneToMany} association usually maps a foreign key column
  * or columns in the table of the associated entity. This mapping may
- * be specified using the {@link JoinColumn} annotation. Alternatively,
- * a unidirectional {@code OneToMany} association is sometimes mapped
- * to a join table using the {@link JoinTable} annotation.
+ * be specified using the {@link JoinColumn} annotation.
```
sebersole, line 34: Maybe make it clear that this is true for unidirectional association?  And/or move the later section about bidirectionally above this since that is more common for one-to-many.

sebersole (approved):
Great changes, much more clear without having to go to the spec.

api/src/main/java/jakarta/persistence/ManyToMany.java
gavinking: @sebersole is this better:

```suggestion
     * property is declared using a raw collection type or 
     * when the associated entity type cannot be inferred
     * from the type argument.
```

api/src/main/java/jakarta/persistence/ManyToOne.java
gavinking: I was trying to find a way to express that the user has to initialize it to a non-null value before the state is flushed, without making any assumptions about *when* the provider reads the value to flush out of the entity instance.

api/src/main/java/jakarta/persistence/OneToMany.java
gavinking: Even for a bidirectional `@OneToMany`, the mapping to the FK is specified using `@JoinColumn`. It's just that the annotation goes on the other side.

api/src/main/java/jakarta/persistence/ManyToMany.java
sebersole: Maybe 
```suggestion
* when the associated entity type cannot be inferred
* from the type argument (e.g. when using a raw collection type).
```

gavinking: I pushed a fix for this.

api/src/main/java/jakarta/persistence/ManyToOne.java
gavinking: I expanded this section to make clear what I mean here.

gavinking:
I'm happy with this now, and so I'm going to merge it; doesn't preclude further improvements later.

PR #1009, opened 2026-04-19
clarify the semantics of null collections

### LINKED ISSUE #964: null collections
It has come to my attention that the spec does not assign any interpretation to a `null` collection in an entity. That's not good at all, and I think we should fix that (or at least call it out as being explicitly undefined).

Hibernate treats a `null` collection as equivalent to an empty collection.

@anija-anil What does EclipseLink do?

gavinking:
see #964


njr-11 (approved):

gavinking:
@sebersole any thoughts?

gavinking:
I'm going to merge this now, so that it makes it into the m2 release.

Issue #784, opened 2025-11-26
more flexible application of entity listeners to entities


gavinking:
This issue follows up on the very ancient issue #43.

Something I guess _nobody_ has ever liked is the `@EntityListeners` annotation, which is sort-of backwards, or at least lacking in indirection. After #687 and #730, entity listeners are already starting to look quite a lot more flexible, and so this sticks out worse than it did before.

I think we should consider a range of ways to address this, not just what's proposed in #43. Off the top of my head, I can think of the following straightforward approaches:

1. Allow `@EntityListeners` at the `PACKAGE` level, to apply to all `@Entity` classes in the package.
2. Allow use of `@EntityListeners` as a meta-annotation, as proposed by #43.
3. Add a new `@EntityListener` annotation, taking a list of classes the listener applies to, including the possibility of super types, so that `@EntityListener(Object.class) class MyListener { ... }` would declare a listener that applies to all entity types.

These options are not mutually-exclusive, and my knee-jerk prejudice is that I would sorta like to do 2+3. Option 2 works rather nicely 

But I would like to hear some feedback on this before writing up a proposal.

gavinking:
Options 1 and 2 probably imply adding "entity listener classes" to the list of managed class types here: https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2#list-of-managed-classes

gavinking:
Addition of `@Discoverable` has made this issue easier to solve / write the spec for.

sebersole:
I'm not a fan of (3), but not particularly opposed to it.

I do actually think the "directionality" of `@EntityListeners` is correct - you use it to name listeners.  I think its just unwieldy because of some of the other points brought up previously - namely the need to attach them to classes and an inability to compose them.

I do really like (1) as well.  Combine with (2) I think it makes `@EntityListeners` much more flexible, which ofc was your intent 😄

gavinking:
@sebersole  Hrrrrmmmmm ... 'cos I just got done putting together a proposal along the lines of 3. :-/

Well I will push it anyway so you can have a look.

sebersole:
> [@sebersole](https://github.com/sebersole) Hrrrrmmmmm ... 'cos I just got done putting together a proposal along the lines of 3. :-/
> 
> Well I will push it anyway so you can have a look.


> but not particularly opposed to it.

It is just not the way *I* would do it as an app developer.  But that's not to say others feel the same.

Anyway, I'll definitely take a look.

[...]

PR #1014, opened 2026-04-22
allow @EntityListeners at the package or module level

gavinking:
leverage @Discoverable

see #784

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -1657,6 +1657,10 @@ declared using annotations or XML.
   by annotating the entity class or one of its mapped superclasses with
   the `EntityListeners` annotation. The listener classes must be explicitly
   listed in the order they are invoked.
+- A list of entity listeners applying to every entity class in a given
+  package or module may be specified by annotating the package or module
+  descriptor with the `EntityListeners` annotation. The listener classes
+  must be explicitly listed in the order they are invoked.
 - An entity listener class annotated with the `EntityListener` annotation
   applies to an entity class if the entity class is assignable to one of
   its lifecycle callback methods.
@@ -1947,8 +1951,16 @@ following rules.
    `EntityListeners` annotation. This specification does not define any
    specific order for the invocation of these listeners.
 
-3. Listeners declared by the entity class itself, via the `EntityListeners`
-   annotation or `<entity-listeners>` XML element are invoked in the same
+3. Listeners declared by a module descriptor, via the `EntityListeners`
+   annotation, are invoked in the same order in which they are declared
+   by the annotation.
+
+4. Listeners declared by a package descriptor, via the `EntityListeners`
+   annotation, are invoked in the same order in which they are declared
+   by the annotation.
+
+5. Listeners declared by the entity class itself, via the `EntityListeners`
+   annotation or `<entity-listeners>` XML element, are invoked in the same
    order in which they are declared by the annotation or XML element.
    If multiple classes in an entity inheritance hierarchy--including both
    entity classes and mapped superclasses--declare entity listeners, the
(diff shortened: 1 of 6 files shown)


gavinking:
@sebersole WDYT?

sebersole (approved):
Makes sense

njr-11 (approved):

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1947,8 +1951,16 @@ following rules.
...
    specific order for the invocation of these listeners.
 
-3. Listeners declared by the entity class itself, via the `EntityListeners`
-   annotation or `<entity-listeners>` XML element are invoked in the same
+3. Listeners declared by a module descriptor, via the `EntityListeners`
+   annotation, are invoked in the same order in which they are declared
+   by the annotation.
+
+4. Listeners declared by a package descriptor, via the `EntityListeners`
+   annotation, are invoked in the same order in which they are declared
```
scottmarlow, line 1956: If the EntityListener is defined at both the class and package level, do we care to specify how that is handled (e.g. the package level order overrides the class level)?

gavinking: We don't currently say what happens if it's specified in two different places (e.g. superclass, or default listeners in XML). So this change doesn't really make the situation any worse.

If you want to clarify that, open a separate issue, because it would take some thought.

marko-bekhta (commented):
Makes sense 👍🏻 🙂 a couple questions:
* Do we want to have `@ExcludePackageListeners`/ `@ExcludeModuleListeners` (to match existing `@Exclude***Listeners` annotations ?
* Do we need to represent these in any way in the XML ?

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: > open a separate issue

+1, I was thinking that it may make sense to do the default -> module -> package -> class listeners (superclass first, subclass ones next..) 
but then ... maybe not 🤔

gavinking:
> * Do we want to have `@ExcludePackageListeners`/ `@ExcludeModuleListeners` (to match existing `@Exclude***Listeners` annotations ?

It's a good point, and one I wondered about.

And I guess my answer is: not with the current way this is handled. Those (rarely-used) annotations are unfortunately declared in the top-level namespace, which is already very crowded.

It would have been much better if they were nested annotations or even just a completely different design. But redesigning how listener exclusion works is just nowhere near the top of the list of things I want to spend our time and complexity budget on right now.

> * Do we need to represent these in any way in the XML ?

I don't think so, since:

1. the XML already has the "default listeners" stuff, and
2. we don't (currently) have any XML element for specifying package/module stuff.

So I think it's reasonable to leave that out.

marko-bekhta (approved):

gavinking:
@marko-bekhta I have addressed the problem of exclusion by just saying that package/module-level entity listeners go in the "default listeners" bucket, and so you can exclude them using `@ExcludeDefaultListeners`. I think that works nicely.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: I think what makes most sense is to say there is at most one listener per listener class. But then I don't know what the order would be.

marko-bekhta (approved):

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: Anyway, it's definitely a separate issue.

gavinking:
Alright, we seem to have a good consensus here; I guess I will just go ahead and merge this one in a bit.

scottmarlow (approved):

spec/src/main/asciidoc/ch03-entity-operations.adoc
scottmarlow: Added https://github.com/jakartaee/persistence/issues/1020 which needs more detail but a start...

PR #1022, opened 2026-04-25
add kind to @Table annotation

### LINKED ISSUE #1021: @Table(kind=...)
In #947 we added `type` and `kind` members to `@Index`. Actually @beikov proposed something similar for the `@Table` annotation a long time ago: https://github.com/hibernate/hibernate-orm/discussions/5607#discussioncomment-16712322.

I think it's weird to have this on `@Index` but not on `@Table`.

gavinking:
see #1021

### DIFF
--- a/api/src/main/java/jakarta/persistence/CollectionTable.java
+++ b/api/src/main/java/jakarta/persistence/CollectionTable.java
@@ -189,7 +189,7 @@ public @interface CollectionTable {
     String comment() default "";
 
     /**
-     * (Optional) A SQL fragment representing the table kind,
+     * (Optional) A SQL fragment indicating the type of table,
      * usually inserted before the keyword {@code table} in
      * the generated DDL which creates this table.
      *
--- a/api/src/main/java/jakarta/persistence/CollectionTable.java
+++ b/api/src/main/java/jakarta/persistence/CollectionTable.java
@@ -189,6 +189,15 @@ public @interface CollectionTable {
     String comment() default "";
 
     /**
+     * (Optional) A SQL fragment representing the table kind,
+     * usually inserted before the keyword {@code table} in
+     * the generated DDL which creates this table.
+     *
+     * @since 4.0
+     */
+    String type() default "";
+
+    /**
      * (Optional) A SQL fragment appended to the generated DDL
      * statement which creates this table. This is only used if
      * table generation is in effect.
--- a/api/src/main/java/jakarta/persistence/Index.java
+++ b/api/src/main/java/jakarta/persistence/Index.java
@@ -67,6 +67,9 @@ public @interface Index {
      * usually inserted before the keyword {@code index} in the
      * generated DDL which creates this index.
      *
+     * <p>For example, {@code type = "UNIQUE"} results in
+     * {@code CREATE UNIQUE INDEX ...}.
+     *
      * @since 4.0
      */
     String type() default "";
@@ -76,6 +79,9 @@ public @interface Index {
      * When specified, it is inserted into the generated DDL which
      * creates this index, usually following a {@code USING} keyword.
      *
+     * <p>For example, {@code using = "BTREE"} results in
+     * {@code CREATE INDEX ... USING BTREE ...}.
+     *
      * @since 4.0
      */
     String using() default "";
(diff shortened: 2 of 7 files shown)


gavinking:
WDYT @beikov @njr-11 ?

api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -64,21 +64,21 @@
...
-     * When specified, it is inserted into the generated DDL 
-     * which creates this index.
+     * (Optional) A SQL fragment representing the indexing method.
+     * When specified, it is inserted into the generated DDL which
+     * creates this index, usually following a {@code USING} keyword.
      *
      * @since 4.0
      */
-    String type() default "";
+    String using() default "";
```
beikov, line 87: I'm a bit torn on the term `using`. PostgreSQL uses this keyword for the definition of this configuration, but it feels a bit too PostgreSQL-specific, since it is the only database with this syntax. I'm not sure if an abstract term like `kind` would be any better, but I thought I just leave this concern here. Maybe someone has a better idea.

gavinking: It's not just Postgres, it's also MySQL. And those are the only databases I've seen which have anything that goes between the index name and the column list. (Remember your list here: https://github.com/jakartaee/persistence/issues/934)

api/src/main/java/jakarta/persistence/Table.java
```diff
@@ -101,6 +101,15 @@
      */
     String comment() default "";
 
+    /**
+     * (Optional) A SQL fragment representing the table kind,
```
njr-11: Given that we are renaming it to `type`,
```suggestion
     * (Optional) A SQL fragment indicating the type of table,
```

api/src/main/java/jakarta/persistence/Table.java
```diff
@@ -101,6 +101,15 @@
...
+     * usually inserted before the keyword {@code table} in
+     * the generated DDL which creates this table.
```
njr-11, line 107: Should we add an example, such as `CREATE TEMPORARY TABLE ...` ?

api/src/main/java/jakarta/persistence/CollectionTable.java
```diff
@@ -188,6 +188,15 @@
      */
     String comment() default "";
 
+    /**
+     * (Optional) A SQL fragment representing the table kind,
```
njr-11: Same suggestion here:
```suggestion
     * (Optional) A SQL fragment indicating the type of table,
```

api/src/main/java/jakarta/persistence/Index.java
```diff
@@ -64,21 +64,21 @@
 
     /**
      * (Optional) A SQL fragment representing the index kind,
```
njr-11: Similar to prior comments,
```suggestion
     * (Optional) A SQL fragment indicating the type of index,
```

api/src/main/java/jakarta/persistence/JoinTable.java
```diff
@@ -164,6 +164,15 @@
      */
     String comment() default "";
 
+    /**
+     * (Optional) A SQL fragment representing the table kind,
```
njr-11:
```suggestion
     * (Optional) A SQL fragment indicating the type of table,
```

api/src/main/java/jakarta/persistence/SecondaryTable.java
```diff
@@ -139,6 +139,15 @@
      */
     String comment() default "";
 
+    /**
+     * (Optional) A SQL fragment representing the table kind,
```
njr-11:
```suggestion
     * (Optional) A SQL fragment indicating the type of table,
```

njr-11 (commented):
Suggested a wording change to align with the renamed method.

api/src/main/java/jakarta/persistence/Table.java
gavinking: @njr-11 Done, WDYT?

njr-11 (approved):

gavinking:
I would say this one is ready to go.

PR #1026, opened 2026-05-01
Query and EntityAgent/Manager options

### LINKED ISSUE #1024: QueryOption
We have discussed several times whether there should be a `QueryOption` interface to replace `Query.setHint()` which lacks type safety.

I think we're coming around to the view that there's no downside to just throwing it in.

gavinking:
Adds `TypedQuery.Option` and `Statement.Option`. See #1024.


gavinking:
cc @sebersole

gavinking:
This turned out much more nicely than I had expected.

sebersole:
Love it.  The options are a much more self-documenting way of providing hints.

The only thing I don't love is hints and options, sometimes "the same", side-by-side.

gavinking:
> The only thing I don't love is hints and options, sometimes "the same", side-by-side.

Yeah well I guess we're moving toward deprecating the hints.

sebersole:
Cool, that was my nudge ;)

api/src/main/java/jakarta/persistence/query/StaticStatementReference.java
```diff
@@ -181,7 +181,7 @@ public String toString() {
                 + "name=" + name + ", "
                 + "parameterNames=" + parameterNames + ", "
                 + "arguments=" + arguments + ", "
-                + "timeout=" + timeout + ", "
+                + "timeout=" + options + ", "
```
njr-11:
```suggestion
                + "options=" + options + ", "
```

api/src/main/java/jakarta/persistence/query/StaticStatementReference.java
```diff
@@ -148,8 +148,8 @@ public String getAnnotatedMemberName() {
     }
 
     @Override
-    public Timeout getTimeout() {
-        return timeout;
+    public Set<Statement.Option> getOptions() {
+        return new HashSet<>(options);
```
njr-11, line 152: Can return the read-only instance that was already created with Set.of()
```suggestion
        return options;
```

api/src/main/java/jakarta/persistence/query/StaticTypedQueryReference.java
```diff
@@ -192,33 +164,13 @@ public String getAnnotatedMemberName() {
...
-        return timeout;
+    public String getEntityGraphName() {
+        return entityGraphName;
     }
 
     @Override
-    public String getEntityGraphName() {
-        return entityGraphName;
+    public Set<TypedQuery.Option> getOptions() {
+        return new HashSet<>(options);
```
njr-11, line 173:
```suggestion
        return options;
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -356,4 +357,49 @@ public interface EntityAgent extends EntityHandler {
...
+     * {@link #setCacheRetrieveMode}, {@link #setCacheStoreMode}.
+     * Mutation of the returned set does not affect the options of
+     * the entity agent.
+     *
+     * @return the options for this query
+     */
+    Set<Option> getOptions();
+
+    /**
+     * An option controlling the creation of an entity manager.
```
```suggestion
     * An option controlling the creation of an entity agent.
```

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -841,4 +842,55 @@ void refresh(Object entity,
...
+
+    /**
+     * Get the {@linkplain Option options} controlling the behavior
+     * of this entity manager. The returned set includes options set
+     * via {@link #addOption}, along with options specified via
+     * {@link #setCacheRetrieveMode}, {@link #setCacheStoreMode}, or
+     * {@link #setFlushMode}. Mutation of the returned set does not
+     * affect the options of the entity manager.
+     *
+     * @return the options for this query
```
njr-11:
```suggestion
     * @return the options for this entity manager
```

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -173,11 +173,13 @@ public interface EntityManagerFactory extends AutoCloseable {
...
      * <p>The {@link EntityManager#isOpen} method will return true
      * on the returned instance.
+     * @param options options controlling the behavior of the entity
+     *                manager
      * @return entity manager instance
      * @throws IllegalStateException if the entity manager factory
      * has been closed
      */
-    EntityManager createEntityManager();
+    EntityManager createEntityManager(EntityManager.CreationOption... options);
```
njr-11, line 182: With this change, existing code will continue to compile, but already compiled code will likely need to be recompiled. I think that already might be a requirement elsewhere for 4.0, in which case this is not an issue because applications will already need to be recompiled anyway.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -190,58 +192,48 @@ public interface EntityManagerFactory extends AutoCloseable {
...
-     * @param synchronizationType  how and when the entity manager should
-     *                             be synchronized with the current JTA
-     *                             transaction
-     * @return entity manager instance
-     * @throws IllegalStateException if the entity manager factory has
-     * been configured for resource-local entity managers or is closed
-     *
-     * @since 2.1
-     */
-    EntityManager createEntityManager(SynchronizationType synchronizationType);
```
njr-11, line 211: This is a breaking change to applications if they are currently using this method.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -190,58 +192,48 @@ public interface EntityManagerFactory extends AutoCloseable {
...
+                                      Map<?, ?> map);
 
     /**
      * Create a new application-managed {@link EntityAgent}. This
      * method returns a new {@code EntityAgent} instance each time
      * it is invoked.
      * <p>The {@link EntityAgent#isOpen} method will return true
      * on the returned instance.
+     * @param options options controlling the behavior of the entity
+     *                manager
```
njr-11:
```suggestion
     * @param options options controlling the behavior of the entity
     *                agent
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -356,4 +357,49 @@ public interface EntityAgent extends EntityHandler {
...
+     *
+     * @param option the option
+     */
+    void addOption(Option option);
+
+    /**
+     * Get the {@linkplain Option options} controlling the behavior
+     * of this entity agent. The returned set includes options set
+     * via {@link #addOption}, along with options specified via
+     * {@link #setCacheRetrieveMode}, {@link #setCacheStoreMode}.
```
```suggestion
     * via {@link #addOption}, along with options specified via
     * {@link #setCacheRetrieveMode} or {@link #setCacheStoreMode}.
```

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -356,4 +357,49 @@ public interface EntityAgent extends EntityHandler {
...
+     * Mutation of the returned set does not affect the options of
+     * the entity agent.
+     *
+     * @return the options for this query
```
```suggestion
     * @return the options for this entity agent
```

njr-11 (commented):
This looks very good and turned out well.  I added a few minor suggestions for copy/paste typos and pointed out a case where an existing API method is removed.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: It's source-level compatible because `SynchronizationType` now implements `CreationOption`.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: Yes, correct, we already broke binary compatibility in a couple of other places, which is OK to do in a major version.

api/src/main/java/jakarta/persistence/query/StaticTypedQueryReference.java
gavinking: There's no perfect solution here, since Java doesn't have proper immutable collection types, but I guess my weak preference is to return a copy of the set rather than return something that throws an exception from half its operations.

gavinking:
Thanks for the review @njr-11 those were some really good catches.

I've replied to a couple of your comments.

njr-11 (approved):

gavinking:
Looks like everyone is happy with this.

PR #1034, opened 2026-05-07
improve the javadoc of @Entity and AccessType

gavinking:
and say explicitly that persistence fields must have a concrete type

### DIFF
--- a/api/src/main/java/jakarta/persistence/Entity.java
+++ b/api/src/main/java/jakarta/persistence/Entity.java
@@ -56,12 +56,15 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * <li>{@linkplain AccessType#FIELD field access}, that is, direct
  *     access to instance variables.
  * </ul>
- * <p>The entity {@linkplain AccessType access type} determines whether
- * the provider accesses the state of the entity using getter and setter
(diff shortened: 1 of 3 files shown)


api/src/main/java/jakarta/persistence/Entity.java
```diff
@@ -48,6 +48,29 @@
...
+ * runtime using either:
+ * <ul>
+ * <li>indirect {@linkplain AccessType#PROPERTY property access} via
+ *     JavaBeans-style property accessors, or
+ * <li>{@linkplain AccessType#FIELD field access}, that is, direct
+ *     access to instance variables.
+ * </ul>
+ * <p>The entity {@linkplain AccessType access type} determines whether
+ * the provider accesses the state of the entity using getter and setter
+ * methods, or by direct to its fields. It is almost never necessary to
```
marko-bekhta:
```suggestion
 * methods, or by directly accessing its fields. It is almost never necessary to
```
maybe ?

gavinking: Yeah, good catch. I did some rewriting. Look now.

marko-bekhta (approved):

njr-11 (approved):

gavinking:
thanks folks

PR #1037, opened 2026-05-08
add overloads of treat(), correlate() for dealing with From instances

### LINKED ISSUE #190: A few missing API methods in the Criteria API
The following methods are inexplicably missing from the Criteria API:

* `CriteriaBuilder.treat(From, Class)`
* `Subquery.correlate(From)`

As a result, programmers have to use a sibling method with a wider type, causing them to have to add unchecked casts to compensate, etc.

gavinking:
also add convenience treat() method to Path and subtypes

See #190

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/From.java
+++ b/api/src/main/java/jakarta/persistence/criteria/From.java
@@ -356,4 +356,12 @@ public interface From<Z, X> extends Path<X>, FetchParent<Z, X> {
      *          Use of the typesafe version is strongly preferred.
      */
     <K, V> MapJoin<X, K, V> joinMap(String attributeName, JoinType jt);
+
+    /**
+     * Downcast the bound type to the given type.
+     * @param type a subtype of the bound type
+     * @return this root or join downcast to the given type
+     * @since 4.0
+     */
+    <T extends X> From<?, T> treat(Class<T> type);
 }
--- a/api/src/main/java/jakarta/persistence/criteria/From.java
+++ b/api/src/main/java/jakarta/persistence/criteria/From.java
@@ -74,6 +74,7 @@ public interface From<Z, X> extends Path<X>, FetchParent<Z, X> {
     /**
      * Create and add an inner join to the given entity.
      * @param entityClass  the target entity class
+     * @param <Y> the target entity type
      * @return the resulting join
      * @since 3.2
      */
@@ -83,6 +84,7 @@ public interface From<Z, X> extends Path<X>, FetchParent<Z, X> {
      * Create and add a join to the given entity.
      * @param entityClass  the target entity class
      * @param joinType  join type
+     * @param <Y> the target entity type
      * @return the resulting join
      * @since 3.2
      */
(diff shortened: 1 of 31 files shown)


gavinking:
@beikov Please take a look at this one. Does it look OK to you?

api/src/main/java/jakarta/persistence/criteria/From.java
```diff
@@ -356,4 +356,12 @@ public interface From<Z, X> extends Path<X>, FetchParent<Z, X> {
...
      */
     <K, V> MapJoin<X, K, V> joinMap(String attributeName, JoinType jt);
+
+    /**
+     * Downcast the bound type to the given type.
+     * @param type a subtype of the bound type
+     * @return this root or join downcast to the given type
+     * @since 4.0
+     */
+    <T extends X> From<?, T> treat(Class<T> type);
```
beikov, line 396:
```suggestion
    <T extends X> From<Z, T> treat(Class<T> type);
```

njr-11 (commented):
I noticed lots of places where type parameters are not documented in the javadoc with `@param` although it also looks like these classes haven't been doing that in general prior to these added methods.

api/src/main/java/jakarta/persistence/criteria/From.java
gavinking: Nope, can't do that, the covariant overrides don't pass the typechecker. (That may be a compiler bug but it is what it is.)

gavinking:
> it also looks like these classes haven't been doing that in general prior to these added methods

A quick search for `@param T`, `@param X`, `@param Y`, and so on turns up no hits.

gavinking:
@njr-11 in [this commit](https://github.com/jakartaee/persistence/pull/1037/changes#diff-467686cb03920d579762f260cc5abfb084f732052b277df002828e5558b6e0f0) I've addressed your objection by consistently adding `@param` tags to all type parameters.

api/src/main/java/jakarta/persistence/criteria/PluralJoin.java
```diff
@@ -38,4 +38,13 @@ public interface PluralJoin<Z, C, E> extends Join<Z, E> {
...
     PluralAttribute<? super Z, C, E> getModel();
+
+    /**
+     * Downcast the element type to the given type.
+     * @param type a subtype of the element type
+     * @param <T> the subtype of the element type
+     * @return this plural join downcast to the given element type
+     * @since 4.0
+     */
+    <T extends E> PluralJoin<Z, ?, T> treat(Class<T> type);
```
marko-bekhta, line 49: mm so we are losing the collection type here?

marko-bekhta (approved):

api/src/main/java/jakarta/persistence/criteria/PluralJoin.java
gavinking: Only if you exactly have a `PluralJoin`. But you're almost always going to have one of its subtypes. (Again, this is a workaround for what might very well be a compiler bug.)

njr-11 (approved):

gavinking:
Thank you all. 🙏

PR #1038, opened 2026-05-08
default entity name for inner class (clarifying footnote)

### LINKED ISSUE #1036: default entity name for an inner class
We say that the entity name

> defaults to the unqualified name of the entity class

but this is quite ambiguous for a nested class, which we explicitly allow.

- Read extremely literally, the name of the outer class is a "qualifier" of some sort, so arguably this means that the entity name defaults to the class name, with the outer class name being ignored.
- For many years Hibernate has defaulted the entity name is such cases to `Outer$Inner`, the format used in Java class files, but not in Java source. But this is actually working out quite badly; it results in very ugly default table names, which are rejected on some databases, requiring the user to explicitly specify an entity name or table name.
- We could in principle resolve this by allowing entity names to be qualified names of form `Outer.Inner` (they are currently required to be legal Java identifiers). This change would impact the grammar of Jakarta Query, but is completely doable.

Thoughts?

gavinking:
See #1036

### DIFF
--- a/spec/src/main/asciidoc/ch04-query-language.adoc
+++ b/spec/src/main/asciidoc/ch04-query-language.adoc
@@ -203,12 +203,15 @@ related entities and use their abstract schema types in the query.
 
 ==== Naming [[naming]]
 
-Entities are designated in query strings by their entity names. The
-entity name is specified by the `name` element of the `Entity` annotation,
-as defined in <<entity-annotation>>, or by the `entity-name` XML descriptor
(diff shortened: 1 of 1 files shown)


sebersole (approved):

njr-11 (approved):
Looks good, and I agree most users would expect the unqualified name of an inner class to be the simple class name, not qualified by the outer class name, and would likely prefer this behavior.

gavinking:
Cool, that looks as close as we're ever going to get to a universal consensus so I will merge this :)

PR #1039, opened 2026-05-09
talk about manager/agent and query options in spec

gavinking:
Mention this new stuff in the existing sections on properties/hints.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityAgent.java
+++ b/api/src/main/java/jakarta/persistence/EntityAgent.java
@@ -369,7 +369,8 @@ public interface EntityAgent extends EntityHandler {
 
     /**
      * Specify an {@linkplain Option option} controlling the
-     * behavior of this entity agent.
+     * behavior of this entity agent, overwriting any existing
+     * option of the same type.
      *
      * @param option the option
      */
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -3399,7 +3399,7 @@ ignored if it is not understood by the provider.
 
 Alternatively, a _query option_ is a type safe object carrying the
 same sort of information which might be provided in a query hint. The
-interfaces `Statememt`, and `TypedQuery`, and `StoredProcedureQuery`
+interfaces `Statement`, `TypedQuery`, and `StoredProcedureQuery`
 define inner `Option` interfaces, and this specification provides a
 number of built-in concrete option types implementing these option
 interfaces. Providers are encouraged to define additional custom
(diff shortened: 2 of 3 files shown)


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -3382,6 +3397,20 @@ vendor-specific hints. Vendor-specific hints must not have names in
 the `jakarta.persistence` namespace. A vendor-specific hint must be
 ignored if it is not understood by the provider.
 
+Alternatively, a _query option_ is a type safe object carrying the
+same sort of information which might be provided in a query hint. The
+interfaces `Statememt`, and `TypedQuery`, and `StoredProcedureQuery`
```
```suggestion
interfaces `Statement`, `TypedQuery`, and `StoredProcedureQuery`
```

gavinking: Thanks, fixed.

api/src/main/java/jakarta/persistence/EntityAgent.java
```diff
@@ -369,7 +369,8 @@ public interface EntityAgent extends EntityHandler {
 
     /**
      * Specify an {@linkplain Option option} controlling the
-     * behavior of this entity agent.
+     * behavior of this entity agent, overwriting any existing
+     * option of the same type.
```
gavinking, line 373: Just noticed that this language has gone missing in two places.

marko-bekhta (approved):
Looks good 👍🏻

Unrelated to the changes of this PR... but since I was looking at the options of the statement/query:

> The returned set includes options set via
     * {@link #addOption}, along with options specified via
     * {@link #setTimeout} or {@link #setQueryFlushMode}.

Since we've introduced these in 4, should we just go ahead and remove them in favour of adding an option? Or maybe replace them with a default implementation delegating to the `addOption` ?

Another thought about `Set<Option> getOptions();`.. I couldn't think of a use case to get all options right away, but having something like
```java
<T extends Statement.Option> T get[find]Option(Class<T> optionType);
```
might be useful ?

gavinking:
> Since we've introduced these in 4, should we just go ahead and remove them in favour of adding an option? Or maybe replace them with a default implementation delegating to the `addOption` ?

I wrestled with precisely this question, but ultimately:

1. `Query` is an interface dedicated directly to dealing with these things, so this is not API bloat
2. having them there makes these options much more discoverable
3. `QueryFlushMode` would have a harder time competing with `FlushModeType` if it doesn't have a matching method
4. `Timeout` would have a harder time competing with `setTimeout(Integer)` without the matching overload
5. The real purpose of `addOption()` is to deal with *custom options*, not to deal with the built-in ones. (FTR, I could make the same argument about `AttributeNode.addOption()` and perhaps I should!)

gavinking:
>  I couldn't think of a use case to get all options right away, but having something like
>
>     <T extends Statement.Option> T getOption(Class<T> optionType);

Certainly considered this and maybe we should eventually add them but:

1. if so we should do it everywhere, also for other option types, and
2. we can easily add them later, they don't compete with anything we already have

njr-11 (approved):

gavinking:
Thanks folks.

PR #1042, opened 2026-05-10
add TCK tests for untested features of JPA 3.2

gavinking:
Some things we added in 3.2 never had TCK tests. This PR rectifies that.


marko-bekhta (approved):
Looking through the tests, I see they follow the pattern by extending the `PMClientBase`, and they use assertions, which is a great improvement and should help us detect problems if a test run fails.

Testing this against our current development branch for Hiberante ORM there are +2 test failures which seem to be bugs on the ORM side. 

With that said, I think we should merge these tests into the next release and use them to help verify the implementation.

gavinking:
After some discussion, we've decided to go ahead and merge this, so that @sebersole can make progress with his implementation of JPA 4.

PR #1045, opened 2026-05-11
fix use of unconstrained type parameter in Expression.selectCase

gavinking:
it's easier to specify the type as an argument than as a type argument

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
+++ b/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
@@ -1910,16 +1910,26 @@ public interface CriteriaBuilder {
 		 */
 		Expression<R> otherwise(Expression<? extends R> result);
 	}
-	
+
+    /**
+     * Create a simple case expression.
+     * @param expression  to be tested against the case conditions
+     * @param type the type of the result of the case expression
+     * @return simple case expression
+     */
+    <C, R> SimpleCase<C,R> selectCase(Expression<? extends C> expression, Class<R> type);
+
     /**
      * Create a simple case expression.
      * @param expression  to be tested against the case conditions
      * @param <C> the type of the case expression
      * @param <R> the result type of the case expression
      * @return simple case expression
+     * @deprecated This operation declares an unconstrained type variable.
+     *             Use {@link #selectCase(Expression,Class)} instead.
      */
+    @Deprecated(since = "4.0")
     <C, R> SimpleCase<C,R> selectCase(Expression<? extends C> expression);
-
 
     /**
      * Interface used to build general case expressions.
@@ -1960,12 +1970,22 @@ public interface CriteriaBuilder {
 		 */
 		Expression<R> otherwise(Expression<? extends R> result);
 	}
-	
+
+    /**
+     * Create a general case expression.
+     * @param type the type of the result of the case expression
+     * @return general case expression
+     */
+    <R> Case<R> selectCase(Class<R> type);
+
     /**
      * Create a general case expression.
      * @param <R> the result type of the case expression
      * @return general case expression
+     * @deprecated This operation declares an unconstrained type variable.
+     *             Use {@link #selectCase(Class)} instead.
      */
+    @Deprecated(since = "4.0")
     <R> Case<R> selectCase();
 
     /**
--- a/api/src/main/java/jakarta/persistence/criteria/Expression.java
+++ b/api/src/main/java/jakarta/persistence/criteria/Expression.java
@@ -198,11 +198,12 @@ public interface Expression<T> extends Selection<T> {
 
     /**
      * Create a simple case expression to test against this expression.
+     * @param type the type of the result of the case expression
      * @param <R> the result type of the case expression
      * @return simple case expression
      * @since 4.0
      */
-    <R> SimpleCase<T,R> selectCase();
+    <R> SimpleCase<T,R> selectCase(Class<R> type);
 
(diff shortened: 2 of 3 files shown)


gavinking:
If there are no objections to this, I think I'm just going to merge it quickly.

beikov:
I'm not a fan of specifying the result type here, because the `CriteriaBuilder#selectCase` method also doesn't require this. It just feels odd to "break" with the pattern of how we moved/copied `CriteriaBuilder` methods to their first argument type class.

gavinking:
The problem is that most users won't immediately realize that they have to explicitly specify the type argument, since Java doesn't require it, and the method simply doesn't work without it; you just get typing errors.

I don't think there's a particularly strong reason to  maintain consistency between two completely different APIs, but if you prefer, I can add an overload to CB that accepts a type, and deprecate the existing method.

beikov:
>The problem is that most users won't immediately realize that they have to explicitly specify the type argument, since Java doesn't require it, and the method simply doesn't work without it; you just get typing errors.

I haven't tried, but I would imagine that the type inference would just infer `T=Object` and unless you're using the `SimpleCase` in a context, that requires an explicit type, you should be fine. Users of that API simply have to understand that they have to provide a type variable assignment via e.g. `var selectCase = expression.<MyResultType>selectCase()` or through inference `SimpleCase<ConditionType, MyResultType> selectCase = expression.selectCase();`

>I don't think there's a particularly strong reason to maintain consistency between two completely different APIs, but if you prefer, I can add an overload to CB that accepts a type, and deprecate the existing method.

I don't see a strong reason to specify a class argument when you can already assign the type variable in Java. Especially, because people are probably used to doing this anyway if they worked with `CriteriaBuilder#selectCase` before. It's roughly the same from a verbosity perspective, but keeping the way it is now, at least we'd follow a "pattern".

gavinking:
> I haven't tried

Well I did try it. That's how I noticed that this was a problem.

> but I would imagine that the type inference would just infer T=Object and unless you're using the SimpleCase in a context, that requires an explicit type, you should be fine. 

But in fact you're not fine, because then:

1. the `SimpleCase<Object>` can't be assigned to the thing you want to assign it to, and
2. it accepts any inconsistent mix of results, resulting in a much greater chance of runtime errors.

> I don't see a strong reason to specify a class argument when you can already assign the type variable in Java.

You are in practice always going to have to be explicit. So the strong reason is that you want the signature of the method to *force* you to be explicit instead of trapping you into thinking you can let the type argument be inferred when it can't be. IDEs give *much* better guidance for parameters than they do for type parameters.

> keeping the way it is now, at least we'd follow a "pattern".

We'd be following an anti-pattern. Unconstrained type parameters are almost always wrong, and wreak havoc with type inference. They were always bad, but they're even worse now that Java has `var`.

gavinking:
Screenshot 2026-05-11 at 3 45 04 PM

beikov:
All I'm saying is that the compiler is already explicit about this, and that people can define the type variable assignment without having to pass a method argument, but if you think that is so much better this way, even though a `Class` argument will produce warnings in the presence of parameterized types, it's fine.

gavinking:
It is *much* better, yes. If you try it you will see. You get clearer errors, better autocompletion, and so on.

gavinking:
> a `Class` argument will produce warnings in the presence of parameterized types, it's fine.

None of our basic types is a parameterized type, and a `case` expression can't evaluate to a collection or entity/embeddable , so I don't think this matters.

gavinking:
An alternative solution to the problem would be to simply move away from the builder pattern for this and write something like:

```java
// simple form
label.selectCase(builder.literal('hello').then('world'),
                 builder.literal('goodbye').then('sweet world'),
                 builder.otherwise(''))
```

```java
// "searched" form
builder.selectCase(label.equalTo('hello').then('world'),
                   label.equalTo('goodbye').then('sweet world'),
                   builder.otherwise(''))
```

or even just:

```java
// simple form
label.selectCase(builder.literal('hello').then('world'),
                 builder.literal('goodbye').then('sweet world'),
                 builder.literal('').otherwise())   // postfix otherwise
```

```java
// "searched" form
builder.selectCase(label.equalTo('hello').then('world'),
                   label.equalTo('goodbye').then('sweet world'),
                   builder.literal('').otherwise())   // postfix otherwise
```


the only issue with that is that I don't see a good way to constrain it to have at most one `otherwise()`.

Note that the `then()` method would be declared on `Expression`, so that any expression can be used as a "case".

Aaaaactually, y'know what? It might be pretty trivial to just repurpose the existing builders to let you write:

```java
// simple form
label.selectCase(builder.literal('hello').then('world'),
                 builder.literal('goodbye').then('sweet world'))
                .otherwise('')
```

```java
// "searched" form
builder.selectCase(label.equalTo('hello').then('world'),
                   label.equalTo('goodbye').then('sweet world'))
                  .otherwise('')
```

But compared to the following, I'm not clear that it's really an improvement:

```java
// simple form
label.selectCase()
        .when('hello', 'world')
        .when('goodbye', 'sweet world')
        .otherwise('')
```

```java
// "searched" form
builder.selectCase(String.class)
        .when(label.equalTo('hello'), 'world')
        .when(label.equalTo('goodbye', 'sweet world')
        .otherwise('')
```

So, while I don't find the use of the builder pattern here to be especially elegant or intuitive, the resulting code is materially less verbose.

gavinking:
Merging this so that it can be rolled into the release.

If necessary we can argue about it further later.

PR #1047, opened 2026-05-11
@Fetch

### LINKED ISSUE #1044: alternative to new approach to named entity graphs
I know it might seem a bit late in the day to be having second thoughts about this, but I would rather have second thoughts now than after JPA 4 is final.

Here is the approach we currently have:

```java
 @NamedEntityGraph(name = "EmployeeWithProjectTasksAndEmployer")
 @Entity // root entity of graph
 public class Employee {
     ...
     // fetched attribute node
     @NamedEntityGraph.AttributeNode(graph = "EmployeeWithProjectTasksAndEmployer")
     @ManyToOne(fetch=LAZY) Employer employer;

     // reference to subgraph defined in Project class
     @NamedEntityGraph.Subgraph(graph = "EmployeeWithProjectTasksAndEmployer")
     @ManyToMany Set<Project> projects;
 }

 @NamedEntityGraph(name = "EmployeeWithProjectTasksAndEmployer")
 @Entity // root entity of subgraph
 public class Project {
     ...
     // reference to subgraph defined in Task class
     @NamedEntityGraph.Subgraph(graph = "EmployeeWithProjectTasksAndEmployer")
     @OneToMany List<Task> tasks;
 }
```

It's ... well, very reasonable, but quite verbose, and it required the introduction of two new annotations which fit sortof uncomfortably with the pre-existing `@NamedAttributeNode` and `@NamedSubgraph`.

I'm wondering if we could do something simpler; something a bit closer to the original design of Hibernate "fetch profiles" which inspired this stuff. Consider:

```java
 @NamedEntityGraph(name = "EmployeeWithProjectTasksAndEmployer")
 @Entity // root entity of graph
 public class Employee {
     ...
     @ManyToOne(fetch=LAZY, 
                graphs="EmployeeWithProjectTasksAndEmployer") 
     Employer employer;

     // reference to subgraph defined in Project class
     @ManyToMany(graphs = "EmployeeWithProjectTasksAndEmployer") 
     Set<Project> projects;
 }

 @Entity // root entity of subgraph
 public class Project {
     ...
     // reference to subgraph defined in Task class
     @OneToMany(graphs = "EmployeeWithProjectTasksAndEmployer") 
     List<Task> tasks;
 }
```

Compared with the current design:

1. this is clearly less verbose
2. it doesn't require new annotations, though it does require adding `String[] graphs default {}` to `@Basic`, `@ElementCollection`, and to all the relationship-defining annotations. (This is reasonable, since historically they have a role in fetching.)
3. subgraphs are implicit, whereas in the previous design they were explicitly declared; this means somewhat more work on the part of the provider to piece them together
4. a subgraph can't be reused across different graphs (this was never an important goal)
5. a given entity cannot have two subgraphs defined within a given graph, depending on how it was reached from the root entity of the graph (this _was_ a goal, but it's probably more of a nice-to-have than a have-to-have)
6. there's no place to add fetch options like batch size, cache modes, and so on; we don't currently offer that on `NamedEntityGraph.AttributeNode`, but it's an obvious thing to consider adding [In fact, if we do stick with the current design, we should add that!]

An "intermediate solution" could be something like this:

```java
 @NamedEntityGraph(name = "EmployeeWithProjectTasksAndEmployer")
 @Entity // root entity of graph
 public class Employee {
     ...
     @ManyToOne(fetch=LAZY)
     @Fetch(graph="EmployeeWithProjectTasksAndEmployer",
            cacheStoreMode=BYPASS, batchSize = 10)
     Employer employer;

     // reference to subgraph defined in Project class
     @ManyToMany
     @Fetch(graph="EmployeeWithProjectTasksAndEmployer"
            subgraph="EmployeeWithProjectTasksAndEmployer") 
     Set<Project> projects;
 }

 @NamedEntityGraph(name = "EmployeeWithProjectTasksAndEmployer")
 @Entity // root entity of subgraph
 public class Project {
     ...
     // reference to subgraph defined in Task class
     @OneToMany
     @Fetch(graph="EmployeeWithProjectTasksAndEmployer")
     List<Task> tasks;
 }
```

Here we've squeezed `@NamedEntityGraph.Subgraph` and `@NamedEntityGraph.AttributeNode` into the one `@Fetch` annotation. This strikes me as perfectly reasonable.

And in principle `@Fetch` could be used even without an entity graph.

```java
@ManyToOne
@Fetch(type=LAZY,
       cacheStoreMode=BYPASS, batchSize = 10)
Employer employer;
```

Anyway, I wanted to throw this out there to see what you all think.

gavinking:
rebase of #1046.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Fetch.java
+++ b/api/src/main/java/jakarta/persistence/Fetch.java
@@ -25,12 +25,21 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
  * Specifies options influencing how an attribute is fetched.
- * When a {@linkplain #graph} is specified, the options apply
- * when the attribute is fetched using the entity graph with
- * the specified name and the named graph contains an attribute
- * node representing the annotated attribute. Otherwise, when
- * no graph is specified, the options apply when the attribute
- * is fetched without the use of an entity graph.
+ * <ul>
+ * <li>By default, the options apply when not overridden by
+ *     an {@linkplain EntityGraph entity graph}. For a load
+ *     graph, options specified via the {@code @Fetch}
+ *     annotation are overridden when the graph has an
+ *     attribute node representing the annotated attribute.
+ *     For a fetch graph, options specified by {@code @Fetch}
+ *     are always considered overridden.
+ * <li>When a {@linkplain #graph graph name} is specified,
+ *     the options apply when the attribute is fetched using
+ *     the {@linkplain NamedEntityGraph named entity graph}
+ *     with the specified name and the named graph contains
+ *     an {@linkplain AttributeNode attribute node}
+ *     representing the annotated attribute.
+ * </ul>
  * <p>
  * In this example, the {@code @Fetch} annotation is equivalent
  * to having specified {@code @ManyToMany(fetch=EAGER)}:
@@ -49,7 +58,7 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * }
  * <p>
  * In this example, the options only apply when the named entity
- * graph is used:
+ * graph {@code BooksWithAuthors} is used:
  * {@snippet :
  * @Fetch(cacheStoreMode=BYPASS, batchSize = 10,
  *        graph = "BooksWithAuthors")
@@ -66,6 +75,8 @@ import static java.lang.annotation.RetentionPolicy.RUNTIME;
  * @ManyToMany
  * List<Author> authors;
  * }
+ * At most one such {@code @Fetch} annotation may have a missing
+ * {@link #graph} name.
  *
  * @see NamedEntityGraph
  *
@@ -86,6 +97,9 @@ public @interface Fetch {
      *     the specified name. The named graph contains an
      *     attribute node representing the annotated attribute.
      * </ul>
+     * <p>
+     * At most one {@code @Fetch} annotation of a given field or
+     * property may have a missing {@code graph} name.
      */
     String graph() default "";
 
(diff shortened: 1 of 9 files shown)


api/src/main/java/jakarta/persistence/Fetch.java
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * Specifies whether the persistence provider is permitted
+     * to real the associated entity data from the cache.
+     */
+    CacheRetrieveMode cacheRetrieveMode() default CacheRetrieveMode.USE;
+
+    // not currently fetch options,
+    // but perhaps they should be?
+
+//    int timeout() default -1;
```
sebersole: This can feasibly only be an option if the fetch is a "subsequent select" style fetch.  It makes no sense for join fetches e.g..  

I don't think making this a FetchOption makes sense.

api/src/main/java/jakarta/persistence/Fetch.java
```diff
@@ -0,0 +1,148 @@
...
+
+//    LockModeType lockMode() default LockModeType.NONE;
```
sebersole: Similarly, this is only a viable option in a certain subset of cases.  Not sure I'd make it a FetchOption, though so long as we define it using the "hint" wording, I won't complain.

sebersole (commented):
I'm fine with this.  Left a few comments inline.

api/src/main/java/jakarta/persistence/Fetch.java
gavinking: OK, thanks, it was just a thought.

api/src/main/java/jakarta/persistence/Fetch.java
gavinking: We could also say that the provider is permitted to throw if it can't honor the specified lock mode for whatever reason.

gavinking:
I discussed the again with @sebersole and we think this is a reasonable simplification compared to the previous iteration.

PR #1049, opened 2026-05-11
Document requirement to update version value of entity

njr-11:
Jakarta Data repositories that use EntityAgent need to know whether the version is automatically updated when invoking update/upsert/insert methods because the contract for the respective Jakarta Data `@Update`/`@Save`/`@Insert` repository methods requires returning an entity with the version updated.  The most convenient behavior, and what Hibernate appears to already be doing in its 8.0.0.0-Alpha1, is to update the version on the given entity.  This PR adds this as a requirement to Javadoc.

It was later identified that the Jakarta Persistence spec already documents the requirement, so this PR has been repurposed to also mention it in the Javadoc that application developers are more likely to see.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Version.java
+++ b/api/src/main/java/jakarta/persistence/Version.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2008, 2023 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2008, 2026 Oracle and/or its affiliates. All rights reserved.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


gavinking:
I'm not sure this is necessary. We don't have any similar javadoc on `EntityManager`.

And the way this is defined is more abstract that this. It's defined in terms of phenomena which are allowed/disallowed depending on the lock mode. We never say "the provider must update the version at this moment"; we say "the provider must ensure that these phenomena can't occur".

gavinking:
Well, excuse me, actually we do say this:

> The entity version must be updated by the persistence provider each
time the state of an entity instance is written to the database.
Furthermore, if the current persistence context contains a revision
of the entity instance when the instance is written to the database,
the persistence provider must verify that the revision held in the
persistence context is identical to the revision held in the database
by comparing the versions held in memory and in the database.
If the versions do not match, the persistence provider must throw an 
`OptimisticLockException`.

If you like, I can copy that passage into the javadoc of `@Version`, where I think it makes sense. I would prefer to have it there than on operations of EM/EA.

njr-11:
The text you found in the Persistence spec adequately covers the requirement for Jakarta Data providers, so the real purpose of this PR is resolved.  Good idea including documentation in the `Version` annotation Javadoc.  That will be helpful for end users who have the same question/requirement about behavior.  It looks like `Version` already covers the `OptimisticLockException` piece and does so in a way that applies regardless of whether there is a persistence context, so all that's missing is the first sentence about the updating the version value on the entity instance.  I switched the commit under this PR to add that sentence to `@Version`.

gavinking:
@njr-11 How about this: #1050

gavinking:
https://github.com/jakartaee/persistence/pull/1050 was merged. Closing.

PR #1050, opened 2026-05-12
more javadoc about @Version for @njr-11

gavinking:


njr-11 (approved):
This is fine in place of #1049 if you prefer to include the additional text from the specification.

gavinking:
> if you prefer to include the additional text from the specification.

I think it's useful information.

PR #1057, opened 2026-05-15
more complete TCK coverage of specialized criteria types

gavinking:


tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/jpa40/criteria/specialized/Client.java
```diff
@@ -137,6 +169,120 @@ public void specializedCriteriaFluentExpressionOperationsTest() {
...
+                        price.power(2).equalTo(16.0),
+                        price.round(0).equalTo(4.0),
+                        quantity.toLong().equalTo(10L),
+                        quantity.toInteger().equalTo(10),
+                        quantity.toFloat().between(9.9F, 10.1F),
+                        quantity.toDouble().between(9.9D, 10.1D),
+                        quantity.toBigDecimal().equalTo(BigDecimal.TEN),
+                        quantity.toBigInteger().equalTo(BigInteger.TEN),
+                        quantity.coalesce(0).equalTo(10),
+                        quantity.nullif(0).equalTo(10));
```
marko-bekhta, line 239: I wonder if we should do one predicate per query ... 🤔 that would be a bit tedious ... but 
Assuming we only have the string equality one implemented (`title.equalTo("Alpha")`), and all others just return `true`, the test would still pass, right?

If we go this route, then maybe we can do some parameterized test 🤔 

other than this, all the tests in this class are passing and I cannot see anything else that stands out 👍🏻

gavinking: I guess it depends what you think the purpose of a TCK is. To me it's to validate that product supports the specified APIs with and complies with specific assertions about behavior. It's not to be a full test suite which guarantees that there are no bugs in the implementation.

gavinking: @marko-bekhta I think I'm just going to merge it like this, otherwise it just gets too messy. It's good enough for now.

PR #1059, opened 2026-05-15
add TCK tests for new operations of StoredProcedureQuery

gavinking:


marko-bekhta (approved):
I can see some of the tests failing, but I think those are bugs or not implemented features:

* convertedStoredProcedureParameterTest - `java.lang.UnsupportedOperationException` (so all good)
* typedStoredProcedureResultClassTest (bug in ORM,  get single result resturns an array, so comparing with an object doesn't work but the data is ok. we should fix it on the ORM side)
* closeStoredProcedureQueryTest - bug in ORM we call `getOutputs().release();` but we haven't registered any in the test, while the procedure actually has some ...) but I'm guessing we shouldn't ...
* registerResultParameterTest - `java.lang.UnsupportedOperationException` (so all good)
* typedStoredProcedureResultSetMappingTest - same issue in ORM: we are trying to compare an object to an array
* typedNamedStoredProcedureParameterApiTest- `java.lang.UnsupportedOperationException` (so all good)

gavinking:
OK, let's pull the trigger on this. If there's things wrong here we'll figure it out.

PR #1063, opened 2026-05-15
rework semantics of EntityExistsException

gavinking:
The provider doesn't always know exactly why the insert failed, and so it can't be sure an `EntityExistsException` is appropriate.

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityAgent.java
+++ b/api/src/main/java/jakarta/persistence/EntityAgent.java
@@ -107,6 +107,11 @@ import java.util.function.Function;
  * and since the entities themselves are detached, modifications to
  * the entities are never automatically detected and made persistent.
  *
+ * <p>An exception thrown by a method of an {@code EntityAgent},
+ * or by a method of an instance of {@link Query} obtained from
+ * an {@code EntityAgent}, is considered recoverable and does not
+ * usually cause the active transaction to be marked for rollback.
+ *
  * @since 4.0
  *
  * @see EntityManager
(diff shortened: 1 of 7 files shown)


marko-bekhta (approved):

njr-11:
Do we have a similar problem in Jakarta Data's [Insert](https://jakarta.ee/specifications/data/1.0/apidocs/jakarta.data/jakarta/data/repository/insert), which is built on this?

> "If an entity of the given type, and with the same unique identifier already exists in the database when the annotated method is called, ... then the annotated method must raise [EntityExistsException](https://jakarta.ee/specifications/data/1.0/apidocs/jakarta.data/jakarta/data/exceptions/entityexistsexception)."

gavinking:
@njr-11 yup, looks like it.

gavinking:
@njr-11 I suppose one could argue that any sort of unique constraint violation can reasonably be interpreted as "the entity already exists", but at the very least the language needs to be loosened if that's how we're going to play this.

gavinking:
Actually I think that's the right thing to do: redefine both these exceptions to say they represent any kind of uniqueness violation that occurs as a result of an INSERT. That's probably more useful anyway, as well as being actually-implementable.

gavinking:
> Actually I think that's the right thing to do: redefine both these exceptions to say they represent any kind of uniqueness violation that occurs as a result of an INSERT. That's probably more useful anyway, as well as being actually-implementable.

I have made that change in this pull request.

njr-11 (approved):
Look good.  I agree the more general meaning for EntityExistsException is most consistent with how it is currently being used and is most useful to applications.

gavinking:
Perfect, so let's do this both here and in Jakarta Data.

PR #1066, opened 2026-05-16
consider dropping a minor requirement that's just unnecessary

gavinking:
We probably don't need to say that the result set mapping inferred from a `@NativeQuery` has a name or what the name is.

Note that we *do* want a third-party implementation of Jakarta Data to have the option of executing the `@NativeQuery` as a named query, so we do want to nail down the name of that query, but a named native query carries its result set mapping with it, and we don't need to say how that happens.


api/src/main/java/jakarta/persistence/query/NativeQuery.java
```diff
@@ -115,8 +115,7 @@
  * mapping} may be specified by annotating the method with one or more
  * {@link jakarta.persistence.EntityResult},
  * {@link jakarta.persistence.ColumnResult}, or
- * {@link jakarta.persistence.ConstructorResult} annotations. The inferred
- * result set mapping has the same name as the query.
```
gavinking, line 119: This is the significant change.

gavinking:
Going to just go ahead and do this one.

PR #1068, opened 2026-05-16
criteria-based manipulation of static and named queries

### LINKED ISSUE #1000: are Data repositories discoverable?
Currently we say:

<blockquote>
A class is considered <em>eligible</em> for generation of a metamodel class if
it **belongs to the persistence unit** and:

- it is annotated with a _discoverable annotation type_, or
- it has a method annotated with one of the two static query annotations,
  `@StaticQuery` or `@StaticNativeQuery`.
</blockquote>

And:

<blockquote>
The set of types **belonging to a persistence unit** is determined by:

- classes bearing discoverable annotation types located in the root of the
  persistence unit (unless `<exclude-unlisted-classes/>` is specified);
- one or more object/relational mapping XML files;
- one or more JAR files to be searched for classes;
- an explicit list of classes.
</blockquote>

So a Jakarta Data repository does not "belong to the persistence unit" unless it's explicitly listed in `persistence.xml` or in an `orm.xml`, and therefore we are not required to generate a static metamodel for it, meaning it's `@StaticQuery` and `@StaticNativeQuery` annotations have no representation in the metamodel.

One fix would be to simply assert that `@Repository` is considered `@Discoverable` ... even though it's not annotated as such.

Also I think we should add Jakarta Data's `@Query` to the list of "static query annotations".

### LINKED ISSUE #1067: criteria-based manipulation of named query
We have added two very useful things which allow Persistence-backed implementations of Jakarta Data:

1. `@JakartaQuery` + `@QueryOptions` backed by a named query
2. `CriteriaBuilder.createQuery(Class,String)` and friends

But we are still missing a bridge between these things. An implementation of Jakarta Data is going to want to:

1. obtain a `TypedQueryReference` from the EMF, and then
2. manipulate it as a `CriteriaQuery` before executing it.

but there's currently no good way to do that. 

Of course the repository can just directly executing compile the query string via `CriteriaBuilder` and execute the result as a completely dynamic query and that works, but it's inferior from several different points of view.

So we either need:

-  a way to get from `TypedQueryReference` -> `CriteriaQuery` and from `StatementReference` -> `CriteriaStatement` (which wouldn't be completely typesafe), or
- a way to obtain a `CriteraQuery`, `CriteraUpdate`, or `CriteriaDelete` just directly from a given name.

Probably the *simplest* solution, and the solution that's most convenient for the Data implementation, would be to add these methods to `EntityManagerFactory`:

```java
<R> Map<String, CriteriaQuery<R>> getNamedCriteriaQueries(Class<R> resultType);

Map<String, CriteriaStatement> getNamedCriteriaStatements();
```

but this still requires some further thought.

We need to do this for JPA 4.

gavinking:
This is needed for Jakarta Data; see #1067.

A couple of different ideas proposed here.

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
+++ b/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
@@ -28,9 +28,12 @@ import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Consumer;
 
 import jakarta.persistence.AttributeConverter;
+import jakarta.persistence.StatementReference;
 import jakarta.persistence.Tuple;
+import jakarta.persistence.TypedQueryReference;
 
 /**
  * Used to construct criteria queries, compound selections, 
@@ -90,6 +93,20 @@ public interface CriteriaBuilder {
      * @since 4.0
      */
     CriteriaQuery<?> createQuery(String jpql);
+
+    /**
+     * Create a {@link CriteriaQuery} object representing the
+     * named Jakarta Persistence query language {@code SELECT}
+     * query designated by the given reference.
+     * @param reference A reference to a Jakarta Persistence
+     *                  query language {@code SELECT} query
+     * @return criteria query object
+     * @throws IllegalArgumentException if the given reference
+     *         does not represent a named Jakarta Persistence
+     *         query language {@code SELECT} query
+     * @since 4.0
+     */
+    <T> CriteriaQuery<T> createQuery(TypedQueryReference<T> reference);
 
     /**
      * Create a {@link CriteriaQuery} object that returns a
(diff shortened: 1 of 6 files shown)


api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -180,6 +197,57 @@ public interface CriteriaBuilder {
...
+     * @param <T> the query result type
+     * @return a reference to the modified query
+     * @throws IllegalArgumentException if the given reference
+     *         does not represent a named Jakarta Persistence
+     *         query language {@code SELECT} query belonging
+     *         to the owning factory
+     * @since 4.0
+     */
+    <T> TypedQueryReference<T> augment(TypedQueryReference<T> reference,
+                                       Consumer<CriteriaQuery<T>> augmentation);
```
njr-11, line 204: The Javadoc is good but could use some clarification as to whether the returned `TypedQueryReference` is a new instance with the modifications vs being the same instance as the given `reference`.  The former would be useful for Jakarta Data repository methods with `Restriction` and `Order` parameters, whereas the usefulness of the latter would be limited to cases such as `@OrderBy` which applies across every usage of the method.

njr-11 (commented):
One concern I thought of while reviewing this is a potential lack of uniqueness of query names.

Currently the query name is defined as,

> the query is treated as a named query, where the query name is the concatenation of the unqualified name of the type, with the string ".", and the name of the annotated member, for example, "Library.findBooksByTitle"

In Jakarta Data, it is valid to reuse the same method name with different parameters. For example,

```java
@Repository
public interface Library extends DataRepository<Book, String> {
    @JakartaQuery("FROM Book WHERE isbn = ?1")
    Optional<Book> find(String isbn);

    @JakartaQuery("FROM Book WHERE title LIKE ?1 AND fiction = ?2")
    List<Book> find(String titlePattern, boolean fiction);
}
```

It will be unclear which query the following resolves to,

```java
TypedQueryReference<Book> queryRef =
        emf.getNamedQueries(Library.class).get("Library.find")
```

Another, more unlikely problem would be if there are two repository interfaces (in different packages) named Library that each have a find method.  That is also valid in Jakarta Data, although much less likely than the example of above.

We probably need a separate issue to look at whether these names need to be more unique.

gavinking:
> In Jakarta Data, it is valid to reuse the same method name with different parameters

Rrrrm yeah, good point. In our implementation we generate names involving the parameter types.  I guess we will have to add something like that to the spec.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
gavinking: It's definitely meant to return a different instance. I *thought* I had made that clear in the javadoc by saying "reference to the modified query" but now I see how that's still ambiguous.

gavinking:
> Rrrrm yeah, good point. In our implementation we generate names involving the parameter types. I guess we will have to add something like that to the spec.

An alternative approach would be to just say that Jakarta Data impl uses the static metamodel (i.e. the static methods returning `TypedQueryReference`). That works, because those metamodel refs are overloadable. But that does depend on the JPA static metamodel.

gavinking:
So @njr-11 which approach to you prefer? I guess I kinda like the `augment()` methods...

njr-11:
> I guess I kinda like the `augment()` methods...

I agree.  This seems like it will work well with the augment approach.

The other approach shown in #1067 ,
```java
<R> Map<String, CriteriaQuery<R>> getNamedCriteriaQueries(Class<R> resultType);
```
seems a little dangerous because Jakarta Data implementations will often be needing to modify the CriteriaQuery for `Restriction` and `Order` and will need to take care not to modify the CriteriaQuery instance obtained from the Map which would pollute subsequent usage.

gavinking:
> I agree. This seems like it will work well with the augment approach.

Alright, so let's go with that for now.

But I'm going to need you guys to try this out in your implementation.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
gavinking: fixed

gavinking:
@njr-11 please take a look now

spec/src/main/asciidoc/ch10-metadata-annotations.adoc
```diff
@@ -995,6 +1003,9 @@ public @interface JakartaQuery {
 The semantics of this annotation are defined by its Javadoc, and in
 <<canonical-metamodel>>.
 
+The persistence provider should treat the annotation
+`jakarta.data.repository.Query` as a synonym for `JakartaQuery`.
```
njr-11:
```suggestion
The persistence provider must treat the annotation
`jakarta.data.repository.Query` as a synonym for `JakartaQuery`.
```

njr-11 (approved):
Looks good. I added one suggestion for wording.

gavinking:
I'm merging the `augment()` variation of this idea so that everyone can try it out in their implementations (on both the Persistence and Data sides).

PR #1072, opened 2026-05-20
Nullability annotations

### LINKED ISSUE #725: use jakarta.annotation to document nullability of APIs

gavinking:
See https://github.com/jakartaee/persistence/issues/725

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityHandler.java
+++ b/api/src/main/java/jakarta/persistence/EntityHandler.java
@@ -16,6 +16,8 @@
 
 package jakarta.persistence;
 
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
 import jakarta.persistence.criteria.CriteriaBuilder;
 import jakarta.persistence.criteria.CriteriaQuery;
 import jakarta.persistence.criteria.CriteriaSelect;
@@ -125,7 +127,8 @@ public interface EntityHandler extends AutoCloseable {
      *
      * @since 4.0
      */
-    <T> T get(Class<T> entityClass, Object id);
+    @Nonnull
+    <T> T get(@Nonnull Class<T> entityClass, @Nonnull Object id);
 
     /**
      * Retrieve an entity representing the record with the
@@ -179,7 +182,10 @@ public interface EntityHandler extends AutoCloseable {
      *
      * @since 4.0
      */
-    <T> T get(Class<T> entityClass, Object id, FindOption... options);
+    @Nonnull
+    <T> T get(@Nonnull Class<T> entityClass,
+              @Nonnull Object id,
+              @Nullable FindOption... options);
 
     /**
      * Retrieve an entity representing the record with the
@@ -235,7 +241,10 @@ public interface EntityHandler extends AutoCloseable {
      *
      * @since 4.0
      */
-    <T> T get(EntityGraph<T> graph, Object id, FindOption... options);
+    @Nonnull
+    <T> T get(@Nonnull EntityGraph<T> graph,
+              @Nonnull Object id,
+              @Nullable FindOption... options);
 
     /**
      * Retrieve entity instances representing the records
@@ -256,43 +265,42 @@ public interface EntityHandler extends AutoCloseable {
      * may not be observed.
      *
      * @param entityClass The class of the entity to retrieve
-     * @param ids The identifiers of the entities to retrieve
-     * @param options Standard and vendor-specific options
-     * @param <T> The entity type
+     * @param ids         The identifiers of the entities to retrieve
+     * @param options     Standard and vendor-specific options
+     * @param <T>         The entity type
      * @return an ordered list of entity instances
-     *
-     * @throws IllegalArgumentException if the given class is
-     *         not an entity class belonging to the persistence
-     *         unit, or if one of the given identifiers is not
-     *         a non-null instance of the identifier type of
-     *         the given entity class
+     * @throws IllegalArgumentException     if the given class is
+     *                                      not an entity class belonging to the persistence
+     *                                      unit, or if one of the given identifiers is not
+     *                                      a non-null instance of the identifier type of
+     *                                      the given entity class
      * @throws TransactionRequiredException if any lock mode
-     *         other than {@link LockModeType#NONE NONE} is
-     *         specified and there is no transaction
-     *         associated with this handler
-     * @throws OptimisticLockException if an optimistic version
-     *         check fails
-     * @throws PessimisticLockException if a pessimistic lock
-     *         could not be obtained and the transaction is
-     *         rolled back
-     * @throws LockTimeoutException if a pessimistic lock
-     *         could not be obtained and only the statement
-     *         is rolled back
-     * @throws EntityNotFoundException if no record exists in
-     *         the database for one of the given identifiers
-     * @throws PersistenceException if a given
-     *         {@linkplain LockModeType lock mode type} is not
-     *         supported for the given entity class or if a
-     *         record could not be read from the database
-     *
+     *                                      other than {@link LockModeType#NONE NONE} is
+     *                                      specified and there is no transaction
+     *                                      associated with this handler
+     * @throws OptimisticLockException      if an optimistic version
+     *                                      check fails
+     * @throws PessimisticLockException     if a pessimistic lock
+     *                                      could not be obtained and the transaction is
+     *                                      rolled back
+     * @throws LockTimeoutException         if a pessimistic lock
+     *                                      could not be obtained and only the statement
+     *                                      is rolled back
+     * @throws EntityNotFoundException      if no record exists in
+     *                                      the database for one of the given identifiers
+     * @throws PersistenceException         if a given
+     *                                      {@linkplain LockModeType lock mode type} is not
+     *                                      supported for the given entity class or if a
+     *                                      record could not be read from the database
      * @apiNote Conceptually similar to {@linkplain #findMultiple},
      * except this form throws {@linkplain EntityNotFoundException}
      * instead of returning nulls.
-     *
      * @since 4.0
      */
-    <T> List<T> getMultiple(Class<T> entityClass, List<?> ids,
-                            FindOption... options);
+    @Nonnull
+    <T> List<T> getMultiple(@Nonnull Class<T> entityClass,
+                            @Nonnull List<?> ids,
+                            @Nullable FindOption... options);
 
     /**
      * Retrieve entity instances representing the records with
@@ -312,44 +320,43 @@ public interface EntityHandler extends AutoCloseable {
      * mechanisms used by the provider, this option may or
      * may not be observed.
      *
-     * @param graph The {@linkplain EntityGraph load graph}
-     * @param ids The identifiers of the entities to retrieve
+     * @param graph   The {@linkplain EntityGraph load graph}
+     * @param ids     The identifiers of the entities to retrieve
      * @param options Standard and vendor-specific options
-     * @param <T> The entity type
+     * @param <T>     The entity type
      * @return an ordered list of entity instances
-     *
-     * @throws IllegalArgumentException if the given class is
-     *         not an entity class belonging to the persistence
-     *         unit, or if one of the given identifiers is not
-     *         a non-null instance of the identifier type of
-     *         the given entity class
+     * @throws IllegalArgumentException     if the given class is
+     *                                      not an entity class belonging to the persistence
+     *                                      unit, or if one of the given identifiers is not
+     *                                      a non-null instance of the identifier type of
+     *                                      the given entity class
      * @throws TransactionRequiredException if any lock mode
-     *         other than {@link LockModeType#NONE NONE} is
-     *         specified and there is no transaction
-     *         associated with this handler
-     * @throws OptimisticLockException if an optimistic version
-     *         check fails
-     * @throws PessimisticLockException if a pessimistic lock
-     *         could not be obtained and the transaction is
-     *         rolled back
-     * @throws LockTimeoutException if a pessimistic lock
-     *         could not be obtained and only the statement
-     *         is rolled back
-     * @throws EntityNotFoundException if no record exists in
-     *         the database for one of the given identifiers
-     * @throws PersistenceException if a given
-     *         {@linkplain LockModeType lock mode type} is not
-     *         supported for the given entity class or if a
-     *         record could not be read from the database
-     *
+     *                                      other than {@link LockModeType#NONE NONE} is
+     *                                      specified and there is no transaction
+     *                                      associated with this handler
+     * @throws OptimisticLockException      if an optimistic version
+     *                                      check fails
+     * @throws PessimisticLockException     if a pessimistic lock
+     *                                      could not be obtained and the transaction is
+     *                                      rolled back
+     * @throws LockTimeoutException         if a pessimistic lock
+     *                                      could not be obtained and only the statement
+     *                                      is rolled back
+     * @throws EntityNotFoundException      if no record exists in
+     *                                      the database for one of the given identifiers
+     * @throws PersistenceException         if a given
+     *                                      {@linkplain LockModeType lock mode type} is not
+     *                                      supported for the given entity class or if a
+     *                                      record could not be read from the database
      * @apiNote Conceptually similar to {@linkplain #findMultiple},
      * except this form throws {@linkplain EntityNotFoundException}
      * instead of returning nulls.
-     *
      * @since 4.0
      */
-    <T> List<T> getMultiple(EntityGraph<T> graph, List<?> ids,
-                            FindOption... options);
+    @Nonnull
+    <T> List<T> getMultiple(@Nonnull EntityGraph<T> graph,
+                            @Nonnull List<?> ids,
+                            @Nullable FindOption... options);
 
     /**
      * Retrieve an entity representing the record with the
@@ -373,7 +380,8 @@ public interface EntityHandler extends AutoCloseable {
      *
      * @since 1.0
      */
-    <T> T find(Class<T> entityClass, Object id);
+    @Nullable
+    <T> T find(@Nonnull Class<T> entityClass, @Nonnull Object id);
 
     /**
      * Retrieve an entity representing the record with the
@@ -424,7 +432,10 @@ public interface EntityHandler extends AutoCloseable {
      *
      * @since 3.2
      */
-    <T> T find(Class<T> entityClass, Object id, FindOption... options);
+    @Nullable
+    <T> T find(@Nonnull Class<T> entityClass,
+               @Nonnull Object id,
+               @Nullable FindOption... options);
 
     /**
      * Retrieve an instance of the root entity of the given
@@ -477,7 +488,10 @@ public interface EntityHandler extends AutoCloseable {
      *
      * @since 3.2
      */
-    <T> T find(EntityGraph<T> graph, Object id, FindOption... options);
+    @Nullable
+    <T> T find(@Nonnull EntityGraph<T> graph,
+               @Nonnull Object id,
+               @Nullable FindOption... options);
 
     /**
      * Retrieve entity instances representing the records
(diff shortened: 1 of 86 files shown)


api/src/main/java/jakarta/persistence/spi/ClassTransformer.java
```diff
@@ -55,10 +58,11 @@ public interface ClassTransformer {
      * @throws TransformerException  if the input does
      *         not represent a well-formed class file
      */
-    byte[] transform(ClassLoader loader,
-                     String className,
-                     Class<?> classBeingRedefined,
-                     ProtectionDomain protectionDomain, 
-                     byte[] classfileBuffer) 
+    @Nonnull
```
marko-bekhta:
```suggestion
    @Nullable
```

> or null if no transform is performed 

🙈

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
```diff
@@ -55,6 +58,7 @@ public interface PersistenceUnitInfo {
      * @return  the fully qualified name of the persistence provider 
      * implementation class
      */
+    @Nullable
```
marko-bekhta, line 61: javadoc doesn't mention when it can be null 🤔 ?

api/src/main/java/jakarta/persistence/criteria/TemporalExpression.java
```diff
@@ -16,6 +16,8 @@
 
 package jakarta.persistence.criteria;
 
+import jakarta.annotation.Nonnull;
+import jakarta.annotation.Nullable;
```
marko-bekhta: seems unused here

api/src/main/java/jakarta/persistence/EntityHandler.java
```diff
@@ -256,43 +265,42 @@ public interface EntityHandler extends AutoCloseable {
...
      * instead of returning nulls.
-     *
      * @since 4.0
      */
-    <T> List<T> getMultiple(Class<T> entityClass, List<?> ids,
-                            FindOption... options);
+    @Nonnull
+    <T> List<T> getMultiple(@Nonnull Class<T> entityClass,
+                            @Nonnull List<?> ids,
+                            @Nullable FindOption... options);
```
marko-bekhta, line 303: These nullables on `Smth...` varargs feel confusing 
this is about null find options in the array, right ? not the options array being null itself ... 
and then do we want null find options ? (probably the same question to other varargs)

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -695,6 +740,7 @@ StoredProcedureQuery registerStoredProcedureParameter(
      *         is rolled back
      */
     @Override @SuppressWarnings("removal")
+    @Nonnull
```
marko-bekhta: ah, I think all getresult variations here can be null (and lists too, according to the javadoc 😖 )?

marko-bekhta (commented):
I haven't checked each annotation 🙈 those I did make sense 🙂 👍🏻  (at least most of them... added some comments)

api/src/main/java/jakarta/persistence/spi/ClassTransformer.java
gavinking: Thanks, good catch.

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
gavinking: Oh, sure, right you are.

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
gavinking: No, but it can be. You don't need to explicitly specify a provider.

api/src/main/java/jakarta/persistence/EntityHandler.java
gavinking: I'm saying that you can explicitly pass in a null array if you like.

api/src/main/java/jakarta/persistence/criteria/TemporalExpression.java
gavinking: yeah I already caught that one.

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
gavinking: Fixed, thanks!

api/src/main/java/jakarta/persistence/EntityHandler.java
gavinking: Note: it's confusing because these are not `TYPE_USE` annotations.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
```diff
@@ -63,7 +67,8 @@ public interface AbstractQuery<T> extends CommonAbstractCriteria {
      * @param restriction  a simple or compound boolean expression
      * @return the modified query
      */    
-    AbstractQuery<T> where(Expression<Boolean> restriction);
+    @Nonnull
+    AbstractQuery<T> where(@Nonnull Expression<Boolean> restriction);
```
beikov, line 71:
```suggestion
    AbstractQuery<T> where(@Nullable Expression<Boolean> restriction);
```

I think that people are doing things like this in the wild, so it might be better to make the argument nullable.

```java
Predicate p = null;

if (...) {
  p = ...
}

query.where(p);
```

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
```diff
@@ -108,15 +116,17 @@ public interface AbstractQuery<T> extends CommonAbstractCriteria {
...
 
     /**
      * Specify a restriction over the groups of the query.
      * Replaces the previous having restriction(s), if any.
      * @param restriction  a simple or compound boolean expression
      * @return the modified query
      */
-    AbstractQuery<T> having(Expression<Boolean> restriction);
+    @Nonnull
+    AbstractQuery<T> having(@Nonnull Expression<Boolean> restriction);
```
beikov, line 129:
```suggestion
    AbstractQuery<T> having(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/CollectionJoin.java
```diff
@@ -41,7 +42,8 @@ public interface CollectionJoin<Z, E>
      * @return the modified join object
      * @since 2.1
      */
-    CollectionJoin<Z, E> on(Expression<Boolean> restriction);
+    @Nonnull
+    CollectionJoin<Z, E> on(@Nonnull Expression<Boolean> restriction);
```
beikov, line 46:
```suggestion
    CollectionJoin<Z, E> on(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/CriteriaDelete.java
```diff
@@ -39,7 +41,8 @@ public interface CriteriaDelete<T> extends CriteriaStatement<T> {
      * @param restriction  a simple or compound boolean expression
      * @return the modified delete query
      */
-    CriteriaDelete<T> where(Expression<Boolean> restriction);
+    @Nonnull
+    CriteriaDelete<T> where(@Nonnull Expression<Boolean> restriction);
```
beikov, line 45:
```suggestion
    CriteriaDelete<T> where(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/CriteriaQuery.java
```diff
@@ -187,7 +191,8 @@ public interface CriteriaQuery<T> extends AbstractQuery<T>, CriteriaSelect<T> {
      * @param restriction  a simple or compound boolean expression
      * @return the modified query
      */
-    CriteriaQuery<T> where(Expression<Boolean> restriction);
+    @Nonnull
+    CriteriaQuery<T> where(@Nonnull Expression<Boolean> restriction);
```
beikov, line 195:
```suggestion
    CriteriaQuery<T> where(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/CriteriaQuery.java
```diff
@@ -250,7 +259,8 @@ public interface CriteriaQuery<T> extends AbstractQuery<T>, CriteriaSelect<T> {
      * @param restriction  a simple or compound boolean expression
      * @return the modified query
      */
-    CriteriaQuery<T> having(Expression<Boolean> restriction);
+    @Nonnull
+    CriteriaQuery<T> having(@Nonnull Expression<Boolean> restriction);
```
beikov, line 263:
```suggestion
    CriteriaQuery<T> having(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/CriteriaUpdate.java
```diff
@@ -97,7 +108,8 @@ public interface CriteriaUpdate<T> extends CriteriaStatement<T> {
      * @param restriction  a simple or compound boolean expression
      * @return the modified update query
      */
-    CriteriaUpdate<T> where(Expression<Boolean> restriction);
+    @Nonnull
+    CriteriaUpdate<T> where(@Nonnull Expression<Boolean> restriction);
```
beikov, line 112:
```suggestion
    CriteriaUpdate<T> where(@Nullable Expression<Boolean> restriction);
```

gavinking:
I think I'm just going to merge this now; easy to fix mistakes later, including even in 4.0.1 or whatever.

api/src/main/java/jakarta/persistence/criteria/Join.java
```diff
@@ -36,7 +38,8 @@ public interface Join<Z, X> extends From<Z, X> {
      * @return the modified join object
      * @since 2.1
      */
-    Join<Z, X> on(Expression<Boolean> restriction);
+    @Nonnull
+    Join<Z, X> on(@Nonnull Expression<Boolean> restriction);
```
beikov, line 42:
```suggestion
    Join<Z, X> on(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/ListJoin.java
```diff
@@ -41,7 +42,8 @@ public interface ListJoin<Z, E>
      * @return the modified join object
      * @since 2.1
      */
-    ListJoin<Z, E> on(Expression<Boolean> restriction);
+    @Nonnull
+    ListJoin<Z, E> on(@Nonnull Expression<Boolean> restriction);
```
beikov, line 46:
```suggestion
    ListJoin<Z, E> on(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/MapJoin.java
```diff
@@ -41,7 +42,8 @@ public interface MapJoin<Z, K, V>
      * @return the modified join object
      * @since 2.1
      */
-    MapJoin<Z, K, V> on(Expression<Boolean> restriction);
+    @Nonnull
+    MapJoin<Z, K, V> on(@Nonnull Expression<Boolean> restriction);
```
beikov, line 46:
```suggestion
    MapJoin<Z, K, V> on(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/SetJoin.java
```diff
@@ -39,7 +40,8 @@ public interface SetJoin<Z, E> extends PluralJoin<Z, Set<E>, E> {
      * @return the modified join object
      * @since 2.1
      */
-    SetJoin<Z, E> on(Expression<Boolean> restriction);
+    @Nonnull
+    SetJoin<Z, E> on(@Nonnull Expression<Boolean> restriction);
```
beikov, line 44:
```suggestion
    SetJoin<Z, E> on(@Nullable Expression<Boolean> restriction);
```

api/src/main/java/jakarta/persistence/criteria/Subquery.java
```diff
@@ -53,7 +56,8 @@ public interface Subquery<T> extends AbstractQuery<T>, Expression<T> {
      * @param restriction  a simple or compound boolean expression
      * @return the modified subquery
      */
-    Subquery<T> where(Expression<Boolean> restriction);
+    @Nonnull
+    Subquery<T> where(@Nonnull Expression<Boolean> restriction);
```
beikov, line 60:
```suggestion
    Subquery<T> where(@Nullable Expression<Boolean> restriction);
```

marko-bekhta (approved):

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -179,7 +181,8 @@ public interface EntityManagerFactory extends AutoCloseable {
      * @throws IllegalStateException if the entity manager factory
      *         has been closed
      */
-    EntityManager createEntityManager(EntityManager.CreationOption... options);
+    @Nonnull
+    EntityManager createEntityManager(@Nullable EntityManager.CreationOption... options);
```
beikov, line 185: IMO, this should be an empty array if not create options should be passed, but never null or contain a null.

```suggestion
    EntityManager createEntityManager(@Nonnull EntityManager.CreationOption... options);
```

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
```diff
@@ -237,7 +242,8 @@ EntityManager createEntityManager(SynchronizationType synchronizationType,
      *
      * @since 4.0
      */
-    EntityAgent createEntityAgent(EntityAgent.CreationOption... options);
+    @Nonnull
+    EntityAgent createEntityAgent(@Nullable EntityAgent.CreationOption... options);
```
beikov, line 246: IMO, this should be an empty array if not create options should be passed, but never null or contain a null.

```suggestion
    EntityAgent createEntityAgent(@Nonnull EntityAgent.CreationOption... options);
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -674,6 +715,7 @@ StoredProcedureQuery registerStoredProcedureParameter(
      */
     @Override @SuppressWarnings({"rawtypes", "removal"})
     @Deprecated(since = "4.0", forRemoval = true)
+    @Nullable
```
beikov, line 718: From an inheritance POV I find this concerning. The super type says that the result is non-null, but here the subtype specifies that the result is nullable. This is not sound as a cast to the supertype would break the contract if suddenly a null is returned.
It's good that the method is deprecated, but I feel like the `Query#getResultList` method shouldn't specify `@Nonnull`, but rather specify `@Nullable` instead and let the subtypes override the nullability to `@Nonnull` where possible.

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -695,6 +737,7 @@ StoredProcedureQuery registerStoredProcedureParameter(
      *         is rolled back
      */
     @Override @SuppressWarnings("removal")
+    @Nullable
```
beikov, line 740: Similarly, this method has the same issue as `getResultList()`

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
beikov: I see that the varargs situation was discussed before with @marko-bekhta, but I don't understand why passing a null array should be ok. If you invoke it like `emf.createEntityManager()`, javac will always pass an empty array, so unless you're passing explicitly `null`, which is weird, you wouldn't normally encounter `null`.

gavinking: Java allows an array to be passed to a varargs. I'm saying that in this case we will treat a null array as if it were an empty array. Which is, it seems to me, an entirely sane thing to do.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
gavinking: Again, that's not what I want. There's no reason to blow up if we don't receive any options.

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
gavinking: OK, but that's just the legacy contract of this operation that you're objecting to. 

`StoredProcedureQuery` *does* break the contract of its supertype. `Query.getList()` was never documented to sometimes return null.

All I'm doing here is explicitly documenting something that was previously implicit.

>  feel like the `Query#getResultList` method shouldn't specify `@Nonnull`

Yeah, I mean, sure, you're right about that: if you thought that this design was _correct_, then that's what you would do. But actually I think the design is wrong and broken, and I'm just making the broken-ness explicit. I doubt that many people are executing this deprecated method of `StoredProcedureQuery` after widening the type to `Query`.

Anyway, both the operations are already marked `@Deprecated(forRemoval=true)`.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
gavinking: I don't think we should make that change.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
gavinking: I don't think we should encourage that. So I don't want to make that change.

api/src/main/java/jakarta/persistence/EntityManagerFactory.java
beikov: I prefer to reject `null` objects for container types that can be empty like arrays or collections, but if you think that passing a `null` array explicitly is something that people want or should do, then so be it.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
gavinking: Their workaround is:

```java
Predicate p = builder.conjunction();

if (...) {
  p = ...
}

query.where(p);
```

or even better to use the overload of `where()` which accepts a `List`. We added that in 3.2.

api/src/main/java/jakarta/persistence/criteria/CollectionJoin.java
gavinking: Ah. What we should do here is add an overload accepting `List`, like what we did with `where()` and `having()` in 3.2.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
beikov: So far, the implementation in Hibernate ORM supported null, so people surely rely on this. I'm just telling you that this would choke up user programs for no good reason IMO. Also, it's asymmetrical with what `getRestriction()` returns.

api/src/main/java/jakarta/persistence/criteria/CriteriaDelete.java
gavinking: Another missing `List` overload.

gavinking:
@beikov see #1075

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
gavinking: No, not anymore, you're looking at obsolete code. That was the issue Marko already spotted.

api/src/main/java/jakarta/persistence/criteria/AbstractQuery.java
gavinking: It will not break any user programs. At worst they will get a warning.

If the method were called `setRestriction()` I would agree with you. But it's called `where()`, and I have no clue what `where(null)` means.

beikov: >It will not break any user programs. At worst they will get a warning.

I would interpret these annotations as prescriptive for an implementor i.e. throw an error if a `null` is passed to a `@Nonnull` method argument, so users wouldn't just get a warning. Maybe your intention was to not influence implementors, but the temptation is high.

gavinking: No, a prescription looks like:

```java
* @throws IllegalArgumentException if the argument is null
```

there is no such prescription here.

PR #1074, opened 2026-05-21
delete CriteriaBuilder.keys(), CriteriaBuilder.values()

### LINKED ISSUE #1073: CriteriaBuilder.keys(), CriteriaBuilder.values()
I have no idea what the operations `keys()` and `values()` in `CriteriaBuilder` are supposed to do.

1. JPQL has no `keys()` or `values()` function
2. They are untested by the TCK and unimplemented in Hibernate
4. They accept a literal Java `Map` ... you could just call `map.keys()` or `map.values()`

There is a code comment (not javadoc!) saying:

```java
    //get the values and keys collections of the Map, which may then
    //be passed to size(), isMember(), isEmpty(), etc
```

But that makes no sense. You could call `map.keys.size()` or `map.keys().isEmpty()` in Java, or use `in()` instead of `isMember()`.

I think they're just a mistake and we should just delete them.

gavinking:
I don't like removing things without a deprecation cycle, but these operations are unclearly documented, apparently completely useless, untested in the TCK, unimplemented in Hibernate. JPQL does not have any equivalent operations.

### DIFF
--- a/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
+++ b/api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
@@ -1442,30 +1442,6 @@ public interface CriteriaBuilder {
                                                        @Nonnull Expression<C> collection);
 
 
-    //get the values and keys collections of the Map, which may then
-    //be passed to size(), isMember(), isEmpty(), etc
-
-    /**
-     * Create an expression that returns the values of a map.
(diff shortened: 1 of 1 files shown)


gavinking:
Does anyone object to what I'm proposing to do here?

anija-anil (approved):

gavinking:
I believe we have agreement to do this.

PR #1075, opened 2026-05-21
add additional overloads of operations accepting restriction lists

gavinking:
for consistency with what we already did to CriteriaQuery in 3.2

cc @beikov

(also add some missing `@Override` annotations)


njr-11 (approved):

gavinking:
Thanks Nathan. I'm going to merge this because it is a simple question of consistency with other similar operations in the criteria API. (And because the overloads are useful.)

PR #1081, opened 2026-05-23
clarify/tighten expectations around AttributeNode removal/addition

gavinking:
When I sat down to actually implement this and pass the TCK, I realized that the current language here wasn't really quite right in light of the subtle but important semantic differences between "load" and "fetch" graphs, and given that we don't know what sort of graph were are constructing when we are constructing it (a really crazy aspect of the design of JPA 2.1 is that there are two kinds of graphs with quite different semantics, but the graph itself doesn't know what kind of graph it is).

### DIFF
--- a/api/src/main/java/jakarta/persistence/Graph.java
+++ b/api/src/main/java/jakarta/persistence/Graph.java
@@ -94,9 +94,7 @@ public interface Graph<T> {
      *
      * <p>Added nodes are reflected in the list of
      * {@linkplain #getAttributeNodes child nodes} as instances of
-     * {@link AttributeNode} with option {@link FetchType#EAGER},
-     * except when the effect of this call is simply to cancel the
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/Graph.java
marko-bekhta: do we remove this as above ?

gavinking: Yes.

gavinking: Fixed.

gavinking:
@marko-bekhta Thanks. Check now.

marko-bekhta (approved):

gavinking:
Thanks, Marko.

I have decide to go ahead and merge this, so we can release m4 and continue making progress on the TCK and implementation side of all this.

PR #1084, opened 2026-05-25
add an @apiNote about query Options

gavinking:


api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -326,17 +325,26 @@ default Stream getResultStream() {
      * @param hintName The name of the property or hint
      * @param value The value for the property or hint
      * @return the same query instance
-     * @throws IllegalArgumentException if the second argument is not
-     *         valid for the implementation
+     * @throws IllegalArgumentException if the given property or
+     *         hint name is recognized by the provider, but the
+     *         second argument is not a legal value for the given
+     *         property or hint
```
marko-bekhta, line 331: statement/typedquery ... use a simple
> @throws IllegalArgumentException if the second argument is not
           valid for the implementation

maybe let's sync this part ?

marko-bekhta (approved):

api/src/main/java/jakarta/persistence/Query.java
gavinking: Good catch. Fixed.

PR #1088, opened 2026-05-26
Javadoc refers to setEntityGraph which does not exist

njr-11:
Javadoc for QueryOptions.entityGraph refers to a `setEntityGraph` method that does not exist.


gavinking:
Thanks Nathan.

PR #1090, opened 2026-05-26
Sort out the nullability of IdentifiableType.getIdType()

gavinking:
it can be null for a mapped superclass

### DIFF
--- a/api/src/main/java/jakarta/persistence/Graph.java
+++ b/api/src/main/java/jakarta/persistence/Graph.java
@@ -113,8 +113,6 @@ public interface Graph<T> {
      *
      * @param attributeName the name of an attribute of the managed type
      * @return true if there is an existing attribute node
-     * @throws IllegalArgumentException if the attribute is not an
-     *         attribute of this entity.
      * @throws IllegalArgumentException if there is no attribute
      *         with the given name
      *
@@ -133,6 +131,8 @@ public interface Graph<T> {
      *
      * @param attribute  attribute
      * @return true if there is an existing attribute node
+     * @throws IllegalArgumentException if the attribute is not an
+     *         attribute of this entity
      *
      * @since 3.2
      */
@@ -144,8 +144,6 @@ public interface Graph<T> {
      *
      * @param attributeName the name of an attribute of the managed type
      * @return the attribute node
-     * @throws IllegalArgumentException if the attribute is not an
-     *         attribute of this entity.
      * @throws java.util.NoSuchElementException if there is no existing
      *         node for the attribute
      * @throws IllegalArgumentException if there is no attribute
@@ -168,6 +166,8 @@ public interface Graph<T> {
      * @param attribute  attribute
      * @param <Y> the type of the attribute
      * @return the attribute node
+     * @throws IllegalArgumentException if the attribute is not an
+     *         attribute of this entity
      * @throws java.util.NoSuchElementException if there is no existing
      *         node for the attribute
      *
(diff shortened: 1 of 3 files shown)


gavinking:
I did not actually mean to include the changes to javadoc of `Graph` in this PR. But, well, they're there now...

api/src/main/java/jakarta/persistence/metamodel/IdentifiableType.java
```diff
@@ -94,36 +94,49 @@ public interface IdentifiableType<X> extends ManagedType<X> {
...
+    boolean hasId();
+
     /**
      * Whether the identifiable type has a single id attribute.
-     * Returns true for a simple id or embedded id; returns false
-     * for an idclass.
-     * @return boolean indicating whether the identifiable
-     *         type has a single id attribute
+     * Returns true for a {@linkplain jakarta.persistence.Id simple id}
+     * or {@linkplain jakarta.persistence.EmbeddedId embedded id}, or
```
marko-bekhta, line 111:
```suggestion
     * or {@linkplain jakarta.persistence.EmbeddedId embedded id}, and
```
maybe ?

api/src/main/java/jakarta/persistence/Graph.java
```diff
@@ -133,6 +131,8 @@ public interface Graph<T> {
      *
      * @param attribute  attribute
      * @return true if there is an existing attribute node
+     * @throws IllegalArgumentException if the attribute is not an
+     *         attribute of this entity
```
marko-bekhta, line 135: I can see an argument for throwing here ... but then to be consistent with the string method version, wouldn't just returning `false` be ok too ?

api/src/main/java/jakarta/persistence/metamodel/IdentifiableType.java
gavinking: Marko, I'm not sure I agree: it returns "true or false", not "true and false".

api/src/main/java/jakarta/persistence/Graph.java
gavinking: Well the string version also throws if there's no attribute with the given name.

api/src/main/java/jakarta/persistence/metamodel/IdentifiableType.java
marko-bekhta: ok 🙂 yeah... just a lot of "or"s in there 🫣 🙂

api/src/main/java/jakarta/persistence/Graph.java
marko-bekhta: ahh I saw that this was removed:

>     * @throws IllegalArgumentException if the attribute is not an
>     *         attribute of this entity.

but the diff wasn't showing the other `@throws` with the same `IllegalArgumentException`

api/src/main/java/jakarta/persistence/metamodel/IdentifiableType.java
gavinking: yeah I get that. I'm depending on the comma placement for clarity, which is not perfect but...

marko-bekhta (approved):

api/src/main/java/jakarta/persistence/Graph.java
gavinking: My *guess* is that what happened here was that someone pasted some of the `@throws IllegalArgumentException` onto the wrong members, and they were meant to go where I have them now.

gavinking:
Any objections to this anyone? @njr-11? @sebersole?

njr-11 (approved):

sebersole:
Nope, outside of a proper hierarchy model type this is needed.  +1

PR #1093, opened 2026-05-27
permit executing stored procedure with no tx context

### LINKED ISSUE #1092: executing stored procedure with no tx context
Some stored procedures do their own transaction management.

But the `executeUpdate()` method of `StoredProcedureQuery` appears to require that the provider throw `TransactionRequiredException` if you call it with no active tx. There's no mention of this in the spec, but there is a TCK test for it. I think that requirement is rubbish and should be trashed.

gavinking:
See #1092


marko-bekhta (approved):

marko-bekhta (approved):

gavinking:
Merging this so we can make progress.

PR #1102, opened 2026-05-29
add nullability constraints to Reference impls

gavinking:


api/src/main/java/jakarta/persistence/TypedQueryReference.java
```diff
@@ -103,6 +103,10 @@ public non-sealed interface TypedQueryReference<R> extends Reference {
      * from the declared return type of the method annotated
      * {@link jakarta.persistence.query.JakartaQuery} or
      * {@link jakarta.persistence.query.NativeQuery}.
+     *
+     * @return The result type as a Java {@link Class} object,
+     *         or {@code void.class} when the result type is
+     *         not explicitly specified and cannot be inferred.
```
njr-11, line 109: It looks like both NamedQuery and NamedNativeQuery default to `void.class` for `resultClass()`, so this looks good.

njr-11 (approved):

api/src/main/java/jakarta/persistence/TypedQueryReference.java
gavinking: Yep, that's why I went with that option.

PR #1109, opened 2026-06-02
make TypedQueryReference.getResultType() nullable

gavinking:
Using void.class turns out to make this harder not easier to implement.

This just rolls back a previous recent change.


gavinking:
Just going to merge this, since it just reverts a change; if anyone violently disagrees, speak up, so we can discuss it.

PR #1119, opened 2026-06-06
results of audit of API appendices

gavinking:
Result of a mechanical code review.


gavinking:
Nothing here that anyone might object to. Merging.

PR #1120, opened 2026-06-11
Consistently use EntityManagerFactory.getSchemaManager().truncate() for data cleanup between tests

marko-bekhta:
We were using this pattern of "within transaction execute a bunch of delete queries" to clean up data between tests. But then @sebersole pointed to exceptions logged in the execution that the table is missing for some of the tests ....

since in the past the schema was created upfront it wasn't an issue that we were just running delete statements, even if the entity wasn't involved in the current test...

With the change where tests create the schema themselves, it makes more sense to just go with `truncate()` ... 

(And as a "bonus", this actually uncovered a few more things in ORM)

### DIFF
--- a/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/jpa22/query/stream/Client.java
+++ b/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/jpa22/query/stream/Client.java
@@ -96,16 +96,16 @@ public class Client extends PMClientBase {
 		assertTrue(ids.size() == 21, "Unexpected result count:" + ids.size());
 		assertTrue(checkEntityPK(ids, expected, false), "Unexpected PKs received");
 
-		Stream<?> s1 = query.getResultStream();
-		try {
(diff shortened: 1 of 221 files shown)


tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/jpa22/query/stream/Client.java
```diff
@@ -96,16 +96,16 @@ public void getResultStreamTest() throws Exception {
 		assertTrue(ids.size() == 21, "Unexpected result count:" + ids.size());
 		assertTrue(checkEntityPK(ids, expected, false), "Unexpected PKs received");
 
-		Stream<?> s1 = query.getResultStream();
-		try {
+		try (Stream<?> s1 = query.getResultStream()) {
```
marko-bekhta, line 99: Note this additional change -- the queries were not running in a transaction, and at the same time, streams were not closed, which is "leaking connections", so now the streams are closed.

gavinking (approved):

PR #1122, opened 2026-06-11
add setParameters() as proposed in #1121

### LINKED ISSUE #1121: bulk setting parameters
In #717 (ultimately rejected) I complained about `setParameter()` being too verbose, but for some reason we didn't really discuss an obvious solution, which is to let you set multiple positional parameters all at once, for example:

```java
em.createQuery("from Book where title like ?1 and author.name like ?2"
    .ofType(Book.class)
    .setParameters(title, authorName)
    .getResultList();
```

Where `setParameters()` is a variadic method that sets positional parameters.

In principle we could have overloads accepting a `List`, or a `Map` of named parameter arguments, but the value of that would be very marginal, since you can already write:

```java
Map<String,Object> namedParamArgs = ... ;
Query query = ... ;
namedParamArgs.forEach(query::setParameter);
```

But the varargs version for positional parameters I think makes sense.

gavinking:
A very simple thing, but convenient, it seems to me.

### DIFF
--- a/api/src/main/java/jakarta/persistence/Query.java
+++ b/api/src/main/java/jakarta/persistence/Query.java
@@ -684,6 +684,22 @@ public interface Query {
                        @Nonnull TemporalType temporalType);
 
     /**
+     * Bind arguments to every positional parameter of the query.
+     * <p>
+     * The <em>n</em>th argument is bound to the positional
+     * parameter {@code ?n},
+     * @param arguments The arguments to positional parameters
+     * @return the same query instance
+     * @throws IllegalArgumentException if the number of arguments
+     *         is not exactly the same as the number of positional
+     *         parameters, or if one of the arguments is of an
+     *         incompatible type
+     * @since 4.0
+     */
+    @Nonnull
+    Query setParameters(@Nonnull Object... arguments);
+
+    /**
      * Get the {@link Parameter} objects representing the declared
      * parameters of the query or an empty set if the query has no
      * parameters.
--- a/api/src/main/java/jakarta/persistence/Query.java
+++ b/api/src/main/java/jakarta/persistence/Query.java
@@ -687,7 +687,7 @@ public interface Query {
      * Bind arguments to every positional parameter of the query.
      * <p>
      * The <em>n</em>th argument is bound to the positional
-     * parameter {@code ?n},
+     * parameter {@code ?n}.
      * @param arguments The arguments to positional parameters
      * @return the same query instance
      * @throws IllegalArgumentException if the number of arguments
--- a/api/src/main/java/jakarta/persistence/Statement.java
+++ b/api/src/main/java/jakarta/persistence/Statement.java
@@ -478,6 +478,23 @@ public interface Statement extends Query {
                            @Nonnull TemporalType temporalType);
 
     /**
+     * Bind arguments to every positional parameter of the statement.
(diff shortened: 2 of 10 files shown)


gavinking:
WDYT, @sebersole @njr-11 ?

api/src/main/java/jakarta/persistence/Query.java
```diff
@@ -676,6 +676,22 @@ Query setParameter(int position, @Nullable Calendar value,
     Query setParameter(int position, @Nullable  Date value,
                        @Nonnull TemporalType temporalType);
 
+    /**
+     * Bind arguments to every positional parameter of the query.
+     * <p>
+     * The <em>n</em>th argument is bound to the positional
+     * parameter {@code ?n},
```
njr-11:
```suggestion
     * parameter {@code ?n}.
```

api/src/main/java/jakarta/persistence/Statement.java
```diff
@@ -477,6 +477,23 @@ Statement setParameter(int position, @Nullable Calendar value,
     Statement setParameter(int position, @Nullable Date value,
                            @Nonnull TemporalType temporalType);
 
+    /**
+     * Bind arguments to every positional parameter of the statement.
+     * <p>
+     * The <em>n</em>th argument is bound to the positional
+     * parameter {@code ?n},
```
njr-11:
```suggestion
     * parameter {@code ?n}.
```

api/src/main/java/jakarta/persistence/StoredProcedureQuery.java
```diff
@@ -407,6 +407,24 @@ <P> StoredProcedureQuery setParameter(int position, @Nullable P value,
     <P> StoredProcedureQuery setConvertedParameter(int position, @Nullable P value,
                                                    @Nonnull Class<? extends AttributeConverter<P, ?>> converter);
 
+    /**
+     * Bind arguments to every positional parameter of the stored
+     * procedure.
+     * <p>
+     * The <em>n</em>th argument is bound to the positional
+     * parameter {@code ?n},
```
```suggestion
     * parameter {@code ?n}.
```

api/src/main/java/jakarta/persistence/TypedQuery.java
```diff
@@ -623,6 +623,23 @@ TypedQuery<X> setParameter(int position, @Nullable Calendar value,
     TypedQuery<X> setParameter(int position, @Nullable Date value,
                                @Nonnull TemporalType temporalType);
 
+    /**
+     * Bind arguments to every positional parameter of the query.
+     * <p>
+     * The <em>n</em>th argument is bound to the positional
+     * parameter {@code ?n},
```
```suggestion
     * parameter {@code ?n}.
```

njr-11 (commented):
Good idea - it will be convenient for users to be able to more concisely supply positional parameters this way.
I spotted a typo that is copied to several places.
This PR also includes a bunch of changes to a TCK class - just mentioning that in case it was included accidentally with this PR rather than another.

gavinking:
Thanks Nathan, good catch.

njr-11 (approved):

gavinking:
I have added a TCK test for `setParameters()`, and mentioned it in the relevant section of the spec.

In writing the TCK test, I noticed that there were some missing covariant overrides in `StatementOrTypedQuery`, and I added them. Lucky break to catch that.

PR #1127, opened 2026-06-12
Switch to TestInstance.Lifecycle.PER_CLASS / Fix flaky tests for date-time / remove individual cleanup methods

marko-bekhta:
The switch to `Lifecycle.PER_CLASS` removes the unnecessary factory setup + DDL work and that results in:
from
```
2248 tests
   in 310 test classes
   executed in 1 project
   in 2m 47.585s serial time
   and 2m 50.231s task execution time
```
to:
```
  2248 tests
   in 310 test classes
   executed in 1 project
   in 1m 3.965s serial time
   and 1m 6.150s task execution time
```

would be a bit slower on CIs, but should still improve execution time.

* There was a date-time test that was "execution-time-sensitive" in the sense that it was relying on setting up test data with  `.now()`, which might be unpredictable ... 
* A few test classes were silently failing data setup or checks -- made some changes there to just use asserts or to rely on DDL created by the provider before applying the stored procedures script -- there were some tests that were not creating necessary tables 😖 
* and last but not least ... removed the `(Logger) ` cast 🫣 

> [!IMPORTANT]
> Tests are now using per-class lifecycle, so it's a single test class instance to run all the test methods in that class, which means that if tests are relying on instance data, it has to be managed with ^ in mind.

### DIFF
--- a/tck/common/src/main/java/ee/jakarta/tck/persistence/common/DriverManagerConnection.java
+++ b/tck/common/src/main/java/ee/jakarta/tck/persistence/common/DriverManagerConnection.java
@@ -23,7 +23,7 @@ import java.util.Properties;
 
 public class DriverManagerConnection {
 
-	private static final Logger logger = (Logger) System.getLogger(DriverManagerConnection.class.getName());
+	private static final Logger logger = System.getLogger(DriverManagerConnection.class.getName());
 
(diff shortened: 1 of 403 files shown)


gavinking:
Thank you, Marko, that is great!!

PR #1130, opened 2026-06-15
rename BatchSize -> BatchFetch

### LINKED ISSUE #1129: rethink BatchSize
#948 introduced `BatchSize` with quite weak semantics. I now think this was a mistake.

As a `FindOption`, there's at least a couple of ways you could interpret it, but neither is completely obvious, and neither are mentioned in its javadoc. And its most useful application is probably as an option only to `findMultiple()` and `getMultiple()`.

I think we should:

- remove `FindOption` as a supertype, leaving it as a `FetchOption` only, where its semantics and usefulness are much clearer, and
- rename it to reflect its more specialized role. 

I'm thinking `BatchFetch` would be a reasonable name reflecting its more specialized application.

gavinking:
and remove its implementation of `FindOption`

see #1129


gavinking:
After discussing with @sebersole we have decided that this is the way to go.

PR #1142, opened 2026-06-27
introduce @PostCreate and @PreClose callbacks

### LINKED ISSUE #12: Lifecycle callbacks for EntityManagerFactory creation
Addition of preCreate/postCreate callbacks for `EntityManagerFactory`.

This would be useful hook for creation and/or modification of named queries.

gavinking:
See #12.

### DIFF
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -2153,6 +2153,35 @@ defined in the `SiameseCat` class:
 . `postPersistSiameseCatListenerMethod`
 . `postPersistAnimal`
 
+==== Persistence Unit Lifecycle Callbacks
+
+An entity listener class declared using the `@EntityListener` annotation
+may also have _persistence unit lifecycle_ callback methods. A persistence
+unit lifecycle callback must be annotated `@PostCreate` or `@PreClose`,
+and must have the signature:
+
+ void method(E argument)
+
+where `method` is a legal Java method name, and `E` is either
+`EntityManagerFactory`, `EntityManager`, or `EntityAgent`.
+
+- The `PostCreate` callback for an entity manager factory, entity manager,
+  or entity agent occurs after the factory, manager, or agent has been
+  created, but before a reference to the factory, manager, or agent is
+  returned to the application.
+- The `PreClose` callback for an entity manager factory, entity manager,
+  or entity agent occurs after the `close()` method of the factory, manager,
+  or agent has been called, but before the factory, manager, or agent is
+  actually destroyed.
+
+A persistence unit lifecycle callback is permitted to call any method of
+its argument. If the callback throws an exception, the behavior is undefined.
+An exception thrown by a `PostCreate` callback typically aborts creation of
+the factory, manager, or agent. An exception thrown by a `PreClose` callback
+typically does not abort destruction of the factory, manager, or agent.
+Portable applications should avoid throwing exceptions from persistence unit
+lifecycle callback methods.
+
 ==== Exceptions
 
 Lifecycle callback methods may throw runtime
--- a/spec/src/main/asciidoc/ch03-entity-operations.adoc
+++ b/spec/src/main/asciidoc/ch03-entity-operations.adoc
@@ -2225,12 +2225,13 @@ lifecycle callback methods as specified in annotations.
 
 ===== Specification of Callback Listeners
 
-The `entity-listener` XML descriptor element
-specifies the lifecycle listener methods of an entity listener
-class. The lifecycle listener methods are specified by using the
-`pre-persist`, `post-persist`, `pre-remove`, `post-remove`,`pre-merge`,
-`pre-insert`, `post-insert`, `pre-delete`, `post-delete`, `pre-upsert`,
-`post-upsert`, `pre-update`, `post-update`, and/or `post-load` elements.
+The `entity-listener` XML descriptor element declares the lifecycle
+callback methods of an entity listener class. Entity lifecycle callback
+methods are declared by using the elements `pre-persist`, `post-persist`,
+`pre-remove`, `post-remove`,`pre-merge`,`pre-insert`, `post-insert`,
+`pre-delete`, `post-delete`, `pre-upsert`, `post-upsert`, `pre-update`,
+`post-update`, and `post-load`. Persistence unit lifecycle methods are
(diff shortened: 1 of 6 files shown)


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2153,6 +2153,35 @@ defined in the `SiameseCat` class:
...
 . `postPersistAnimal`
 
+==== Persistence Unit Lifecycle Callbacks
+
+An entity listener class declared using the `@EntityListener` annotation
+may also have _persistence unit lifecycle_ callback methods. A persistence
+unit lifecycle callback must be annotated `@PostCreate` or `@PreClose`,
+and must have the signature:
+
+ void method(E argument)
```
marko-bekhta, line 2163: Should it be wrapped with ` ```java ``` ` ? Or do the spaces also work in Adoc, as in Markdown?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2153,6 +2153,35 @@ defined in the `SiameseCat` class:
 . `postPersistSiameseCatListenerMethod`
 . `postPersistAnimal`
 
+==== Persistence Unit Lifecycle Callbacks
+
+An entity listener class declared using the `@EntityListener` annotation
+may also have _persistence unit lifecycle_ callback methods. A persistence
+unit lifecycle callback must be annotated `@PostCreate` or `@PreClose`,
+and must have the signature:
```
marko-bekhta, line 2161: The fact that we allow these on the entity listener somehow doesn't sit right with me 🫣 🙂 
Maybe we should go with some dedicated `@PersistenceUnitListener` instead ?

Same about the event annotations ... we don't have granular packages, so they all end up mixed with entity-related ones 😕  but then I also cannot think of an alternative name that is not way too verbose 😖 

btw we probably should update the xsd (either `<xsd:complexType name="entity-listener">` or a new type if we end up not using the entity listener annotation ...)

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -2153,6 +2153,35 @@ defined in the `SiameseCat` class:
...
+  or agent has been called, but before the factory, manager, or agent is
+  actually destroyed.
+
+A persistence unit lifecycle callback is permitted to call any method of
+its argument. If the callback throws an exception, the behavior is undefined.
+An exception thrown by a `PostCreate` callback typically aborts creation of
+the factory, manager, or agent. An exception thrown by a `PreClose` callback
+typically does not abort destruction of the factory, manager, or agent.
+Portable applications should avoid throwing exceptions from persistence unit
+lifecycle callback methods.
```
marko-bekhta, line 2183: I was thinking about this case when a listener throws ... and it feels like these postcreate-predestroy should (almost?) always go in pair -- assuming multiple listeners are added post create goes one by one and fails somewhere mid iteration .. those that already got executed would assume they have a fully functional created factory, but in fact the factory is getting destroyed in the meantime ... 

I guess I'm probably missing what use cases we are trying to cover ... I'm thinking more about some sort of integrator where I'd get a reference to `E` and would "hold"to it till things get destroyed ...

marko-bekhta (commented):
+1 on the idea 🙂 👍🏻 but not sure about the "container" for these new events ...

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: The indent works. I'm just copying what is already earlier used in that section.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: So that's what I thought at first too ... but then I could not think of a single semantic difference between an "entity listener" class and a "persistence unit listener" class, and it just seemed completely redundant to have to specify the exact same stuff twice.

gavinking: > btw we probably should update the xsd

Fixed.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: Not sure what you're saying here Marko.

marko-bekhta: ah 🙂, I was trying to say that it seems that those (`@PostCreate` and `@PreClose`) almost always go in pairs.. so the observer interface (as we have in ORM) feels like a more natural fit, but IDK... I guess the user just needs to add two handlers

gavinking: Not necessarily. I think it's very common to have only a PostCreate. Anyway, this is consistent with the programming model used in other places in EE, in CDI, for example.

spec/src/main/asciidoc/ch03-entity-operations.adoc
njr-11: Should we add `post-create` and `pre-close` in this list?
```suggestion
class. The lifecycle listener methods are specified by using the
`pre-persist`, `post-persist`, `pre-remove`, `post-remove`, `pre-merge`,
`pre-insert`, `post-insert`, `pre-delete`, `post-delete`, `pre-upsert`,
`post-upsert`, `pre-update`, `post-update`, `post-load`, `post-create`,
and/or `pre-close` elements.
```

njr-11 (commented):
The changes look good, but there are some additional places that could be updated.  One is noted in a suggestion.  The other is a file not modified by the PR, EntityListener.java, which currently has a list of lifecycle events,
```
 * The annotated class may have callback methods annotated with
 * any of the standard lifecycle callback annotations:
 * <ul>
 * <li>{@link PostLoad},
 * <li>{@link PrePersist}, {@link PostPersist},
 *     {@link PreRemove}, {@link PostRemove}, and {@link PreMerge},
 * <li>{@link PreInsert}, {@link PostInsert}, {@link PreUpdate},
 *     {@link PostUpdate}, {@link PreUpsert}, {@link PostUpsert},
 *     {@link PreDelete}, and {@link PostDelete}.
 * </ul>
```

gavinking:
Thanks, @njr-11, I think my last commit addresses your comments.

njr-11 (approved):

PR #1144, opened 2026-06-28
relationship b/w Jakarta Data repositories and persistence units

### LINKED ISSUE #1143: Jakarta Data repository discovery
Currently we say:

> Exceptionally, the annotation `jakarta.data.repository.Repository` must be treated as a discoverable annotation type, even though it is not explicitly annotated `@Discoverable`.

I'm thinking that this is not quite enough, since it makes an implicit assumption that the Jakarta Data repository is deployed in the persistence unit, but application programs might want some other modular architecture.

I'm thinking it would be good to add some additional language that the EE container treats every Jakarta Data repository which depends on the persistence unit as a discoverable type of that persistence unit.

gavinking:
Here's something we could say. I'm not sure if it's a good idea or not.

### DIFF
--- a/spec/src/main/asciidoc/ch08-entity-packaging.adoc
+++ b/spec/src/main/asciidoc/ch08-entity-packaging.adoc
@@ -288,6 +288,14 @@ All classes must be on the classpath to ensure that entity managers from
 different persistence units that map the same class will be accessing the
 same identical class.
 
+In a Jakarta EE container, every Jakarta Data repository interface that
+depends on a persistence unit is considered to belong to the persistence
+unit, even if the repository interface is packaged in a separate archive.
+A repository depends on a persistence unit if its methods operate on
+entity types belonging to the persistence unit. When an entity type
+belongs to multiple persistence units, ambiguities are resolved via the
+`dataStore` member of `@Repository`.
+
 ====== Discoverable annotation types [[discoverable]]
 
 A _discoverable annotation type_ is an annotation type annotated with


njr-11 (approved):
The text looks fine if we decide to add this

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -288,6 +288,14 @@ All classes must be on the classpath to ensure that entity managers from
 different persistence units that map the same class will be accessing the
 same identical class.
 
+In a Jakarta EE container, every Jakarta Data repository interface that
+depends on a persistence unit is considered to belong to the persistence
+unit, even if the repository interface is packaged in a separate archive.
+A repository depends on a persistence unit if its methods operate on
+entity types belonging to the persistence unit. When an entity type
+belongs to multiple persistence units, ambiguities are resolved via the
+`dataStore` member of `@Repository`.
```
scottmarlow, line 297: > even if the repository interface is packaged in a separate archive.

If the entity classes are not accessible to the separate archive that might not work.

gavinking: Well the entity classes have to be accessible since the repository is defined in terms of them.

scottmarlow: Good point as clearly the application will not work otherwise.

scottmarlow: So if the persistence unit definition + entity classes are in a sub-deployment in an ear, other sub-deployment modules may or may not have access to the entity classes was what I was trying to say.

gavinking: Sure. Of course the persistence unit has to be a dependency of the module containing Jakarta Data repositories; that goes without saying.

The problem I'm trying to solve here is the opposite one: what if the module containing the Jakarta Data repositories is *not* a dependency of the persistence unit? How can the persistence provider get hold of the repository interfaces to scan them for `@JakartaQuery` and `@NativeQuery` annotations? My proposed solution is: pass the `Class` objects to the persistence provider.

gavinking:
> The text looks fine if we decide to add this

@njr-11 In Liberty, what are the restrictions on packaging of repositories with respect to persistence units?

spec/src/main/asciidoc/ch08-entity-packaging.adoc
scottmarlow: Will @JakartaQuery + @NativeQuery be tagged as [Discoverable?](https://github.com/jakartaee/persistence/blob/main/api/src/main/java/jakarta/persistence/spi/Discoverable.java)

I'm not sure if that helps but it might.

scottmarlow: https://github.com/jakartaee/persistence/issues/937 is the related issue for Discoverable.  

Sadly I think @Discoverable class scanning would be done by the EE container at the root of the persistence unit definition but repository interfaces might not be in that same root of the persistence unit location.

It sounds like the root deployment archive that contains repository interfaces need to be scanned for @Discoverable (@JakartaQuery and @NativeQuery which would need to be Discoverable) as well but accessible via a different PersistenceUnitInfo#getAllDiscoverableClassNames that would need to scan from the root of the deployment for all top level + sub deployment @Discoverable classes.  Would that help?

gavinking: > Sadly I think DISCOVERABLE class scanning would be done by the EE container at the root of the persistence unit definition but repository interfaces might not be in that same root of the persistence unit location.

Yes, that is certainly the intent.

What I'm trying to do here is add a separate, additional kind of discovery just for `@Repository` interfaces. I admit that it's an inelegant solution.

njr-11:
> > The text looks fine if we decide to add this
> 
> @njr-11 In Liberty, what are the restrictions on packaging of repositories with respect to persistence units?

Liberty does not require the class annotated Repository to be packaged with or even accessible to the persistence unit.  To confirm, I just tried out packaging the entities and defining a persistence unit within a JAR file of the application (which cannot load from web modules) and in a web module defined a repository that uses the entity and has the persistence unit as its data store. This worked fine as expected (using Data 1.0 and Persistence 3.2/EclipseLink), with the web module being able to successfully inject and perform operations on the repository, despite the persistence unit not having access to the repository.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
scottmarlow: CC @bstansberry

PR #1147, opened 2026-06-28
concurrency fixes and simplifications to PersistenceProviderResolverHolder

gavinking:
This code was a real mess and very difficult to understand in a way which hid bugs.

### DIFF
--- a/api/src/main/java/jakarta/persistence/spi/PersistenceProviderResolverHolder.java
+++ b/api/src/main/java/jakarta/persistence/spi/PersistenceProviderResolverHolder.java
@@ -11,6 +11,7 @@
  */
 
 // Contributors:
+//     Gavin King      - 4.0
 //     Lukas Jungmann  - 2.2
 //     Linda DeMichiel - 2.1
 //     Linda DeMichiel - 2.0
@@ -20,8 +21,6 @@ package jakarta.persistence.spi;
 import jakarta.annotation.Nonnull;
 import jakarta.annotation.Nullable;
 
-import java.lang.ref.Reference;
-import java.lang.ref.ReferenceQueue;
 import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
 import java.lang.System.Logger;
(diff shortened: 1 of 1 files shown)


api/src/main/java/jakarta/persistence/spi/PersistenceProviderResolverHolder.java
```diff
@@ -62,77 +64,66 @@ public static PersistenceProviderResolver getPersistenceProviderResolver() {
...
-         */
-        private final Map<CacheKey, PersistenceProviderReference> providers = new HashMap<>();
-
-        /**
-         * Queue for reference objects referring to class loaders or persistence providers.
+         * Cached lists of available providers keyed by {@linkplain LoaderKey class loader}
+         * to avoid provider visibility issues.
          */
-        private static final ReferenceQueue<Object> referenceQueue = new ReferenceQueue<>();
+        private final Map<LoaderKey, SoftReference<List<PersistenceProvider>>> providers = new HashMap<>();
```
marko-bekhta, line 87: I was looking at 
https://github.com/jakartaee/validation/blob/ccaaa92c61b5bc347d8186289ab41538e1adc944/src/main/java/jakarta/validation/Validation.java#L320-L388
and I think we can go with `WeakHashMap` + removing the key type and `removeClearedReferences()`.
That is to simplify it even a bit further 🙂

gavinking: So that would have at least two impacts:

1. It would change the key equality to be determined by `ClassLoader.equals()` instead of by `==`, which is likely OK, but it would be a change in externally-visible behavior.
2. We would still have to do something about the null thread context class loader case.

So we could do this, but is it worth the (admittedly tiny) risk?

marko-bekhta: yeah ... I guess let's keep it the way you have it 🙂 
even if these resolvers are quite similar, probably safer not to go too far in unifying 🙂

marko-bekhta (approved):

scottmarlow (approved):
Looks good to me.

gavinking:
Thanks for the approvals.

PR #1150, opened 2026-06-30
jakarta.persistence.EntityListener XML declaration

### LINKED ISSUE #1148: entity-listener XML element
The `entity-listener` was pre-existing, and has different semantics to the new `@EntityListener` annotation.

Because `entity-listener` only defines listener methods of classes declared using `@EntityListeners`, whereas `@EntityListener` declares a sort of auto-applied listener class.

So I think we should do something like add `auto-apply` or whatever to `entity-listener`.

sebersole:
Possible solution to https://github.com/jakartaee/persistence/issues/1148.  This option uses in-line repeatable elements following the lead from other repeatable elements in this XSD.  E.g.

```xml
<persistence ...>
    <persistence-unit ....>
        <class>com.acme.A</class>
        <class>com.acme.B</class>

        ...

        <entity-listener>com.acme.Journaler</entity-listener>
        <entity-listener>com.acme.PermissionChecker</entity-listener>
    </persistence-unit>
</persistence>
```

### DIFF
--- a/spec/src/main/asciidoc/ch08-entity-packaging.adoc
+++ b/spec/src/main/asciidoc/ch08-entity-packaging.adoc
@@ -522,6 +522,11 @@ The `default-to-one-fetch-type` element determines the default fetch
 type for one-to-one and many-to-one associations, as specified in
 <<fetching-policies>>.
 
+
+===== entity-listener
+
+The `entity-listener` element allows to specify zero-or-more `jakarta.persistence.EntityListener`-style listener classes.
+
 ===== properties [[a12384]]
 
 The `properties` element lists settings for both standard and
--- a/api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
+++ b/api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
@@ -264,6 +264,17 @@
 
               <!-- **************************************************** -->
 
+              <xsd:element name="entity-listener" type="xsd:string" minOccurs="0" maxOccurs="unbounded">
+                <xsd:annotation>
+                  <xsd:documentation>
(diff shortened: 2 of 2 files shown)


spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -522,6 +522,11 @@ The `default-to-one-fetch-type` element determines the default fetch
 type for one-to-one and many-to-one associations, as specified in
 <<fetching-policies>>.
 
+
+===== entity-listener
+
+The `entity-listener` element allows to specify zero-or-more `jakarta.persistence.EntityListener`-style listener classes.
```
gavinking:
```suggestion
The `entity-listener` element declares an entity listener class with 
semantics equivalent to the `@EntityListener` annotation defined in 
<<entity-listener-annotation>>.
```

gavinking (approved):
LGTM, after the change I suggest below.

marko-bekhta (approved):
maybe worth having some quick test next to https://github.com/jakartaee/persistence/blob/469e7e437bad75716ecfcde590fa64ea2e0a36b3/tck/spec-tests/src/main/java/ee/jakarta/tck/persistence/jpa40/emflistener/PersistenceUnitLifecycleClient.java#L35 ?

gavinking:
Is there any reason we didn't merge this yet??

marko-bekhta:
> Is there any reason we didn't merge this yet??

Hmm, yeah, not sure 🤷🏻. That mention of a test definitely shouldn't be blocking this one from being merged.

gavinking:
OK, merged, thanks @sebersole, good idea.

PR #1153, opened 2026-07-03
fix stale documentation in persistence_4_0.xsd

gavinking:
+ very minor improvement to spec

### DIFF
--- a/api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
+++ b/api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
@@ -55,29 +55,29 @@ public interface PersistenceUnitInfo {
      * Returns the fully qualified name of the persistence provider
      * implementation class. Corresponds to the {@code provider} element
      * in the {@code persistence.xml} file.
-     * @return  the fully qualified name of the persistence provider 
-     * implementation class
+     * @return the fully qualified name of the persistence provider
+     *         implementation class
      */
     @Nullable
     String getPersistenceProviderClassName();
 
     /**
-     * Returns the fully-qualified class name of an annotation annotated
+     * Returns the fully qualified class name of an annotation annotated
      * {@code Scope} or {@code NormalScope}. Corresponds to the {@code scope}
      * element in {@code persistence.xml}.
-     * @return  the fully-qualified class name of the scope annotation,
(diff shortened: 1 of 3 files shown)


api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
```diff
@@ -200,9 +202,9 @@
                 <xsd:annotation>
                   <xsd:documentation>
 
-                    When set to true then only listed classes and jars will
-                    be scanned for persistent classes, otherwise the
-                    enclosing jar or directory will also be scanned.
+                    When set to true, only explicitly listed jars are scanned
+                    for classes annotated with discoverable annotation types;
+                    otherwise, the enclosing jar or directory is also scanned.
                     Not applicable to Java SE persistence units.
```
marko-bekhta, line 208: maybe we should sync it with https://github.com/jakartaee/persistence/blob/7beb854f7ef067e8935e94f1faa0816104a408e3/api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java#L215 ? (sync as in  copy-paste 🙈 to keep the descriptions matching ?)

also.. should we mention module/package here as well ?

api/src/main/java/jakarta/persistence/spi/PersistenceUnitInfo.java
```diff
@@ -194,33 +194,33 @@ public interface PersistenceUnitInfo {
...
-     * @return whether classes in the root of the persistence
-     * unit that have not been explicitly listed are to be
-     * included in the set of managed classes
+     * Determines whether the root directory of the persistence
+     * unit is scanned for classes annotated with discoverable
+     * annotation types. The returned boolean value corresponds
+     * to the value of the {@code exclude-unlisted-classes}
+     * element in the {@code persistence.xml} file.
+     * @return {@code false} if the root directory of the
+     *         persistence unit is scanned
```
gavinking, line 212: @marko-bekhta better?

api/src/main/resources/jakarta/persistence/persistence_4_0.xsd
gavinking: > should we mention module/package here as well

No need. We all agree that these are actually classes.

marko-bekhta (approved):
thanks 🙇🏻 🎉

PR #1156, opened 2026-07-03
Rename TCK client classes / TCK compatibility with h2

gavinking:
This PR does two things:

1. Robotically renames all the millions of `Client` classes to more identifiable names.
2. Improves execution of the TCK on h2.


marko-bekhta (approved):
I'll admit that I haven't checked each new calss name 🙂 but it all looks ok in general. 👍🏻 🙂 

btw the rename changed the execution order and I think it uncovered an issue with classloading -- looking into that one..

oh and do you think we should make the tck passable on h2 in general ? (I suspect some tests related to stored procedures might not work right now...)

gavinking:
> I think it uncovered an issue with classloading -- looking into that one..

Oh, OK, thanks!

> oh and do you think we should make the tck passable on h2 in general? 

I got all tests passing on h2 except the SP tests. I doubt there's any reasonable way to get them working on h2.

PR #1160, opened 2026-07-30
add remaining List/varargs counterparts in the criteria API

### LINKED ISSUE #1159: remaining varargs/List overload gaps in the criteria API
#137 added `where(List<Predicate>)` and friends to `CriteriaQuery` in 3.2, and #1075 extended this to `CriteriaUpdate`/`CriteriaDelete` and the join `on()` methods for 4.0, "for consistency with what we already did to CriteriaQuery in 3.2". A sweep of the rest of the criteria API turns up four operations still missing the corresponding counterpart:

1. **`CriteriaBuilder#construct(Class, Selection<?>...)`** — its siblings `tuple()` and `array()` both gained `List<Selection<?>>` overloads in 3.2, but `construct()` did not, even though constructor argument lists are just as often assembled dynamically.

2. **`CriteriaBuilder#function(String, Class, Expression<?>...)`** — the canonical dynamic-arity operation (argument count depends on the SQL function, so arguments are typically collected in a loop), with no `List` counterpart.

3. **`Expression#in(Expression<?>...)`** — no list counterpart, while its sibling `in(Object...)` has had `in(Collection<?>)` since 2.0. Two notes:
   - the overload must be declared with `List` rather than `Collection`, since `in(Collection<? extends Expression<?>>)` clashes with the erasure of the existing `in(Collection<?>)`;
   - today `expr.in(listOfExpressions)` statically binds to `in(Collection<?>)`, treating the expressions as *literal values* — surely never what the caller meant. The new overload is more specific, so on recompile such calls re-bind to the intended semantics. Flagging this binding change explicitly since it deserves deliberate sign-off, though it seems strictly an improvement.

4. **`CriteriaBuilder#concat(List<Expression<String>>)`** (3.2, #408) — the inverse gap: a `List` form with no varargs counterpart. In 3.2 a varargs form wasn't cleanly expressible (`Expression<String>...` triggers unchecked generic array creation warnings at every call site), but 4.0's non-generic `TextExpression` makes `concat(TextExpression...)` expressible — the same device that gave `and()`/`or()` their varargs forms.

As far as I can tell these are the only remaining varargs/`List` asymmetries in `jakarta.persistence.criteria` (deliberate exclusions like the deprecated `multiselect()`, `Fetch` having no `on()`, and the `coalesce()`/`in()` builder objects aside).

ymajoros:
Fixes #1159 — completes the overload alignment of #137 (3.2) and #1075 (4.0) across the rest of the criteria API:

- `CriteriaBuilder#construct(Class, List<Selection<?>>)`, matching the 3.2 List forms of its siblings `tuple()` and `array()`
- `CriteriaBuilder#function(String, Class, List<? extends Expression<?>>)`
- `CriteriaBuilder#concat(TextExpression...)`, now expressible thanks to the non-generic `TextExpression`, mirroring `and()`/`or()`
- `Expression#in(List<? extends Expression<?>>)`, the list counterpart of `in(Expression<?>...)`

All four are additive; javadoc and `@since 4.0` follow the surrounding style, wildcards follow the `and()`/`or()` precedent (invariant `List<Selection<?>>` for `construct`, matching `tuple`/`array`). Verified with a call-site smoke compile under `-Xlint:all` that every pre-existing call shape still binds to the same overload (in particular binary `concat(x, y)` is still selected in the applicable-by-arity phases before the new varargs form, and `in(Collection<?>)` still receives literal-value collections); the one deliberate re-binding — `expr.in(listOfExpressions)` moving from literal-value to expression semantics — is called out in the issue.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -440,11 +440,29 @@ StatementReference augment(@Nonnull StatementReference reference,
...
+     * @param selections  list of arguments to the constructor
+     * @param <Y> the type of the constructed result
+     * @return compound selection item
+     * @throws IllegalArgumentException if an argument is a
+     *         tuple- or array-valued selection item
+     * @since 4.0
+     */
+    @Nonnull
+    <Y> CompoundSelection<Y> construct(@Nonnull Class<Y> resultClass,
+                                       @Nonnull List<Selection<?>> selections);
```
gavinking: I don't think this one is really very useful. The arguments to a typical constructor are not varargs, and they're not usually going to be passed around as a `List`. Typically, you're going to have an exact number of arguments of different types. 

That's different to the case of `array()` and, at least arguably, to the case of `tuple()`.

So I think the divergence here is reasonably well-motivated, and that there was a reason @lukasj and I didn't add this in 3.2.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -1783,6 +1801,17 @@ Predicate notLike(@Nonnull Expression<String> x,
...
+    /**
+     * Create an expression for string concatenation.
+     * If no expressions are given, returns an expression
+     * equivalent to {@code literal("")}.
+     * @param expressions  zero or more string expressions
+     * @return expression corresponding to concatenation
+     * @since 4.0
+     */
+    @Nonnull
+    Expression<String> concat(@Nonnull TextExpression... expressions);
```
gavinking, line 1795: This one I'm not sure about. This introduces the possibility of nullary `concat()`, which is not currently defined. Of course we could say that `concat()` returns the empty string, as described in the Javadoc proposed here, but it's not enough to say that just in the Javadoc of the criteria API. We would also have to update the Jakarta Query spec for completeness. I don't object to this change; just saying that it has a wider blast radius.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
```diff
@@ -2575,6 +2604,29 @@ Case<R> when(@Nonnull Expression<Boolean> condition,
...
+     * @param name  function name
+     * @param type  expected result type
+     * @param args  a list of function arguments
+     * @param <T> the function result type
+     * @return expression
+     * @since 4.0
+     */
+    @Nonnull
+    <T> Expression<T> function(@Nonnull String name, @Nonnull Class<T> type,
+                               @Nonnull List<? extends Expression<?>> args);
```
gavinking: I'm not at all a fan of the `function()` function, so I don't care much what happens to it.

That said, I think this is similar to `construct()`, and that the same reasoning applies here: the arguments to a typical SQL function are not varargs, and they're not usually going to be passed around as a `List`. Typically, you're going to have an exact number of arguments of different types.

api/src/main/java/jakarta/persistence/criteria/Expression.java
```diff
@@ -104,6 +105,16 @@ public interface Expression<T> extends Selection<T> {
...
 
+    /**
+     * Create a predicate to test whether the expression is a member
+     * of the argument list.
+     * @param values  list of expressions to be tested against
+     * @return predicate testing for membership
+     * @since 4.0
+     */
+    @Nonnull
+    Predicate in(@Nonnull List<? extends Expression<?>> values);
```
gavinking, line 116: This one would be convenient, but an incautious developer risks collisions with this existing (since JPA 1.0) overload:

```java
Predicate in(@Nonnull Collection<?> values);
```

That's not a super-strong objection, since as long as the static type is really exactly `List<? extends Expression<?>>`, the right overload will be called, but it's just slightly to close for my complete comfort.

gavinking (commented):
Thanks for the suggestions, [ymajoros](https://github.com/ymajoros).

None of your suggestions are terrible, but I do have some concerns/caveats about every one of them; which probably explain why none of those overloads already exist.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
ymajoros: Fair enough, and the fixed-arity point convinces me. Unlike `array()` and `tuple()`, a constructor has a specific signature, so a dynamically assembled list has nothing to check itself against until execution time, and the ergonomic win mostly evaporates.

The case I had in mind was a generic result-mapping layer that reflects over a DTO constructor and builds the selection list from metadata, but that is niche enough to live with `toArray()`. Dropped in 8804632.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
ymajoros: Dropped this one too, in 8804632.

For the record I think the arity argument is a little weaker here than for `construct()`: a SQL function's arity is fixed by the function, but the calling code is sometimes generic over the function rather than over the arguments, e.g. a small dispatch layer mapping a name plus an argument list onto `function()`. That is a much narrower audience than `where()` though, and given your view of `function()` in general it does not seem worth your review time.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
ymajoros: I think the blast radius is already there, though. `concat(List)` has specified since 3.2 (b404407, the commit that introduced it) that an empty list returns an expression equivalent to `literal("")`:

```java
/**
 * Create an expression for string concatenation.
 * If the given list of expressions is empty, returns
 * an expression equivalent to {@code literal("")}.
 */
Expression<String> concat(List<Expression<String>> expressions);
```

So `cb.concat(List.of())` is already defined and already reachable, and the varargs form only gives that same expression a shorter spelling rather than introducing new semantics. If Jakarta Query is silent on empty concatenation, that gap exists independently of this PR, and arguably wants fixing either way.

That said, if you would rather not make nullary `concat()` reachable by accident, `concat(TextExpression, TextExpression...)` requires at least one argument and side-steps the question entirely, at the cost of a slightly uglier signature. Happy with either, just say which you prefer.

api/src/main/java/jakarta/persistence/criteria/Expression.java
ymajoros: Agreed they sit close together, but I would argue the collision already exists today and currently resolves the wrong way round.

`expr.in(listOfExpressions)` compiles right now and binds to `in(Collection<?>)`, since `List` is a `Collection` and that overload is applicable in an earlier phase than `in(Object...)`. Its contract is "collection of values to be tested against", so the portable reading is that those `Expression` objects are literal values, which is surely never what the caller meant. Providers may well be lenient about it in practice, but the spec-level meaning is the unhelpful one, and it is silent either way.

So the new overload does not really create the hazard, it makes the intended reading the one the compiler picks: being more specific, it wins for a static type of `List<? extends Expression<?>>` while genuine literal collections keep binding to `in(Collection<?>)` exactly as before.

The part that does deserve scrutiny is that this re-binds existing source on recompile. It is binary compatible, and the re-binding moves calls from a meaning nobody wants to the meaning they wrote, but it is a real behavioural change rather than a pure addition, so I would not want it slipped in quietly.

api/src/main/java/jakarta/persistence/criteria/CriteriaBuilder.java
ymajoros: Not reopening this — it stays dropped — but one data point for the record, since I ran into it right after.

Sweeping our own codebase for array conversions that the `List` overloads made unnecessary, every single one had a `List` counterpart to move to except one, and it was `function()`:

```java
public Expression<Long> countDistinct(List<? extends Expression<?>> expressions) {
    Expression[] expressionArray = expressions.toArray(new Expression[0]);
    // postgres-specific: group multiple columns
    Expression<Object> rowExpression = criteriaBuilder.function("ROW", Object.class, expressionArray);
    return criteriaBuilder.countDistinct(rowExpression);
}
```

That is the shape you described as atypical: the caller is generic over the column list rather than over the function, so the arguments genuinely do arrive as a `List` and the arity is not known at the call site. It is a helper working around `countDistinct` not accepting multiple expressions, so arguably it is a workaround begetting a workaround, and the raw `Expression[]` it forces is its own small wart.

One case is not a mandate, and I still think it is the weakest of the four. Just noting the use case exists in the wild rather than leaving the impression I could not name one.

PR #1165, opened 2026-08-20
clarify status of Java value types

### LINKED ISSUE #1164: Value types in JPA
[Value Objects](https://openjdk.org/jeps/401) have been integrated into Java 28 in preview mode. Is there any bearing on JPA with regard to value objects that needs to be documented or implemented in the future? A lot of the Basic types will be [converted](https://openjdk.org/jeps/401#Programming-with-value-objects) to value objects.

For example, if I declare
```java
value class Pages {

    private long pages;  // implicitly final
}
```
will there be any difference with respect to the same class without the `value` declaration?

It will be some time before value types leave preview, but it's good to have some foresight for JPA/JakartaEE because their releases are less frequent than Java's.

gavinking:
Fixes #1164.

### DIFF
--- a/spec/src/main/asciidoc/ch02-entities.adoc
+++ b/spec/src/main/asciidoc/ch02-entities.adoc
@@ -23,6 +23,7 @@ declared as an `entity` in its XML descriptor.
 
 - The entity class must be a top-level class or a static inner class.
   An enum, record, or interface may not be designated as an entity.
+  A value class may not be designated as an entity.
 
 - The entity class must have a `public` or `protected` constructor with no
   parameters, which is called by the persistence provider runtime to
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -23,6 +23,7 @@ declared as an `entity` in its XML descriptor.
 
 - The entity class must be a top-level class or a static inner class.
   An enum, record, or interface may not be designated as an entity.
+  A value class may not be designated as an entity.
```
marko-bekhta, line 26:
```suggestion
An enum, record, value class, or interface may not be designated as an entity.
```
I'd just add to that list.

marko-bekhta (approved):
Looks reasonable 👍🏻

njr-11 (commented):
The changes make sense and look good.  Do we have any precedent on whether to update Jakarta EE specs to account for preview features of Java SE vs waiting for it to be announced that the feature will be going into a long term support release of Java SE?  I'll ask some of the developers that I work with if they have any thoughts on this topic.  At this point, my own preference would be to make the statements about `value` types as notes and clearly indicating that `value` types are a preview feature as of the point in time that the Persistence 4.0 specification was written.

spec/src/main/asciidoc/ch02-entities.adoc
nlisker: A value class is not disjoint from a record, but it may be splitting hairs.

spec/src/main/asciidoc/ch02-entities.adoc
```diff
@@ -1594,14 +1596,19 @@ persistence provider is not required to respect the object identity of
...
 requirements specified in <<a18>> for entities, with the exception that
-an embeddable class is not annotated as `Entity`, and an embeddable
-class must not be `abstract`.
+an embeddable class:
+
+- is not annotated as `Entity`;
+- must not be `abstract`;
+- but may be a `value` type.
 
 Alternatively, an embeddable class may be a Java record type.
```
nlisker, line 1613: It's not "Alternatively", a record type can be a value type as well (`value record`).

nlisker (commented):
I didn't go over the specs outside of this PR to see if something was missed. Looks sensible.

PR #1167, opened 2026-08-28
add lists of cases to exception type javadoc

### LINKED ISSUE #1169: EM.getLockMode() should return NONE when there is no active tx
Currently the Javadoc for `EM.getLockMode()` states that it throws `TransactionRequiredException` when called outside a tx. However, this requirement has never been documented in the relevant sections of the spec, and looks silly to me.

I propose to change that to say that `getLockMode()` return `NONE` in this situation.

In principle this could break code which uses the `TransactionRequiredException` thrown by `getLockMode()` to detect the case of there being no tx, but I really doubt that there is any code which does this, and so this would be IMO an acceptable change in behavior for a major version.

gavinking:
there is no semantics change here, I've just copied information from the spec to the jdoc.

EDIT: now also realizes https://github.com/jakartaee/persistence/issues/1169

### DIFF
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -795,12 +795,11 @@ public interface EntityManager extends EntityHandler {
      * Get the current {@linkplain LockModeType lock mode} held by
      * this persistence context on the given managed entity instance.
      * @param entity  a managed entity instance
-     * @return the lock mode currently held
-     * @throws TransactionRequiredException if there is no active
-     *         transaction, or if the entity manager has not been
-     *         joined to the current transaction
-     * @throws IllegalArgumentException if a transaction is active
-     *         but the given instance is not a managed entity
+     * @return the lock mode currently held, or {@link LockModeType#NONE}
+     *         if there is no active transaction or if the entity manager
+     *         has not been joined to the current transaction
+     * @throws IllegalArgumentException if the given instance is not
+     *         a managed entity
      * @since 2.0
      */
     @Nonnull
--- a/api/src/main/java/jakarta/persistence/EntityManager.java
+++ b/api/src/main/java/jakarta/persistence/EntityManager.java
@@ -796,7 +796,8 @@ public interface EntityManager extends EntityHandler {
      * this persistence context on the given managed entity instance.
      * @param entity  a managed entity instance
      * @return the lock mode currently held, or {@link LockModeType#NONE}
-     *         if there is no active transaction or if the entity manager
+     *         if the given {@code entity} instance is managed, but
+     *         there is no active transaction or the entity manager
      *         has not been joined to the current transaction
      * @throws IllegalArgumentException if the given instance is not
      *         a managed entity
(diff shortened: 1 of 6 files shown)


api/src/main/java/jakarta/persistence/TransactionRequiredException.java
```diff
@@ -18,8 +18,36 @@
...
+ *     {@link EntityManager#remove}, {@link EntityManager#merge}, or
+ *     {@link EntityManager#refresh} is called on an entity manager with
+ *     a {@linkplain PersistenceContextType#TRANSACTION transaction-scoped}
+ *     persistence context with no active transaction context.
+ * <li>The {@link EntityManager#lock} operation is called on an entity
+ *     manager with no active transaction context.
+ * <li>Any operation of {@code EntityManager} or {@code EntityAgent} which
+ *     accepts a {@linkplain LockModeType lock mode} is called on a manager
+ *     or agent with no active transaction context, and the given lock mode
+ *     is not {@link LockModeType#NONE}.
```
marko-bekhta, line 41: there also seems to be 

https://github.com/jakartaee/persistence/blob/2272b2eab119939fcd8d698d61d31ea7d27f61e5/api/src/main/java/jakarta/persistence/EntityManager.java#L799-L807

so maybe something like "accepts or returns a {@linkplain LockModeType lock mode}` ?

marko-bekhta (commented):
not sure whether we should mention that `getLockMode` or not ... otherwise looks good 🙂 👍🏻

api/src/main/java/jakarta/persistence/EntityManager.java
```diff
@@ -795,12 +795,11 @@ void refresh(@Nonnull Object entity,
...
      * @param entity  a managed entity instance
-     * @return the lock mode currently held
-     * @throws TransactionRequiredException if there is no active
-     *         transaction, or if the entity manager has not been
-     *         joined to the current transaction
-     * @throws IllegalArgumentException if a transaction is active
-     *         but the given instance is not a managed entity
+     * @return the lock mode currently held, or {@link LockModeType#NONE}
+     *         if there is no active transaction or if the entity manager
+     *         has not been joined to the current transaction
```
njr-11: Eliminates potential ambiguity over the behavior when the entity is unmanaged and the entity manager is not joined to a transaction
```suggestion
     *         if the given {@code entity} instance is managed, but
     *         there is no active transaction or the entity manager
     *         has not been joined to the current transaction
```

njr-11 (commented):
Good idea adding detail to the Javadoc.  Regarding the change from em.getLockMode raising TransactionRequiredException to returning LockModeType.NONE, Jakarta EE allows changes that make what used to be an exception path into a working path as compatible, so I think a change like this would be acceptable even in a minor version release. Applications should not have been catching TransactionRequiredException to find out whether the entity manager is joined to the transaction because JPA already provides EntityManager.isJoinedToTransaction() as a better way of determining that without forcing an exception.

njr-11 (approved):

marko-bekhta (approved):

PR #1180, opened 2026-09-10
Preserve class, package, and module descriptor categories in persistence-unit metadata

sebersole:
JPA 3.2 added support for some package-level annotations.  JPA 4 has added support for some module-level annotations.  

However, at no point has the XSD allowed definition of either.  I know historically various providers allowed specifying packages via `<class/>` but that is explicitly not allowed, per https://jakarta.ee/specifications/persistence/3.2/jakarta-persistence-spec-3.2.html#list-of-managed-classes

> A list of named managed persistence classes—entity classes, embeddable classes, mapped superclasses, and converter classes—may be specified instead of, or in addition to, the listed JAR files and mapping files. Any mapping metadata annotations found on these classes are processed. Where mapping annotations are missing, the classes are mapped using the mapping annotation defaults. The class element is used to list a managed persistence class.
> 
> In Java SE environments, an explicit list of all managed persistence class names must be specified to insure portability. Portable Java SE applications should not rely on the other mechanisms described here to determine the managed persistence classes of a persistence unit. In Java SE environments, a persistence provider may require that the set of entity classes and other classes to be managed is fully enumerated in each persistence.xml file.

This pull request rectifies that with explicit support for specifying packages and modules which can be searched for annotations (not recursively):

* adds `<package/>` and `<module/>` elements to the XSD
* adds `managedPackages()`, `managedModules()`, `managedPackage(String)` and `managedModule(String)` methods to `PersistenceConfiguration`
* adds `getManagedPackageNames()`, `getManagedModuleNames()`, `getAllPackageNames()` and `getAllModuleNames()` to `PersistenceUnitInfo`
* clarifies that `getAllClassNames()` on `PersistenceUnitInfo` refers TYPE exclusively
* clarifies that only entries from `getAllClassNames()` are eligible for transformation.


spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -490,21 +508,35 @@ app.ear
...
+annotations belong to the persistence unit. Its value is the qualified package
+name, for example, `com.example.model`, and not
+`com.example.model.package-info`. Naming a package descriptor does not cause
+the ordinary Java types in that package to belong to the persistence unit.
+
+The `module-descriptor` element identifies one module descriptor whose
+annotations belong to the persistence unit. Its value is the declared JPMS
+module name and not `module-info`. Naming a module descriptor does not cause
+the ordinary Java types or package descriptors in that module to belong to the
+persistence unit.
```
gavinking: @sebersole Can you turn this into a bullet list please.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -490,21 +508,35 @@ app.ear
...
-determine the managed persistence classes of a persistence unit. In
-Java SE environments, a persistence provider may require that the set
-of entity classes and other classes to be managed is fully enumerated
-in each `persistence.xml` file.
+In Java SE environments, an explicit list of all managed persistence class
+names and any package or module descriptors whose annotations must be processed
+must be specified to ensure portability. Portable Java SE applications should
+not rely on the other mechanisms described here to determine the program
+elements of a persistence unit. In Java SE environments, a persistence provider
+may require that these sets are fully enumerated in each `persistence.xml` file.
```
gavinking:
```suggestion
may require that this list is fully enumerated in each `persistence.xml` file.
```

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -490,21 +508,35 @@ app.ear
...
 
-In Java SE environments, an explicit list of all managed persistence
-class names must be specified to ensure portability. Portable Java SE
-applications should not rely on the other mechanisms described here to
-determine the managed persistence classes of a persistence unit. In
-Java SE environments, a persistence provider may require that the set
-of entity classes and other classes to be managed is fully enumerated
-in each `persistence.xml` file.
+In Java SE environments, an explicit list of all managed persistence class
+names and any package or module descriptors whose annotations must be processed
```
gavinking:
```suggestion
In Java SE environments, an explicit list of all managed persistence classes
and any package or module descriptors whose annotations must be processed
```

spec/src/main/asciidoc/ch09-container-provider-contracts.adoc
```diff
@@ -839,6 +855,14 @@ environment.footnote:[Persistence units defined programmatically using the
 `PersistenceConfiguration` class do not support JNDI lookup or injection via
 the `PersistenceContext` or `PersistenceUnit` annotations.]
 
+The `managedClass()`, `managedPackageDescriptor()`, and
+`managedModuleDescriptor()` operations correspond to the `class`,
+`package-descriptor`, and `module-descriptor` elements of `persistence.xml`.
+A package is specified by its qualified package name without
```
gavinking:
```suggestion
`package-descriptor`, and `module-descriptor` elements of `persistence.xml`,
respectively. A package is specified by its qualified package name without
```

gavinking (commented):
@sebersole looks good but I made some minor suggestions around wording.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
sebersole: Since this is discussing all 3 groupings, it seems like the plural is proper.  

I don't much care about the list/set distinction I guess, though conceptually they are sets.

So I'll change this to "these lists are" if you prefer

gavinking: The reason is that it's an anaphora for something you called one list at the start of the paragraph. "These sets" has no antecedent.

spec/src/main/asciidoc/ch08-entity-packaging.adoc
```diff
@@ -490,21 +508,35 @@ app.ear
...
-determine the managed persistence classes of a persistence unit. In
-Java SE environments, a persistence provider may require that the set
-of entity classes and other classes to be managed is fully enumerated
-in each `persistence.xml` file.
+In Java SE environments, an explicit list of all managed persistence classes
+and any package or module descriptors whose annotations must be processed
+must be specified to ensure portability. Portable Java SE applications should
+not rely on the other mechanisms described here to determine the program
+elements of a persistence unit. In Java SE environments, a persistence provider
+may require that these lists are fully enumerated in each `persistence.xml` file.
```
gavinking:
```suggestion
In Java SE environments, the persistence provider is not required to process any 
type, package descriptor, or module descriptor which is not explicitly listed in the 
`persistence.xml` file.
```

PR #1182, opened 2026-09-13
documentation changes and tests relating to LockModeType

### LINKED ISSUE #1181: Are lockModes other than NONE compatible with EntityAgent?
Documentation of QueryOptions.lockMode needs clarification around whether it is applies to EntityAgent/stateless entities.
Currently, the documentation says:

```java
    /**
     * The lock mode type to use in query execution.
     * If a {@code lockMode} other than {@link LockModeType#NONE}
     * is specified, the query must be executed in a transaction
     * and the persistence context joined to the transaction.
     * <p> If a lock mode is explicitly specified for a
     * {@linkplain NativeQuery native query}, the behavior
     * is undefined and unportable between persistence providers.
     * <p>
     * This option applies when the query is a JPQL {@code select}
     * query or a native query that returns a result set.
     *
     * @see jakarta.persistence.TypedQuery#setLockMode
     */
    LockModeType lockMode() default LockModeType.NONE;
```

Specifically, `and the persistence context joined to the transaction` could be read as
"There must be a persistence context, and it must be joined to the transaction"
or
"If there is a persistence context, then it must be joined to the transaction"

Regarding behavior, some of the lock modes, such as optimistic locking and pessimistic read/write locking would be valuable for EntityAgent.  Whereas a couple of them, such as `PESSIMISTIC_FORCE_INCREMENT`, seem like they would need special handling from the Persistence provider if they are going to work (in this case, to write the version increment to the database and unmanaged entity instance that is immediately returned to the caller rather than deferring until commit time).

This impacts Jakarta Data users that will be applying the `@QueryOptions(lockMode = ...)` annotation to repository methods and will be wondering whether it can be used on stateless repositories or only on stateful repositories.

gavinking:
- The spec did not completely generalize lock modes to `EntityAgent`
- The TCK had very poor coverage of anything related to locking
- There were some issues with the documentation of `refresh()`
- Added missing citation to Berenson 1995.

There should be no semantic changes here.

This now fixes #1181.


api/src/main/java/jakarta/persistence/EntityAgent.java
marko-bekhta, line 319:
```suggestion
     * @throws IllegalArgumentException if the given instance is not an entity
     * @throws PersistenceException if a record could not be
```

maybe ? since we are adding that to a refreshmultiple ?

api/src/main/java/jakarta/persistence/LockModeType.java
```diff
@@ -23,11 +23,12 @@
...
  * {@code LockModeType} as an argument to:
  * <ul>
- * <li>one of the methods of {@link EntityManager} which obtains
- *     locks ({@link EntityManager#lock lock()},
- *     {@link EntityManager#find find()}, or
+ * <li>one of the methods of {@link EntityManager} or
+ *     {@link EntityAgent} which obtains locks
+ *     (for example, {@link EntityManager#lock lock()},
+ *     {@link EntityHandler#find find()}, or
  *     {@link EntityManager#refresh refresh()}), or
```
marko-bekhta, line 30: We have EM and EH examples here; maybe, for completeness, we should have one for an EA as well?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1336,10 +1336,13 @@ that enables the effect of “repeatable read” semantics for the items
...
 or “pessimistically” (as described in <<a2113>>).
 
+See Berenson et al. <<a19504>>, Section 2, for definitions of the
+phenomena P1 and P2 used below.
+
 A lock mode may be explicitly specified as an argument to the
 `lock()` method of `EntityManager` or to any other method of
-`EntityManager`, `Query`, and `TypedQuery` which accepts a lock
-mode, or via the `NamedQuery` annotation.
+`EntityManager`, `EntityAgent`, `Query`, or `TypedQuery` which
```
marko-bekhta, line 1344: since we say query/typedquery should we also mention the entityhandler ?

spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -1425,7 +1428,7 @@ call to `EntityManager.lock` was made.
 For portability, an application should not
 depend on vendor-specific hints or configuration to ensure repeatable
 read for objects that are not updated or removed via any mechanism other
-than the use of version attributes and the EntityManager `lock` method.
+than the use of version attributes and the `lock` method of `EntityManager`.
```
marko-bekhta, line 1431: is it still just `lock()` ? Or can those get methods with FindOptions also be a portable solution too ?

njr-11 (approved):
The Javadoc and spec updates look great.  Approving.  I'll defer to others on the review of the TCK test because I haven't been involved in that aspect of this spec.

marko-bekhta (commented):
that's a nice improvement 🙂👍🏻 

I had some questiosn -- added them inline + and tried running the new tests -- there seem to be a few that failed so trying to check if that's on the impl or on the tests 🤞🏻🙂

api/src/main/java/jakarta/persistence/EntityAgent.java
gavinking: Indeed, good catch, thanks.

api/src/main/java/jakarta/persistence/LockModeType.java
gavinking: The trouble is it would read as the same thing twice. It's fine, it's just a link.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: We're assuming every `EH` is an `EA` or an `EM`.

gavinking: I see what you're saying but it's a bit different because a `TypedQuery` redeclares all those methods.

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: Yeah probably but if so that's an existing incompleteness in the spec arising from the fact that I have not yet sat down and rewritten that section carefully from scratch. There's a separate issue for that.

marko-bekhta (approved):
thanks 🙏🏻

spec/src/main/asciidoc/ch03-entity-operations.adoc
marko-bekhta: ok I'll get myself prepared for reading through the rewrite 🙂 (when it comes 🙂)

PR #1187, opened 2026-09-19
replace old "available" language with proper def in terms of "loaded"

gavinking:
Astra pointed out that there were some real inconsistencies here arising from the fact that the section talking about "available" state was written before the existence of `EntityAgent`.


spec/src/main/asciidoc/ch03-entity-operations.adoc
```diff
@@ -569,6 +569,64 @@ must not assume that pending modifications have not been written to the
...
 information about synchronization may be found above in <<flush>>.
 
+==== Loaded State of an Entity [[a2019]]
+
+An instance of an entity or embeddable class is considered _loaded_ or
+_fetched_ if every attribute of the entity or embeddable class mapped
+(explicitly or by default) as `FetchType.EAGER`, including relationships
+and element collections, has been loaded from the database or assigned
+by the application. A loaded entity or embeddable object might have an
+unloaded attribute if and only if the attribute is mapped, explicitly or
```
marko-bekhta, line 579:
```suggestion
_unloaded_ attribute if and only if the attribute is mapped, explicitly or
```

maybe? since it seems that the only place we are talking about what an unloaded attribute is ?

marko-bekhta (approved):

spec/src/main/asciidoc/ch03-entity-operations.adoc
gavinking: Mmmmmm dunno, I guess I think it's clear that "unloaded" is the same thing as "not loaded". Also that sentence is written as a prescription, not as a definition ("if and only if").

